PR #8, opened 2020-05-12
Add access_token to the logged output.

julianpas:
Useful for some experimentation with this code.

### DIFF
--- a/static/idp.html
+++ b/static/idp.html
@@ -21,6 +21,7 @@
 	    console.log(profile.getName());
 	    console.log(profile.getEmail());
 	    console.log(user.getAuthResponse().id_token);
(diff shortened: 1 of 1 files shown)


julianpas:
Btw I am working on getting my own web app for testing so this should be the last time I need to ask for a pull request to your repo :)

I am Julian from the enterprise team btw in case you are wondering who this is :)

asr-enid:
Just became aware of the stub code here, because of this PR. I would be good to discuss access_token handling with this API separately while speaking of it. We would not want access_tokens to our backends handled client side without a better understanding of the security implications . Does Google have an opinion on that?

dj2:
I'm going to close this due to lack of reply. Feel free to re-open if this is still an issue.

PR #21, opened 2020-07-14
Create privacy_threat_model.md

kenrb:
Add a proposed privacy threat model for federated identity.


privacy_threat_model.md
```diff
@@ -0,0 +1,75 @@
+*WebID Privacy Threat Model*
```
samuelgoto, line 1: may be worth having your github handle here somewhere as the author to allow people to find the domain expert easily.

PR #32, opened 2020-08-18
change problem links to readme doc

samuelweiler:
I think this is what you intended?  Maybe?


HowardWolosky:
I was noticing the same thing when reviewing it.  You missed two others:

https://github.com/WICG/WebID/blob/f7b12b81892bc9210fda807a9ecd3567d4b15a78/design.md#L45

https://github.com/WICG/WebID/blob/f7b12b81892bc9210fda807a9ecd3567d4b15a78/design.md#L226

PR #37, opened 2020-10-08
Add delegation api

majido:
Adds basic explanation for delegation based API.

TODO:
 - explain main sign up sequence 
 - Add diagram for flow
 - link from main solution document.


kenrb:
lgtm

PR #38, opened 2020-10-08
Mediation oriented api

samuelgoto:


mediation_oriented_api.md
```diff
@@ -0,0 +1,29 @@
...
+
+# Benefits
+
+- Fewer prompts compared to the [Permission-oriented](consumers.md#the-permission-oriented-variation) variation
+- Prompts are "functional" (part of a decision the user has to make) rather than "informational" (speed bumps)
+- The browser has the ability to control the defaults
+
+# Challenges
+
+- Browsers need to become opinionated about sign-in and the subset of flows that enables it.
```
majido, line 28: nit: s/enables/enable

majido, line 28: Is ossification a separate challenge or something that follows from this?

majido (approved):
lgtm % nits

mediation_oriented_api.md
samuelgoto: "enables" refers to the "subset" (not to the preprositional phrase "of flows") which is a singular?

samuelgoto: i think it is the same thing. i'll add a link to the terminology doc.

samuelgoto: ah, there isn't one ... i guess we should create an entry somewhere for ossification. will do that separately.

PR #39, opened 2020-10-08
Update permission_oriented_api.md

kenrb:


kenrb:
@samuelgoto This moves any additional details (and the mock) from consumers.md to permission_oriented_api.md, leaving only a high-level description in consumers.md. If this lgty then we can take a similar approach with other two variations.

samuelgoto (approved):

samuelgoto:
I think it would be worth keeping the mocks in both places because they tell a lot of the story at high density, but I can add them back after this PR.

PR #55, opened 2021-03-04
Update explainer with up to date understanding of enterprises, the HTTP/HTML API, naming, roadmap, etc.

samuelgoto:


README.md
```diff
@@ -1,85 +1,287 @@
...
+
+# Classification
+
+Like we said above, the classification problem is the browser's inability to distinguish identity federation from tracking due to the fact that both use low level primitives (namely, redirects, popups, iframes and cookies) and its consequence is the application of lowest common denominator policies.
+
+The first thought that occurred to us was to look into each of these low-level primitives and offer for each an indentity-specific high-level affordance, **trading generality for awareness**.
+
+With that in mind, lets look into each specific low-level primitive and what a high-level identity-specific affordance would look like.
+
+## The HTTP API
```
kenrb, line 144: I'm not clear on why "HTTP API" is used because this doesn't seem like we are defining an API at all (as opposed to the JS API, where we very clearly are adding API methods). An HTTP API is usually a way of interacting with a service via an HTTP endpoint. Does "API" here refer to the API provided by the IdP, and we would intercept requests being sent to it? That's not clear.

The term API has a specific meaning, so the text below should be clear about why it applies, or else we should consider a different term (examples could be HTTP sniffing, throttling, filtering, inspecting, gating, interception, etc).

Same comment for "The HTML API", where it seems like we are calling an element an API.

README.md
```diff
@@ -1,85 +1,287 @@
...
+// This is just a possible starting point, largely TBD.
+await navigator.credentials.store({
+  idtoken: JWT,
+});
+```
 
-- An early exploration of [The Solution Space](design.md)
+## The HTML API
 
+Relying Parties also typpically embed iframes served by identity providers for personalization (e.g. showing the user's profile picture / name on buttons). Browsers do (or are intending to) block third party cookies in iframes, making them uncredentialed and hence unable to personalize.
```
kenrb, line 193: "typpically" -> "typically"

README.md
```diff
@@ -1,85 +1,287 @@
...
+1. whether the same approach would work for other protocols (e.g. SAML).
+1. whether we need an opt-in / explicit API and if so which (e.g. perhaps a special URL marker, like a reserved URL parameter or a scheme)
+
+## The JS API
+
+Popups are harder to classify because each IDP seems to use a custom protocol to open the popup as well as to communicate via postMessage.
+
+It is hard to know what that will exactly look like right now, but as a starting point, here is what it could look like.
+
+Instead of the low level `window.open` and listening to `window` events, one coule write at a high-level:
```
kenrb, line 172: "coule" -> "could"

README.md
```diff
@@ -1,85 +1,287 @@
...
+
+For one, the user has to make **two** choices (on the consequences of tracking) that are unrelated to the job to be done (sign-in) which we don't expect to be the most effective way to affect change. That leads us to the [mediation-oriented](consumers.md#the-mediation-oriented-variation) variation which bundles these prompts into a browser mediated experience (which also comes with trade-offs).
+
+Secondly, we believe that we are possibly making the user make a determination (to be tracked) that isn't necessary. The [delegation-oriented](consumers.md#the-delegation-oriented-variation) variation (which, again, comes with its set of trade-offs too) tries to solve the tracking risks by pulling more responsibilites for the user agent.
+
+It is an active area of investigation to determine the **relationship** between these approaches. To the best of our knowledge so far, we expect these to be mutually complementary (rather than exclusive) and to co-exist long term. Each comes with trade-offs and it is still early to know what market (if any) each fits.
+
+## Enterprise
```
kenrb, line 244: This puts enterprise under "Control", but should it actually be under "Classification"? The presumption is that enterprise federation is already under the control of the admin, and if we can properly classify it then the browser can just get out of the way. But classification is the challenge.

README.md
```diff
@@ -1,85 +1,287 @@
...
+- a firm and principled understanding of where we want to get
+- a pragmatic and flexible understanding of what steps to take us there
+
+We believe a convincing path needs to have a clearly defined end state but also a plausible sequencing strategy.
+
+At the moment, we are actively working with identity providers to test our APIs ([instructions](HOWTO.md)) and help us determine product requirements, ergonomics and deployment strategies that minimize change and maximize control.
+
+Much of this explainer is evolving as a result of this field experimentation.
+
+It is an active area of investigation the order in which these APIs and controlls rollout and the precise time.
```
kenrb, line 275: "controlls" -> "controls"

kenrb (commented):
The structuring here is definitely an improvement, adding some clarity. Separating this as "How to classify federated identity" and "How to promote user control over identity information exchange" seems like the right way to look at it. My comments are mostly pointing out typos but there are a couple of questions about substance.

README.md
```diff
@@ -1,85 +1,287 @@
...
 created: 01/01/2020
-updated: 07/10/2020
+updated: 01/03/2021
 redirect_from: "index.html"
 ---
 
-**TL;DR**; This is an **early exploration** of ways that Browsers can help users of the Web make safer decisions [[1](#the-rp-tracking-problem), [2](#the-idp-tracking-problem)] while logging-in on websites with identity providers with identity-specific APIs.
+> not to be confused with [this](https://www.w3.org/2005/Incubator/webid/spec/) WebID whose authors have [graciously](https://github.com/WICG/WebID/issues/54#issuecomment-783605484) allowed us to use this as a codename until we [find](https://github.com/WICG/WebID/issues/41#issuecomment-712304910) a better one
+
+**TL;DR**; This is an **active** exploration to proactively **preserve** and **elevate** identity federation (e.g. Sign-in with X), **forwards compatible** with a more **private** web.
```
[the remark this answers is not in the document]
samuelgoto: done.

README.md
```diff
@@ -1,85 +1,287 @@
...
 
-Because these general purpose primitives can (by design) be used for an open ended number of use cases (some of them increasingly allowing users to be tracked), browsers have to apply policies that capture the **lowest common denominator**, at best applying cumbersome permissions (e.g. popup blockers) and at worst entirely blocking them.
+Because these general purpose primitives can be used for an open ended number of use cases (again, notably, by design), browsers have to apply policies that capture the **lowest common denominator** of abuse, at best applying cumbersome permissions (e.g. popup blockers) and at worst entirely blocking them (for example, [here](https://github.com/michaelkleber/privacy-model)).
 
-If browsers are applying stricter policies around them, how do we keep identity federation around?
+Over the years, as these low level primitives get abused, browsers intervene and federation adjusts itself. For example, popup blocks became common and federation had to adjust itself to work in a world where popups blockers were widely deployed:
```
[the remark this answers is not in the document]
samuelgoto: done.

README.md
```diff
@@ -1,85 +1,287 @@
...
 
-Before we dig into the solution space, lets take a deeper dive into the problem:
+![](static/mock11.svg)
 
-- [The classification problem](#the-classification-problem)
-- [The RP tracking problem](#the-rp-tracking-problem)
-- [The IDP tracking problem](#the-idp-tracking-problem)
+The challenge, now more than ever, is that some of these low level primitives are getting increasingly abused to allow users on the web to be tracked. So, as a result, browsers are applying strictier and stricter policies around them.
```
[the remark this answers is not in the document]
samuelgoto: done.

README.md
```diff
@@ -1,85 +1,287 @@
...
 
-If the classification problem was solved alone it would delineate the boundaries and applications of the affordances, but it wouldn't necessarily move the needle in making sure that the users are making safer choices.
+![](static/mock23.svg)
 
-In order to prevent cross-site tracking in federation, one has to solve what we've been calling the [RP tracking problem](#the-rp-tracking-problem) in federation and the [IDP tracking problem](#the-idp-tracking-problem).
+Because of the tracking risk, browsers are starting to disable third party cookies in iframes.
 
-# The RP Tracking problem
+Because these cross-site communication takes place in a general purpose medium, it is hard for browsers to distinguish between cross-site communication that is used for exchanging identity data or other cases where intervention is needed.
```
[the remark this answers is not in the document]
samuelgoto: Done. Rephrased this to: 

```
Because these cross-site communication takes place in a general purpose medium, it is hard for browsers to distinguish between cross-site \
communication that is used for exchanging identity data deliberately (e.g. federation) or unintentionally (e.g. tracking).
```

README.md
samuelgoto: Unclear what to call this either, but "sniffing/throttling/filtering/inspecting/gating/interception" doesn't seem right either. I don't think we need to be too attached to names right now: these are going to change substantially and their nature is going to become clearer as we go along. We don't want to use any term that corners us into a box right now, so something abstract/imprecise as "API" seems ok to me. LMK if you disagree strongly, but otherwise resolving.

README.md
samuelgoto: done.

README.md
samuelgoto: done.

README.md
samuelgoto: done

README.md
samuelgoto: It does put Enterprise under "control" because I thought that that was the "implementation" / "specific permission model" rather than the ability to distinguish federation from tracking (hence, not under classification).

Does that make sense?

samuelgoto:
Done, updated my master branch with your feedback. PTAL?

samuelgoto:
Ah, FWIW, there is a build here in case you want to read through what this would look like:

https://code.sgo.to/WebID/README.html

README.md
kenrb: Okay I see what you are saying. I see this as classification because if we can load a page in a special enterprise mode then the user agent won't care if this is tracking or a sign in, so effectively there is an early fork based on whether this looks like: {Sign In | Something Not Sign In | Enterprise Use Case}. For enterprise, it could be a tracker, or could be an identity flow, we are not really concerned.

But maybe that presupposes we will have a signal that allows us to activate some kind of enterprise mode, and that isn't clear because that discussion is still in progress. I'm fine with leaving this for now while that gets clarified.

README.md
kenrb: What about just 'activation'? Because it is being activated by the form of HTTP traffic?

kenrb:
lgtm

A couple of comments still open if I don't mind you merging as is.

samuelgoto:
k, merging and resolving them asynchronously/

PR #57, opened 2021-03-17
fix typo for JWT

ritou:
s/JTW/JWT/

For IPR issue, please set "Mark as non-substantive".


samuelgoto:
`
For IPR issue, please set "Mark as non-substantive".
`

That seems right, but I've never done this before. Can you point me how to accomplish that?

kenrb:
If you click on the IPR error there is a button to mark the PR as non-substantive.

PR #60, opened 2021-04-14
Update mediated API example to match latest iteration

majido:


samuelgoto:
Thanks for the clarification!

PR #70, opened 2021-05-20
Update HOWTO.md

kenrb:
Adding details on the current prototype which has evolved significantly since this was last updated.

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -15,8 +15,8 @@ We will do our best to keep these instructions up to date as protocol and API ch
 It will subsequently roll into downstream channels, but with lag in functionality and bug fixes since this is under development.
 
 ## Enabling WebID in Chrome
-1. Download a version of Chrome with WebID implemented, ideally Canary. ([Link for Canary channel](https://www.google.com/chrome/))
-2. Enable WebID. For Chrome 91+ this can be done directly from chrome://flags. Older versions require manually running with the command line flag --enable-features=WebID. ([Instructions for different platforms.](https://www.chromium.org/developers/how-tos/run-chromium-with-flags))
+1. Download [Chrome Canary](https://www.google.com/chrome/). Since this is a prototype with ongoing development, the implementation in other channels will be out of date with respect to functionality and bug fixes.
+2. Enable WebID. As of Chrome 91 this can be done directly from chrome://flags.
 
 ## Available functionality
 * Automatic permission prompts on [top-level Open ID Connect protocol navigations](navigations.md#the-sign-in-api)
@@ -28,12 +28,12 @@ https://github.com/WICG/WebID/blob/main/cookies.md#logout
 These are described in more detail in the sections below.
 
 ### Permission persistence
-In the flows where permission prompts are shown and the user accepts them, their acceptance is stored in the Chrome user profile as a website setting. The permission prompt will not subsequently be shown for the same RP and IDP pair. There are two distinct permissions, corresponding to the two prompts shown during the above flows.
+In the flows where permission prompts are shown and the user accepts them, their acceptance is stored in the Chrome user profile as a website setting. The permission prompt will **not** subsequently be shown for the same RP and IDP pair. There are two distinct permissions, corresponding to the two prompts shown during the above flows.
 
 At the moment the only way to reset these permissions is to clear browsing data on the profile. Testing can be done in Guest mode or with single-use profiles if permission persistence is not desired.
 
 ## Automatic permission prompts
-When WebID is enabled and a user attempts a federated sign-in using the OpenID Connect protocol over a top-level cross-origin navigation, the navigation request can be blocked by WebID and a permission prompt shown to the user. The same happens with a different prompt when an OIDC response is returned to the relying party. There is a rough heuristic to identify these navigations (at present we test for the presence of `client_id`, `scope` and `redirect_uri` query parameters), so it is possible some requests might be missed.
+When WebID is enabled and a user attempts a federated sign-in using the OpenID Connect protocol over a top-level cross-origin navigation, the navigation request can be blocked by WebID and a permission prompt shown to the user. The same happens with a different prompt when an OIDC response is returned to the relying party. There is [a rough heuristic](README.md#the-permission-oriented-api) to identify these navigations, so it is possible some requests might be missed.
 
 This is intended to illustrate a simple backward-compatible relaxation of web privacy restrictions for identity flows. It has some known limitations, such as behaving poorly in the presence of redirects on the request to the IDP.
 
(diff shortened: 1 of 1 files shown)


kenrb:
Majid can you have a look at the mediation section here?

HOWTO.md
```diff
@@ -8,21 +8,43 @@ updated: 10/12/2020
...
 We will do our best to keep these instructions up to date as protocol and API changes occur.
 
-As of approximately December 11, 2020, it is available on Chrome Canary channel. It will subsequently roll into downstream channels.
+It will subsequently roll into downstream channels, but with lag in functionality and bug fixes since this is under development.
 
 ## Enabling WebID in Chrome
-1. Download a version of Chrome with WebID implemented. ([Link for Canary channel](https://www.google.com/chrome/))
-2. Launch Chrome with the command line flag --enable-features=WebID. ([Instructions for different platforms.](https://www.chromium.org/developers/how-tos/run-chromium-with-flags))
-3. Navigate to a Relying Party page that will exercise the API.
+1. Download a version of Chrome with WebID implemented, ideally Canary. ([Link for Canary channel](https://www.google.com/chrome/))
```
samuelgoto, line 18: I think, at this stage, we should be opinionated about users using canaries. We don't want to support multiple setups, e.g. betas. see point below.

HOWTO.md
```diff
@@ -8,21 +8,43 @@ updated: 10/12/2020
...
+2. Enable WebID. For Chrome 91+ this can be done directly from chrome://flags. Older versions require manually running with the command line flag --enable-features=WebID. ([Instructions for different platforms.](https://www.chromium.org/developers/how-tos/run-chromium-with-flags))
```
samuelgoto, line 19: For example, we can remove the --enable-features=WebID flag from this instrucitons if we assume canaries (i.e. chrome 91+).

HOWTO.md
```diff
@@ -8,21 +8,43 @@ updated: 10/12/2020
...
 
-## Relying Party implementation
+## Available functionality
+* Automatic permission prompts on [top-level Open ID Connect protocol navigations](navigations.md#the-sign-in-api)
```
samuelgoto, line 22: I am planning to move this to here:

https://github.com/wicg/WebID#the-permission-oriented-api

I already started, but was planning to move the entire section there at some point. I can fix this link separately once I do if you want me to.

HOWTO.md
```diff
@@ -8,21 +8,43 @@ updated: 10/12/2020
...
+* `navigator.id.get()` API call under [the permission-oriented flow](navigations.md#the-permission-oriented-variation)
+https://github.com/WICG/WebID/blob/main/cookies.md#logout
+* `navigator.id.get()` API call under [the mediation-oriented flow](navigations.md#the-mediation-oriented-variation)
+* A potential [Logout API](cookies.md#logout)
+
+These are described in more detail in the sections below.
+
+### Permission persistence
+In the flows where permission prompts are shown and the user accepts them, their acceptance is stored in the Chrome user profile as a website setting. The permission prompt will not subsequently be shown for the same RP and IDP pair. There are two distinct permissions, corresponding to the two prompts shown during the above flows.
```
samuelgoto, line 31: **not**

HOWTO.md
```diff
@@ -8,21 +8,43 @@ updated: 10/12/2020
...
+
+At the moment the only way to reset these permissions is to clear browsing data on the profile. Testing can be done in Guest mode or with single-use profiles if permission persistence is not desired.
```
samuelgoto, line 33: Ah, what an interesting question this raises: how does one erase this data? I mean, long term, does this mean that we are going to need to expose these in chrome://settings?

HOWTO.md
```diff
@@ -8,21 +8,43 @@ updated: 10/12/2020
...
+
+## Automatic permission prompts
+When WebID is enabled and a user attempts a federated sign-in using the OpenID Connect protocol over a top-level cross-origin navigation, the navigation request can be blocked by WebID and a permission prompt shown to the user. The same happens with a different prompt when an OIDC response is returned to the relying party. There is a rough heuristic to identify these navigations (at present we test for the presence of `client_id`, `scope` and `redirect_uri` query parameters), so it is possible some requests might be missed.
+
```
samuelgoto, line 37: You can direct people here for the heuritics:

https://github.com/wicg/WebID#the-permission-oriented-api

HOWTO.md
```diff
@@ -37,6 +59,7 @@ async function login() {
     const token = await navigator.id.get({
         provider: "https://idp.example",
         request: "client_id=1234&nonce=Ct60bD&response_type=code&scope=email openid profile",
```
samuelgoto, line 61: blargh, we reallyb need to turn this into a JSON.

HOWTO.md
```diff
@@ -46,57 +69,128 @@ async function login() {
...
+### Permission-oriented variation protocol details
+
+#### Well-Known configuration request
 After the RP initiates a sign-in flow by calling the API, the browser learns about the IdP's WebID support with a fetch to `https://idp.example/.well-known/webid`, where `https://idp.example` was specified as the provider by the RP.
 
 The browser expects a response with MIME type `application/json`, currently containing only one field:<br>
 ```json
 {
-  "idp_endpoint": "https://idp.example/webid/signin"
+  "idp_endpoint": "https://idp.example/webid/idp_endpoint"
```
samuelgoto, line 97: the value here doesn't matter right?

HOWTO.md
```diff
@@ -46,57 +69,128 @@ async function login() {
...
+  "id_token": "ID_token_here"
 }
 ````
 
 2. If the IdP requires user interaction such as sign-in or choosing an account, before issuing a token:<br>
 
 ```json
 { 
-  "signin_url" : "https://idp.example/webid/user_login"
+  "signin_url": "https://idp.example/webid/user_login"
```
samuelgoto, line 119: should this be "signin_endpoint" for consistency? we had this discussion already, but can't remember de outcome.

HOWTO.md
```diff
@@ -46,57 +69,128 @@ async function login() {
...
+ ],
+ "request_id": "xyz123123zyx",
+ "tos": "https://idp.example/tos.html",
+ "privacy_policy": "https://idp.example/privacy.html"
+}
+```
+
+Several of the fields have obvious function.
+
+`enrolled_relying_parties` provides a list of RPs that this account has been used to sign in to before. The browser can determine whether the current RP is on the list and adapt the user experience accordingly.
```
samuelgoto, line 170: This seems like a weird name ... I'm also unclear what precise problem this solves ... I'll follow up over chat.

samuelgoto: Like, for example, what's the trust model for this information?

HOWTO.md
```diff
@@ -46,57 +69,128 @@ async function login() {
...
+  }
+}
+```
+
+`account` and `request_id` are taken from the response of the previous request, and the `request` field contains information from the RP's original identity request.
+
+## Logout API
+The prototype contains a separate API, `navigator.id.logout()` which is being explored as a way to preserve OIDC front-channel logout and SAML Single Signout with loss of access to third-party cookies in embedded contexts. It is intended to replace situations where an IDP logging out a user also must log out the user in RP contexts and would normally do it using iframes with each RP's known logout URL.
+
+The API takes an array of URLs as an argument. For each URL, the browser determines if the user is known to have previously logged in to the RP using that IDP, and if it has, it sends a credentialed GET request to that URL. The determination is currently based on the permission set during an identity request from the RP to the IDP, either via `navigator.id.get()` or from an intercepted OIDC navigation with an automatic permission prompt. That logic should be considered very tentative.
```
samuelgoto, line 192: maybe add an example here?

HOWTO.md
```diff
@@ -46,57 +69,128 @@ async function login() {
...
+  "request": {
+    "client_id": "myclientid",
+    "nonce": "abc987987cba"
+  }
+}
+```
+
+`account` and `request_id` are taken from the response of the previous request, and the `request` field contains information from the RP's original identity request.
+
+## Logout API
```
samuelgoto, line 189: Maybe put this under a bigger section called "session management"?

samuelgoto (commented):
This is a substantial improvement over the existing document, thank you! Some clarification comments inline. There are a couple that caught me by surprise, so I'll follow up over chat.

HOWTO.md
kenrb: This is a better question for @majido but I agree it could use some more discussion.

HOWTO.md
kenrb: We could change it, but a difference is that it doesn't come from the .well-known so it is less of a defined endpoint than the other values; it could change based on the request, for instance.

HOWTO.md
kenrb: No, I just updated it because the endpoints in the mediated flow had them matching and it makes the page more consistent.

HOWTO.md
kenrb: Ack. I'll do that soon.

HOWTO.md
kenrb: Ack

HOWTO.md
kenrb: I brought this up a while back and the answer at the time was that we would integrate into whatever controls privacy sandbox folks want to build for their mechanisms. This needs some kind of management settings page, but that made it sound like it wouldn't be productive to build our own.

samuelgoto:
LGTM

PR #91, opened 2021-10-14
Draft a revocation IDP API

cbiesinger:
Specifies the HTTP requests for the revocation API

### DIFF
--- a/docs/index.bs
+++ b/docs/index.bs
@@ -409,6 +409,7 @@ The [=IDP=] proactively and cooperatively exposes itself as a comformant agent b
 1. An [[#accounts_endpoint]] endpoint
 1. A [[#client_id_metadata_endpoint]] endpoint
 1. An [[#idtoken_endpoint]] endpoint
+1. A [[#revocation_endpoint]] endpoint
 
 <!-- ============================================================ -->
 ## Manifest ## {#well-known}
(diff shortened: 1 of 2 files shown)


docs/index.bs
```diff
@@ -624,7 +628,48 @@ For example:
 ## Revocation ## {#revocation_endpoint}
 <!-- ============================================================ -->
 
-Note: support revocation.
+
+The revocation endpoint is responsible for revoking all ID tokens for the specified client ID for the user.
```
samuelgoto, line 632: linkify "id token" and "client id" to something anchorable.

docs/index.bs
```diff
@@ -624,7 +628,48 @@ For example:
...
+    :   <dfn>account_id</dfn>
+    :: The account identifier for the user (e.g. email).
+    :   <dfn>request</dfn>
+    :: The [=revocation request=]
+</dl>
+
+The <dfn>revocation request</dfn> has the following properties:
+
+<dl dfn-type="argument" dfn-for="revocation_endpoint_request_request">
+    :   <dfn>client_id</dfn>
```
samuelgoto, line 648: Is client_id sufficient? Didn't you also say that we needed the origin?

docs/index.bs
```diff
@@ -624,7 +628,48 @@ For example:
...
+{
+  "account_id": "123",
+  "request": {
+    "client_id": "client1234"
+  }
+}
+```
+</div>
+
+If successful, the response is an empty page with an HTTP 204 code, otherwise
```
samuelgoto, line 671: The response isn't a "page" right? It is a JSON object with a certain scheme, right? Describe the scheme?

docs/index.bs
cbiesinger: Linked ID token but as far as I can tell we have nothing to link client ID to right now :/

cbiesinger:
Also re-ran bikeshed. PTAL.

docs/index.bs
samuelgoto: Should the referer be something that gets passed explicilty rather than implicitly? Isn't the referer header going away at some point?

samuelgoto: Sounds reasonable to me.

PR #93, opened 2021-10-14
[spec] Move spec into spec/ folder.

dj2:
This PR moves the spec file into a spec/ folder. A simple Makefile is
added into spec/ and a top level compile.sh script is added. The
spec is copied into the docs/ folder currently after being built.

The spec is updated to directly include the referenced svg/ images
and the paths are updated to point to `../src/static` as needed.


spec/Makefile
```diff
@@ -0,0 +1,5 @@
+all: index.html
+
+index.html: Makefile index.bs
+	bikeshed --die-on=fatal --allow-nonlocal-files spec index.bs
+	cp index.html ../docs
```
samuelgoto, line 5: Should this be a "mv" instead so that "git status" doesn't show a new file?

PR #104, opened 2021-10-15
Add identity provider argument to revoke() function

cbiesinger:
Revoking a id token needs to know the identity provider URL and
client ID.

***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/104.html" title="Last updated on Oct 20, 2021, 5:21 PM UTC (a367041)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/FedCM/104/b2ceff8...cbiesinger:a367041.html" title="Last updated on Oct 20, 2021, 5:21 PM UTC (a367041)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -252,6 +252,13 @@ which the [=RP=] can use to authenticate the user.
 
     var user;
 
+    const identityProvider = {
+      url: provider,
+      // the pre-registered client id
+      clientId: clientId,
+      nonce: "456"
+    };
+
     async function login(provider, clientId) {
       const {idToken} = await navigator.credentials.get({
         // controls whether auto sign-in is enabled or not
@@ -262,12 +269,7 @@ which the [=RP=] can use to authenticate the user.
         // controls which identity providers to use
         federated: {
           // controls whether the account chooser is modal or not
-          providers: [{
-            url: provider,
-            // the pre-registered client id
-            clientId: clientId,
-            nonce: "456"
-          }],
+          providers: [identityProvider],
         }
       });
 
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -1021,7 +1021,7 @@ This specification extends the {{FederatedCredential}} interface with an extra s
 <xmp class=idl>
 [Exposed=Window, SecureContext]
 partial interface FederatedCredential : Credential {
-  static Promise<void> revoke(USVString accountId);
+  static Promise<void> revoke(USVString clientId, USVString accountId);
```
dj2, line 1024: `clientId, accountId` or `accountId, clientId`? We seem to mark the `accountId` as the higher order bit of information so should it come first? (I say that base off things like the idtoken_endpoint having accountId at the top level of the post body and client_id inside a request sub-object).

cbiesinger: I changed the ordering; the first example here does put the clientid last: https://wicg.github.io/FedCM/#examples

dj2: Out of curiosity, instead of just `clientId` would it make sense to pass a `FederatedIdentityProvider`? Would remove a bit of the confusion of passing the two IDs and differentiating which is which.

spec/index.bs
```diff
@@ -284,8 +286,8 @@ which the [=RP=] can use to authenticate the user.
       await navigator.credentials.preventSilentAccess();
     }
 
-    async function revoke() {
-      await FederatedCredential.revoke(user.id);
+    async function revoke(clientId) {
```
dj2, line 289: Is this `clientId` still needed? Or should it be `identityProvider` ?

dj2 (commented):
That seems like a much clearer API.

spec/index.bs
cbiesinger: Good catch, done.

PR #107, opened 2021-10-18
Continue renaming from WebID to FedCM.

### LINKED ISSUE #97: Rename .well-known/webid?
With the renaming of the spec to FedCM, should the `.well-known/webid` be moved to `.well-known/fedcm` to match or will it continue to have the `webid` path?

### LINKED ISSUE #101: Update Sec-WebID request header name to FedCM name
Possibly going with something like `Sec-FedCM` instead of `Sec-WebID`?

This is forked off from #97

### LINKED ISSUE #102: Update remaining IDL from WebID to FedCM
A few of the entries in the IDL have a `WebID` prefix. These should change to be `FedCM` or something else.

Forked off of #97

dj2:
This PR renames some missed webid's to fedcm. The `.well-known/webid`
location is updated to `.well-known/federated-credentials`. Various examples are updated
to use `fedcm` instead of `webid`.

The IDL files are updated to use `FederatedCredential` instead of `WebID` as a prefix.

Fixes #97
Fixes #101
Fixes #102

***
<a href="https://pr-preview.s3.amazonaws.com/dj2/FedCM/pull/107.html" title="Last updated on Oct 18, 2021, 6:53 PM UTC (5c6d968)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/FedCM/107/bd13566...dj2:5c6d968.html" title="Last updated on Oct 18, 2021, 6:53 PM UTC (5c6d968)">Diff</a>

### DIFF
--- a/explainer/HOWTO.md
+++ b/explainer/HOWTO.md
@@ -81,7 +81,7 @@ _Note that the UI for the mediation-oriented variation is still in development, 
 ### Permission-oriented variation protocol details
 
 #### Well-Known configuration request
-After the RP initiates a sign-in flow by calling the API, the browser learns about the IdP's FedCM support with a fetch to `https://idp.example/.well-known/federated-credentials`, where `https://idp.example` was specified as the provider by the RP.
+After the RP initiates a sign-in flow by calling the API, the browser learns about the IdP's FedCM support with a fetch to `https://idp.example/.well-known/fedcm`, where `https://idp.example` was specified as the provider by the RP.
 
 The browser expects a response with MIME type `application/json`, currently containing only one field:<br>
 ```json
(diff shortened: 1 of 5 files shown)


explainer/HOWTO.md
```diff
@@ -81,19 +81,19 @@ _Note that the UI for the mediation-oriented variation is still in development,
 ### Permission-oriented variation protocol details
 
 #### Well-Known configuration request
-After the RP initiates a sign-in flow by calling the API, the browser learns about the IdP's FedCM support with a fetch to `https://idp.example/.well-known/webid`, where `https://idp.example` was specified as the provider by the RP.
+After the RP initiates a sign-in flow by calling the API, the browser learns about the IdP's FedCM support with a fetch to `https://idp.example/.well-known/federated-credentials`, where `https://idp.example` was specified as the provider by the RP.
```
samuelgoto, line 84: can we use the shorter version of `.well-known/fedcm` instead? was there a particular reason you used the expanded version?

samuelgoto: That's a reasonable point, but seems like it would be applicable elsewhere too (i.e. coherent, but inconsistently applied?).

FWIW, `.well-known` is public facing but mostly machine-readable, not human-readable.

I'd start with a change to `fedcm` right now for inconsistency in other places we use (e.g. the X-FedCM-XSRF) and then move this to a more human readablae one once we come up with the principles that helps us reason between them.

WDYT?

dj2: That's fair, is the common case for `.well-known` to be lower case? Would `.well-known/FedCM` work? Or should we stick with `.well-known/fedcm`?

I don't think renaming the security token would be good, that would make for a long header name `Sec-FederatedCredential-CSRF` is long, heh. But, a consistent story is a good idea.

Issue #121, opened 2021-10-22
Rational for having FederatedCredential.revoke(accountId) be a static method?


dj2:
The `revoke` method seems like it makes sense to call on the credential itself instead of as a static method. (Won't we also need things like the `provider` and `client_id` to make the call?

If feels like you're saying, I want to revoke this credential. But instead of saying that we're saying, hey revoke stuff for this `account_id` which seems like it's up a level?

cbiesinger:
https://github.com/WICG/FedCM/pull/139

Please take a look!

dj2:
`revoke` is no-longer a `static`. Closing.

Issue #138, opened 2021-11-03
Example 20 is inconsistent with IDL definitions


cbiesinger:
https://wicg.github.io/FedCM/#browser-api-idp-sign-in

```
  const credential = new FederatedCredential({
    id: id,
    provider: 'https://idp.example',
    name: name,
    iconURL: iconUrl,
    client_id: "1234",
  });
```

But FederatedCredential does not have a client_id field:
https://wicg.github.io/FedCM/#idl-index
https://w3c.github.io/webappsec-credential-management/#federatedcredential
https://w3c.github.io/webappsec-credential-management/#credential

https://w3c.github.io/webappsec-credential-management/#credentialuserdata

cbiesinger:
Additionally, FedCM really should add idToken and approvedBy to FederatedCredentialInit, though that won't make a difference for this example.

cbiesinger:
@dj2 fyi

dj2:
There is currently a mis-match in the spec. I think that example should pass a `FederatedIdentityProvider` instead of the `provider` string. The provider string is what's in the Credential Mangement FederatedCredential but we over-ride it to make it `string | FederatedIdentityProvider`. Then `client_id` would move into that object.

I think? Would have to check the IDL a bit more to know if there are other changes needed.

cbiesinger:
> we over-ride it to make it string | FederatedIdentityProvider

Where do we do that...? I only see it in FederatedCredentialRequestOptions, not in FederatedCredential or FederatedCredentialInit

dj2:
Oh, we did it in the RequestOptions only. I think we _also_ need to do it in FederatedCredential. My preference would be to just remove the `string` option as do we want to support that? My guess it was left off of `FederatedCredential` as the current spec doesn't technically return a `credential` object we just return the `idToken` so...why add it. But we want it now, heh, and hopefully will evolve the spec such that we're returning the credential object.

I think both init and credential should store the FederatedIdentityProvider.

(I also wonder if our `FederatedIdentityProvider` should have a `protocol` string in it as the CredentialManagement FederatedCredential has.

dj2:
@cbiesinger can this be closed with the new API updates?

samuelgoto:
I think this problem goes away when [this](https://github.com/fedidcg/FedCM/pull/207) gets merged.

cbiesinger:
The initial issue is fixed but not this one:

[...]

PR #139, opened 2021-11-03
Make revoke() a non-static method

cbiesinger:
Fixes: #121

***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/139.html" title="Last updated on Nov 3, 2021, 10:08 PM UTC (b6204db)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/FedCM/139/2ae2dc6...cbiesinger:b6204db.html" title="Last updated on Nov 3, 2021, 10:08 PM UTC (b6204db)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -294,7 +294,11 @@ which the [=RP=] can use to authenticate the user.
     }
 
     async function revoke() {
-      await FederatedCredential.revoke(user.id, identityProvider);
+      await new FederatedCredential({
+          id: user.id,
+          provider: identityProvider.url,
+          client_id: identityProvider.clientId,
+        }).revoke();
     }
     </script>
   </body>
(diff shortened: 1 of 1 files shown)


dj2:
This change LGTM. As pointed out in the issue, the examples aren't quite right because our `FederatedCredential` isn't quite right, but these examples match `Example 20` in the spec and so I think we could land this and then do a followup to update `FederatedCredential` to have a `FederatedIdentityProvider` and then use that to store the `provider` and `client_id`.

dj2 (approved):

cbiesinger:
("the issue" Dan is referring to is #138 )

samuelgoto:
Is this PR now outdated given @dj2 's recent changes to the v1.5 API? Can I close this?

cbiesinger:
Yep

PR #140, opened 2021-11-04
Add background to SVG images.

### LINKED ISSUE #66: Dark mode
The explainer is unreadable when GitHub is in dark mode.  Consider adding a background to the images.

dj2:
This PR sets the background of the SVG images to white so they are
readable when using dark mode to view the explainer.

Fixes #66


samuelgoto:
Ugh, isn't there a way to solve this problem without having to change all of the SVGs? Wouldn't it be possible to create a CSS class somewhere where the background-color is set for all images?

I generate a good amount of diagrams as I go along, and chances are I / one of us will forget of painting the background.

dj2:
I fixed it in the spec with CSS, but the explainer is all MD files rendered by github which, as far as I know, don't have CSS. Without rendering the explainer files to HTML, this is the only way I know how to fix it (and yes, I agree it sucks)

PR #141, opened 2021-11-04
Update introduction and abstract.

### LINKED ISSUE #90: Abstract and Introduction in the Spec are unnecessarily broad
The current text in the abstract and introduction in the current spec that we are working on are unnecessarily broad:

https://wicg.github.io/FedCM/

> This specification defines a set of high-level APIs that enables users to continue to use Identity Providers to authenticate to Relying Partys without incurring into Unsanctioned Web Tracking. It accomplishes that by exposing the explicit user controls needed to manage the lifecycle of their federated accounts.

A few ways we can narrow this description down:

- to be precise about offering alternatives to federated sign in flows that break without third party cookies (i.e. that this spec is offered *in replacement of* third party cookies)
- to be precise about the non-mutually exclusive optionality of existing federated sign in flows with decorated links (i.e. that this spec is offered *in addition* to what currently works with decorated links -- whether/when/if it gets mitigated is left as a separate concern)
- to be precise about the narrow scope of account management (e.g. an account chooser) and session management (e.g. front channel logout)

dj2:
This PR takes a pass at updating the intro and abstract to make them a
bit more focused and concrete.

Fixes #90


spec/index.bs
```diff
@@ -16,7 +16,8 @@ Text Macro: TRUE <code>true</code>
 Text Macro: RP Relying Party
 Text Macro: IDP Identity Provider
 
-Abstract: This specification defines a set of [=high-level API=]s that enables users to continue to use [=Identity Provider=]s to authenticate to [=Relying Party=]s without incurring into [[UNSANCTIONED-TRACKING]]. It accomplishes that by exposing the explicit user controls needed to manage the lifecycle of their federated accounts.
+Abstract: This specification defines a set of primitives that enable federated identity without enabling [[UNSANCTIONED-TRACKING]].
```
samuelgoto, line 19: [[UNSACTIONED_TRACKING]] is somewhat broad and fuzzy, so I'd avoid making a reference to it in the Abstract. Also, not ALL unsactioned tracking is going to be prevented with what's currently specified (e.g. see the lack of protections against global identifiers). I'd narrow it down to something along the lines of things that are strictly needed / dependent on third party cookies as it gets deprecated

spec/index.bs
```diff
@@ -77,19 +78,26 @@ div.image {
 
 *This section is non-normative.*
 
-Over the last decade, identity federation has unquestionably played a central role in raising the bar for authentication on the web, in terms of ease-of-use (e.g. passwordless single sign-on), security (e.g. improved resistance to phishing and credential stuffing attacks) and trustworthiness compared to its preceding pattern: per-site usernames and passwords.
-
-The standards that define how identity federation works today on the Web were built independently of the Web Platform (namely, [[SAML]], [[OpenID]] and [[OAuth]]), and their designers had to (rightfully so) work around its limitations rather than extend them.
-
-Because of that, existing user authentication flows were designed on top of general-purpose web platform capabilities such as top-level navigations/redirects with parameters, window popups, iframes and cookies.
+The web is evolving. That evolution has caused, and is causing,
```
samuelgoto, line 81: This first paragraph feels a bit off / duplicative to start with. Revisit it? Or maybe skip as the next one points to the [[PRIVACY-MODEL]] already as well as the evolution?

spec/index.bs
```diff
@@ -77,19 +78,26 @@ div.image {
...
+changes to the underlying [[PRIVACY-MODEL]] of the web platform.
 
-However, because these general purpose primitives can be used for an open ended number of use cases (again, notably, by design), browsers have to apply policies that capture the lowest common denominator of abuse, at best applying cumbersome permissions (e.g. popup blockers) and at worst entirely blocking them (e.g. blocking third party cookies).
+With the evolution of the [[PRIVACY-MODEL]], fundamental assumptions of the web
```
samuelgoto, line 84: One of the challenges with using the [[PRIVACY-MODEL]] up front in the spec here is that I think it is not a threat model that all browsers are bought into quite yet, so it is a presupposition that may read as a non-starter for other browsers.

samuelgoto: Maybe if you pointed to all of the ways in which all browsers are evolving (there are links to safari ITP and firefox's ETP), that would address it.

spec/index.bs
```diff
@@ -77,19 +78,26 @@ div.image {
...
+are being redefined or removed. Third-party cookies are one of those
+assumptions being removed. While overall good for the web,
+the third-party cookie deprecation removes fundamental building blocks
+in how some existing technologies were designed and deployed.
 
-Over the years, as these [=low-level API=]s get abused, browsers intervene and federation adjusts itself. For example, popup blockers became common and federation had to adjust itself to work in a world where popups blockers were widely deployed.
+Historically, federated login has relied on third-party cookies or
+navigational redirects as the building blocks. The Federated Credentials
+Management API aims to replace the federated login use of third-party
+cookies without enabling [[UNSANCTIONED-TRACKING]]. The API provides the
```
samuelgoto, line 93: Again, UNSACTIONED-TRACKING is probably too broad of a statement, there is probably something more specific to call out here.

dj2:
PTAL, updated and, hopefully, clarified.

spec/index.bs
```diff
@@ -74,19 +75,25 @@ dl.domintro dt code {
 
 *This section is non-normative.*
 
-Over the last decade, identity federation has unquestionably played a central role in raising the bar for authentication on the web, in terms of ease-of-use (e.g. passwordless single sign-on), security (e.g. improved resistance to phishing and credential stuffing attacks) and trustworthiness compared to its preceding pattern: per-site usernames and passwords.
-
-The standards that define how identity federation works today on the Web were built independently of the Web Platform (namely, [[SAML]], [[OpenID]] and [[OAuth]]), and their designers had to (rightfully so) work around its limitations rather than extend them.
-
-Because of that, existing user [=authentication=] flows were designed on top of general-purpose web platform capabilities such as top-level navigations/redirects with parameters, window popups, iframes and cookies.
+With the evolution of the [[PRIVACY-MODEL]], fundamental assumptions of the web
```
samuelgoto, line 78: The [[PRIVACY-MODEL]] isn't a model that I think is generally accepted as the "right" privacy model across all browsers, so, I'm a bit leery leading the introduction with it.

spec/index.bs
```diff
@@ -74,19 +75,25 @@ dl.domintro dt code {
...
 
-However, because these general purpose primitives can be used for an open ended number of use cases (again, notably, by design), browsers have to apply policies that capture the lowest common denominator of abuse, at best applying cumbersome permissions (e.g. popup blockers) and at worst entirely blocking them (e.g. blocking third party cookies).
+The Federated Credential Management API aims to bridge the gap for the
+federated identity designs which relied on third-party cookies.
+The API provides the primitives needed to support federated credentials, from
+sign-in to sign-out and revocation.
 
-Over the years, as these [=low-level API=]s get abused, browsers intervene and federation adjusts itself. For example, popup blockers became common and federation had to adjust itself to work in a world where popups blockers were widely deployed.
+In order to provide the federated identity primitives without the use of
+third-party cookies the API places the UA as a mediator between [=RPs=]
```
samuelgoto, line 90: Either linkify [=UA=] or expand it to be User Agent.

spec/index.bs
```diff
@@ -74,19 +75,25 @@ dl.domintro dt code {
...
+With the evolution of the [[PRIVACY-MODEL]], fundamental assumptions of the web
+platform are being redefined or removed. Access to cookies in a third-party
+context are one of those assumptions. While overall good for the
+web, the third-party cookie deprecation removes a fundamental building block
+used by certain designs of federated identity.
 
-However, because these general purpose primitives can be used for an open ended number of use cases (again, notably, by design), browsers have to apply policies that capture the lowest common denominator of abuse, at best applying cumbersome permissions (e.g. popup blockers) and at worst entirely blocking them (e.g. blocking third party cookies).
+The Federated Credential Management API aims to bridge the gap for the
+federated identity designs which relied on third-party cookies.
+The API provides the primitives needed to support federated credentials, from
```
samuelgoto, line 86: "The API provides the primitives needed to support federated credentials, from
sign-in to sign-out and revocation."

=>

"The API provides the primitives needed to support federated identity when/where it depends on third party cookies, from
sign-in to sign-out and revocation."

spec/index.bs
samuelgoto: I think a variation of what you have in the explainer works best: 

```
As the web has evolved there have been ongoing privacy-oriented changes ([example](chrome's), [example](safari's), etc) and underlying privacy principles ([example](PRIVACY-MODEL)). 
```

spec/index.bs

spec/index.bs

samuelgoto (changes requested):
Preview helped me review this

Issue #145, opened 2021-11-06
Shape of the FedCM API


dj2:
The question of the current shape of the FedCM API has come up in #121 and #118 along with a few dicussions. Opening this issue with an sketch of the updated API in order to get feedback and opinions.

This API is high-level. The API is what I think we'd like it to be, it may be too high level but starting here means we can incrementally deprecate portions and move to a lower level. If we start with a lower level api we won't be able to move to high level with breaking adopters. I think this gives a good amount of flexibility to adapt in the future.

# RP Methods
## Get
We move back to a `get` request that matches what the Credential Management API suggests. Creating and returning a `FederatedCredential` object which stores the `provider` and selected `account_id` for the user. This will allow the RP to access and store the `account_id`.

```javascript
const creds = await navigator.credentials.get({
  providers: [{
    url: "https://idp.example"
  }],
  mediation: "silent",
});
```

The credentials will either be returned from the credential store (if `store` had previously been called on them) or the user will go through the authorization flow.

## FetchTokens
In order to get a token the user calls `fetchTokens()`. This method takes one parameter the `FetchTokensOptions` which maybe `undefined` or an object. The object has one field:

```javascript
FetchTokensOptions {
  tokens: <string>
}
```
The `string` can currently be `id_token`. If `undefined` then the object defaults to `{tokens: "id_token"}`.

```javascript
const tokens = await creds.fetchTokens();
```
The return is a hash where the keys contain the tokens:
```javascript
{
  "id_token": "asdfasdf"
}
```

## Revoke
The call to `revoke` is done on the credentials object. Because the RP can create the object _without_ having used FedCM for the account signup then the UA must only send the revoke to the IDP if the user has used FedCM to authorize previously. Otherwise the method returns silently without doing anything.

```javascript
await creds.revoke();
```
**Question**: If the user hasn't logged in, `navigator.credentials.get().revoke()` would trigger an auth flow which is weird ....

## Logout
The `logout` call is a high level call to sign out of the RP. It is required in the UA to call `navigator.credentials.preventSilentAccess()` along with any other session updates required.

```javascript
await navigator.credentials.get().logout()
```

**Question**: If the user hasn't logged in, this would trigger an auth flow which is weird ....

# IDP Logout
The call from the IDP to `logout` is done directly against the `FederatedCredentials` object and provides the RP URLs and account_ids to logout. The call does not return. The UA will only send the logout call to the RP if the user selected during sign-in to allow the IDP to log them out (they have the permission bit set).

[...]

PR #150, opened 2021-11-18
Update the FedCM API.

dj2:
This PR updates the FedCM API to be a closer match to the Credential
Management Level 1 API. The `get` method is changed to return the
`credential` instead of an `id_token`. A new `fetchTokens` method is
added in order to retrieve tokens. The `preventSilentAccess` is moved
behind an explicit `logout` API call. The IDP `logout` is renamed to
`idpLogout` to remove confusion.

Issue #145

***
<a href="https://pr-preview.s3.amazonaws.com/dj2/FedCM/pull/150.html" title="Last updated on Nov 18, 2021, 3:07 PM UTC (40079b4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/WICG/FedCM/150/b100996...dj2:40079b4.html" title="Last updated on Nov 18, 2021, 3:07 PM UTC (40079b4)">Diff</a>


dj2:
In a discussion with Sam, this PR raises some questions that we need to answer. This PR sets the account_id into the FederatedCredential which allows the RP to access the account_id. This is providing more information to the RP than we originally planned.

Question 1, is this exposure of account_id acceptable? The account_id maybe PII which is an issue. If the account_id is the email, do the IDPs expect us to be giving that information to the RP? Does the User?

Question 2, if exposing the account_id is not acceptable, how do we build an ID for the credential? We currently use the account_id in a few places which means we need to know which account the user selected when they signed up. 

Taking a step back, what do we need to know and when do we need to know it.

During sign-up, we get the account list and present to the user. This list includes an account id which is seen by the browser but not the RP. If the user accepts sign-up we create a `FederatedCredential` object. That object needs an ID. This ID will be visible to the PR in the credential object.

Question 3, does having a stable ID in the credential object cause an issue? If we set `FederatedCredential.id = SHA256(account_id)` would that be an issue?

In order to fetch an idToken, we pass the `account_id`. If we decouple that idToken fetch from the initial `get()` of the credential then we have to know the account_id at a later point in the flow. In this case the browser needs to store it if it isn’t exposed to the RP. (Having the id as the SHA is more of a determent here as then the IDP has to lookup the sha’d version of the account_id). The existing `id_token` endpoint requires the `account_id` but it works because it’s only called as part of the sign-up flow when we have the ID. 

The IDP Sign-out currently uses the `account_id` (we never spec’d what the logout URL looks like on the RP side). If our credential id is the SHA then we can retrieve the correct credential object to logout. If the ID is a random value the browser needs to know how that random value maps to a given account_id. What does the RP expect to receive here? What do they need to know?

The `revocation_endpoint` as currently spec’d also requires the `account_id`. This would not be available to the RP, so does the browser have to store that ID in storage somewhere?

Where do we fit refresh tokens in here? We originally modelled off the implicit flow of OIDC which does not provide refresh_tokens. We’re almost at an authorization model where we split out the idToken fetch, but that gives us an authorization `code` which we can exchange. Still leaves the question of what do we do for the `ID` and how do we store the refresh_token.

Question 4, do we expect an idToken to _always_ be fetched? Is it a required return value? How stable is it, does it change? If the idToken is our FederatedCredential.id is that an issue (if it changes it probably won't work very well)

Apetree100122 (commented):
Keep doing well

PR #151, opened 2021-11-21
Make clientId and nonce required fields in FederatedIdentityProvider

yi-gu:


dj2 (approved):

samuelgoto:
Thanks Yi (and Dan for reviewing!)!

PR #165, opened 2022-01-11
docs: update new repository links

konojunya:
It's a new repository, so I fixed the link.


dj2:
The change LGTM but it looks like the IRP check bot doesn't run after the repo move. We'll need to verify you've signed the W3C IPR before the contribution can be accepted (I'll see if I can figure that out in the morning).

samuelgoto:
Seems non-substantive enough that I'll bypass the IPR checks. Merging.

samuelgoto:
Thanks @konojunya!

PR #168, opened 2022-01-12
Update client metadata example to use absolute urls

yi-gu:
The client metadata endpoint is on idp.example so using relative urls could be confusing. This PR updates the example to use the absolute urls with the rp domain to reduce ambiguity.


dj2 (approved):

cbiesinger:
lgtm although I am confused why making nonce optional is listed as a commit for this PR

dj2:
I'm guessing he branched off a PR to add nonce optional; nonce optional merged to main; he merged main into this branch. 

We squash and merge so it shouldn't matter as the intervening commits will all go away.

PR #170, opened 2022-01-13
Use form url encoding as opposed to JSON to mint idtokens

samuelgoto:


samuelgoto:
@yi-gu did I get this right?

samuelgoto:
on a related note, @cbiesinger should I also update the spec in the revocation API to use form encoding too as opposed to JSON?

dj2 (approved):

samuelgoto:
sgtm, will send another patch for the revocation API.

PR #171, opened 2022-01-13
Move the revocation endpoint to use x-www-form-urlencoded

samuelgoto:


samuelgoto:
@cbiesinger does this seem right?

cbiesinger (approved):
lgtm

PR #173, opened 2022-01-26
Add consent_acquired to the spec as part of the id_token endpoint req…

samuelgoto:


yi-gu (approved):
Do we need a separate issue to track an alternative name? Or the existing crbug is sufficient? Otherwise LGTM.

samuelgoto:
> Do we need a separate issue to track an alternative name? Or the existing crbug is sufficient? Otherwise LGTM.

I'll track that in the crbug.

PR #174, opened 2022-01-26
Document the branding options

samuelgoto:
I'm taking a pass at documenting what has been implemented, which may or may not be ideal. We can take another pass, separately, at discussing whether this is the right approach or not, but I'd like to start by making sure that the spec is consistently representing the implementation.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -548,6 +548,8 @@ The response is expected to have the following properties:
 <dl dfn-type="argument" dfn-for="accounts_endpoint_response">
     :   <dfn>accounts</dfn> (required)
     ::  A list of [=Account JSON=].
+    :   <dfn>branding</dfn> (optional)
+    ::  A set of [=Branding JSON=] options.
 </dl>
 
 Every <dfn>Account JSON</dfn> is expected to have the following properties:
@@ -563,6 +565,26 @@ Every <dfn>Account JSON</dfn> is expected to have the following properties:
     ::  The user's email address.
     :   <dfn>picture</dfn> (optional)
     :: The user's profile picture.
+</dl>
+
+The <dfn>Branding JSON</dfn> is expected to have the following properties:
+
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_branding">
+    :   <dfn>background_color</dfn> (optional)
+    ::  The background color of the button.
+    :   <dfn>foreground_color</dfn> (optional)
+    ::  The foreground color of the button.
+    :   <dfn>icons</dfn> (optinal)
+    ::  A list of [=Icon JSON=] objects.
+</dl>
+
+The <dfn>Icon JSON</dfn> is expected to have the following properties:
+
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_branding_icons">
+    :   <dfn>url</dfn> (required)
+    ::  The url pointing to the icon inmage.
+    :   <dfn>size</dfn> (optional)
+    ::  The size of the icon (assumed to be squared).
 </dl>
 
 For example:
@@ -582,7 +604,15 @@ For example:
    "name": "Johnny",
    "email": "johnny@idp.example",
    "picture": "https://idp.example/profile/456"
-  }]
+  }],
+  "branding": {
+   "background_color": "green",
+   "foreground_color": "0xFFEEAA",
+   "icon": [{
+     "url": "https://idp.example/icon.ico",
+     "size": 10
+   }],
+ }
 }
 ```
 </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -582,7 +582,7 @@ The <dfn>Icon JSON</dfn> is expected to have the following properties:
 
 <dl dfn-type="argument" dfn-for="accounts_endpoint_response_branding_icons">
     :   <dfn>url</dfn> (required)
-    ::  The url pointing to the icon inmage.
+    ::  The url pointing to the icon image.
     :   <dfn>size</dfn> (optional)
     ::  The size of the icon (assumed to be squared).
 </dl>
@@ -608,7 +608,7 @@ For example:
   "branding": {
    "background_color": "green",
    "foreground_color": "0xFFEEAA",
-   "icon": [{
+   "icons": [{
      "url": "https://idp.example/icon.ico",
      "size": 10
    }],


spec/index.bs
```diff
@@ -565,6 +567,26 @@ Every <dfn>Account JSON</dfn> is expected to have the following properties:
...
+    ::  The foreground color of the button.
+    :   <dfn>icons</dfn> (optinal)
+    ::  A list of [=Icon JSON=] objects.
+</dl>
+
+The <dfn>Icon JSON</dfn> is expected to have the following properties:
+
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_branding_icons">
+    :   <dfn>url</dfn> (required)
+    ::  The url pointing to the icon inmage.
```
yi-gu, line 585: nit:s/inmage/image

spec/index.bs
```diff
@@ -582,7 +604,15 @@ For example:
    "name": "Johnny",
    "email": "johnny@idp.example",
    "picture": "https://idp.example/profile/456"
-  }]
+  }],
+  "branding": {
+   "background_color": "green",
+   "foreground_color": "0xFFEEAA",
+   "icon": [{
```
yi-gu, line 611: nit:s/icon/icons

yi-gu (approved):
LGTM % nits

spec/index.bs
samuelgoto: done

spec/index.bs
samuelgoto: ah, thanks, done!

spec/index.bs
```diff
@@ -565,6 +567,26 @@ Every <dfn>Account JSON</dfn> is expected to have the following properties:
     :: The user's profile picture.
 </dl>
 
+The <dfn>Branding JSON</dfn> is expected to have the following properties:
+
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_branding">
+    :   <dfn>background_color</dfn> (optional)
+    ::  The background color of the button.
```
dj2, line 574: What are the valid values for background and foreground? Anything that is a valid colour in CSS? (Could I say`rgb(0, 0, 1)` as the colour?

spec/index.bs
```diff
@@ -565,6 +567,26 @@ Every <dfn>Account JSON</dfn> is expected to have the following properties:
...
+    :   <dfn>foreground_color</dfn> (optional)
+    ::  The foreground color of the button.
+    :   <dfn>icons</dfn> (optinal)
+    ::  A list of [=Icon JSON=] objects.
+</dl>
+
+The <dfn>Icon JSON</dfn> is expected to have the following properties:
+
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_branding_icons">
+    :   <dfn>url</dfn> (required)
```
dj2, line 584: Is this required to be a full URL? Or could I say "/icon.png"?

What formats do we allow for the image itself? png, ico, jpg?

spec/index.bs
```diff
@@ -565,6 +567,26 @@ Every <dfn>Account JSON</dfn> is expected to have the following properties:
...
+    ::  The url pointing to the icon image.
+    :   <dfn>size</dfn> (optional)
+    ::  The size of the icon (assumed to be squared).
```
dj2, line 587: s/assumed to be/must be/ ?

spec/index.bs
```diff
@@ -565,6 +567,26 @@ Every <dfn>Account JSON</dfn> is expected to have the following properties:
...
+    :   <dfn>icons</dfn> (optinal)
+    ::  A list of [=Icon JSON=] objects.
+</dl>
+
+The <dfn>Icon JSON</dfn> is expected to have the following properties:
+
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_branding_icons">
+    :   <dfn>url</dfn> (required)
+    ::  The url pointing to the icon image.
+    :   <dfn>size</dfn> (optional)
```
dj2, line 586: What size is used if this isn't provided? Is it taken from the icon itself? Is there a minimum or maximum?

spec/index.bs
cbiesinger: Yes, CSS colors. (we use https://source.chromium.org/chromium/chromium/src/+/main:content/common/color_parser.cc;l=60?q=ParseCssColorString&sq=&ss=chromium, so rgb() and hsl() work but e.g. rgba() or lab() does not.)

spec/index.bs
cbiesinger: The size is only used to find which of the icons is the best fit. We only download the one icon we pick, so this is not taken from the icon itself. Implementation-wise, we pick whatever FindBestMatchingSquareIcon returns (https://source.chromium.org/chromium/chromium/src/+/main:content/browser/webid/idp_network_request_manager.cc;l=231?q=idp_network_req&ss=chromium)

spec/index.bs
samuelgoto: Any chance you'd have a link handy for the CSS ref that defines "CSS colors"? I found [1] but it doesn't seem to match the implementation that you sent [2].

Any chance you'd know where that's defined?

[1] https://www.w3.org/TR/css-color-4/#typedef-absolute-color-base
[2] https://source.chromium.org/chromium/chromium/src/+/main:content/common/color_parser.cc;l=70?q=ParseCssColorString&ss=chromium

samuelgoto: I sent this PR here with an enumeration of what we currently support:

https://github.com/fedidcg/FedCM/pull/177

Resolving this comment now and picking up from there.

spec/index.bs
samuelgoto: Do we need to specify what image formats the spec supports? Can't we leave that unspecified?

spec/index.bs
samuelgoto: Wouldn't it be appropriate for an implementation to "assume" that it is "squared" (e.g. crop it squarely)?

spec/index.bs
dj2: If it's unspecified, how does an IDP know what they can return? It seems like it would be more interoperable to set a specific list?

spec/index.bs
dj2: The implementation cropping is not how I read the above. (I'm a fan of explicit instead of assumed). We could also say "size of the icon, will be cropped to be square" or something.

cbiesinger: That would be a confusing phrasing since this is a single integer. Also, we don't use this integer to actually size the final image.

PR #175, opened 2022-01-26
Dont use well-known for the Manifest file

samuelgoto:


dj2 (approved):
LGTM. Has this change landed in Chrome already and rolled out? Or should we hold off landing this until it's in Chrome? (Or should we list both and point out what Chrome version it will hit stable?)

cbiesinger (approved):
lgtm but could you link to https://github.com/fedidcg/FedCM/issues/160 ?

cbiesinger:
@dj2 not in chrome yet

samuelgoto:
> LGTM. Has this change landed in Chrome already and rolled out? Or should we hold off landing this until it's in Chrome? (Or should we list both and point out what Chrome version it will hit stable?)

I'll kick off a crbug to track the implementation.

PR #176, opened 2022-01-26
Document Sec-FedCM-CSRF

cbiesinger:


dj2 (approved):

spec/index.bs
```diff
@@ -760,6 +760,11 @@ otherwise an HTTP error code.
 ## <code><dfn data-export="">`Sec-FedCM-CSRF`</dfn></code> ## {#Sec-FedCM-CSRF}
 <!-- ============================================================ -->
 
+All FedCM HTTP requests sent by the browser must contain a header
+`Sec-FedCM-CSRF` with a random number as the value. This allows servers
```
samuelgoto, line 764: There is a <dfn> here, could you please linkify?

https://fedidcg.github.io/FedCM/#Sec-FedCM-CSRF

cbiesinger: I think you're missing a word? Also that is the section I'm changing?

samuelgoto: Ah, of course. Duh, done.

PR #179, opened 2022-01-27
Document which requests need a Referer header

cbiesinger:


spec/index.bs
```diff
@@ -529,8 +529,9 @@ For example:
 The accounts list endpoint provides the list of accounts the user has at the [=IDP=].
 
 The accounts list endpoint is fetched
-(a) **with** [=IDP=] cookies and
-(b) **with** a special [[#Sec-FedCM-CSRF]] header.
+(a) **with** [=IDP=] cookies,
+(b) **with** a special [[#Sec-FedCM-CSRF]] header and
+(c) **without** a Referer header.
```
samuelgoto, line 534: Can you (a) define Referer somewhere, make it explicit that it refers only to the orign and (b) and link to it here?

cbiesinger: Updated, does that look better?

dj2 (approved):

PR #181, opened 2022-02-02
Add approved_clients to the spec

samuelgoto:


yi-gu (approved):

spec/index.bs
```diff
@@ -565,8 +565,8 @@ Every <dfn>Account JSON</dfn> is expected to have the following properties:
     ::  The user's given name.
     :   <dfn>email</dfn> (optional)
     ::  The user's email address.
-    :   <dfn>picture</dfn> (optional)
```
cbiesinger, line 568: did you intentionally remove that...?

spec/index.bs
```diff
@@ -602,12 +602,14 @@ For example:
    "name": "John Doe",
    "email": "john_doe@idp.example",
    "picture": "https://idp.example/profile/123",
+   "approved_clients": ["123", "456", "789"],
   }, {
    "account_id": "5678",
    "given_name": "Johnny",
    "name": "Johnny",
    "email": "johnny@idp.example",
    "picture": "https://idp.example/profile/456"
```
cbiesinger, line 611: because if you did you should remove it here too

spec/index.bs
samuelgoto: ooops, no, that was a bug. sorry, will revert this specific part.

PR #182, opened 2022-02-02
Move the branding configuration to the fedcm.json object and make it …

samuelgoto:
…match CSS terminology

Does anyone know what's Peter's github handle for me to add him as a reviewer?

### DIFF
--- a/explainer/HOWTO.md
+++ b/explainer/HOWTO.md
@@ -118,7 +118,15 @@ containing four fields:
   "accounts_endpoint": "https://idp.example/fedcm/accounts_endpoint",
   "client_id_metadata_endpoint": "https://idp.example/fedcm/client_id_metadata_endpoint",
   "id_token_endpoint": "https://idp.example/fedcm/token_endpoint",
-  "revocation_endpoint": "https://idp.example/fedcm/revocation_endpoint"
+  "revocation_endpoint": "https://idp.example/fedcm/revocation_endpoint",
+  "theme": {
+   "background-color": "green",
+   "color": "0xFFEEAA",
+   "icons": [{
+     "url": "https://idp.example/icon.ico",
+     "size": 10
+   }],
+  },
 }
 ```
 
--- a/explainer/HOWTO.md
+++ b/explainer/HOWTO.md
@@ -119,12 +119,13 @@ containing four fields:
   "client_id_metadata_endpoint": "https://idp.example/fedcm/client_id_metadata_endpoint",
   "id_token_endpoint": "https://idp.example/fedcm/token_endpoint",
   "revocation_endpoint": "https://idp.example/fedcm/revocation_endpoint",
-  "theme": {
+  "branding": {
    "background-color": "green",
    "color": "0xFFEEAA",
    "icons": [{
      "url": "https://idp.example/icon.ico",
-     "size": 10
+     "size": 10,
+     "background-brightness": "light"
    }],
   },
 }
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -507,28 +507,34 @@ The file is parsed expecting the following properties:
     ::  A URL that points to an HTTP API that complies with the [[#idp-api-id-token-endpoint]] API.
     :   <dfn>revocation_endpoint</dfn>
     :: A URL that points to an HTTP API that complies with the [[#idp-api-revocation-endpoint]] API.
-    :   <dfn>theme</dfn> (optional)
-    ::  A set of [=Theme JSON=] options.
+    :   <dfn>branding</dfn> (optional)
+    ::  A set of [=Branding JSON=] options.
 </dl>
 
-The <dfn>Theme JSON</dfn> is expected to have the following properties:
+The <dfn>Branding JSON</dfn> allows an [=IDP=] to express their branding preferences, which can be used by User Agents in customization entry points.
 
-<dl dfn-type="argument" dfn-for="accounts_endpoint_response_theme">
+Note: The branding preferences are deliberaly designed to be high level / abstract (rather than opinionated about a specific UI structure), to allow different [=User Agents=] to offer different UI experiences and for them to evolve independently over time.
+
+It is expected to have the following properties:
+
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_branding">
     :   <dfn>background-color</dfn> (optional)
     ::  The [=IDP=]'s preferred background [=color=].
     :   <dfn>color</dfn> (optional)
-    ::  The [=IDP=]'s foreground [=color=] of text.
+    ::  The [=IDP=]'s preferred foreground [=color=] of text.
     :   <dfn>icons</dfn> (optional)
     ::  A list of [=Icon JSON=] objects.
 </dl>
 
 The <dfn>Icon JSON</dfn> is expected to have the following properties:
 
-<dl dfn-type="argument" dfn-for="accounts_endpoint_response_theme_icons">
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_branding_icons">
     :   <dfn>url</dfn> (required)
     ::  The url pointing to the icon image.
     :   <dfn>size</dfn> (optional)
     ::  The size of the icon (assumed to be squared).
+    :   <dfn>background-brightness</dfn> (optional)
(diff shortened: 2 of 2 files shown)


explainer/HOWTO.md
```diff
@@ -118,7 +118,15 @@ containing four fields:
   "accounts_endpoint": "https://idp.example/fedcm/accounts_endpoint",
   "client_id_metadata_endpoint": "https://idp.example/fedcm/client_id_metadata_endpoint",
   "id_token_endpoint": "https://idp.example/fedcm/token_endpoint",
-  "revocation_endpoint": "https://idp.example/fedcm/revocation_endpoint"
+  "revocation_endpoint": "https://idp.example/fedcm/revocation_endpoint",
+  "theme": {
+   "background-color": "green",
```
samuelgoto, line 123: Something that I thought was a little awkward when I was writing this down is that this feels like we are setting a "global background-color" but when in effect we are only allowing the customization of a very small part.

For example, we would want to allow setting the "background-color" of the button but NOT of the title/header part.

The other alternative that occurred to me was to start with the most constrained specification/scope rather than the most general one, e.g.

```json
"theme": {
  "button": {
     "background-color": "..."
  }
}
```

rather than

```json
"theme": {
   "background-color": "..."
}
```

samuelgoto: WDYT?

dj2: I think the difference, at least to me, is that in specifing it as a "theme" or "branding" we aren't specifying that you can style the entire interface. You can only style the bits the "theme" refers too. These are the default values we'll use when we're theming, not the default values we'll use for the entire UI.

So, I'm not opposed to being more specific, but I worry that that will mean duplication in the future. Having the top level keys lets us have defaults for any other specific key we want to add later without duplication.

spec/index.bs
```diff
@@ -507,8 +507,38 @@ The file is parsed expecting the following properties:
...
+
+The <dfn>Icon JSON</dfn> is expected to have the following properties:
+
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_branding_icons">
+    :   <dfn>url</dfn> (required)
+    ::  The url pointing to the icon image.
+    :   <dfn>size</dfn> (optional)
+    ::  The size of the icon (assumed to be squared).
+    :   <dfn>background-brightness</dfn> (optional)
+    ::  The preferred background brightness for the icon. Either "dark" or "light".
```
pkotwicz, line 537: Maybe "intended_background" is clearer?
An IDP may provide several icons:
- An icon for a dark background
- An icon for a light background

spec/index.bs
```diff
@@ -507,8 +507,38 @@ The file is parsed expecting the following properties:
     ::  A URL that points to an HTTP API that complies with the [[#idp-api-id-token-endpoint]] API.
     :   <dfn>revocation_endpoint</dfn>
     :: A URL that points to an HTTP API that complies with the [[#idp-api-revocation-endpoint]] API.
+    :   <dfn>branding</dfn> (optional)
+    ::  A set of [=Branding JSON=] options.
+</dl>
+
+The <dfn>Branding JSON</dfn> allows an [=IDP=] to express their branding preferences, which can be used by User Agents in customization entry points.
```
dj2, line 514: s/can/MAY/ ? do we use the specific MAY, SHOULD, MUST set of words?

dj2: s/in customization entry points/to style UI/ ? entry points seems like a strange wording here (but I always think of entry point as your shader entry point now, heh).

spec/index.bs
```diff
@@ -507,8 +507,38 @@ The file is parsed expecting the following properties:
...
+
+Note: The branding preferences are deliberaly designed to be high level / abstract (rather than opinionated about a specific UI structure), to allow different [=User Agents=] to offer different UI experiences and for them to evolve independently over time.
```
dj2, line 516: s/deliberaly/deliberately

spec/index.bs
```diff
@@ -507,8 +507,38 @@ The file is parsed expecting the following properties:
...
+
+It is expected to have the following properties:
```
dj2, line 518: Are all properties required or should this be `MAY have the following`

spec/index.bs
```diff
@@ -507,8 +507,38 @@ The file is parsed expecting the following properties:
...
+
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_branding">
+    :   <dfn>background-color</dfn> (optional)
+    ::  The [=IDP=]'s preferred background [=color=].
```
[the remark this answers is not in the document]
dj2: I'm not sure if the extra `brand` is useful in this context as it's explicitly in a `branding` json block. You'd end up with:

```
"branding" {
  "brand-background-colour": "blue";
}
```

spec/index.bs
```diff
@@ -507,8 +507,32 @@ The file is parsed expecting the following properties:
...
     :   <dfn>revocation_endpoint</dfn>
     :: A URL that points to an HTTP API that complies with the [[#idp-api-revocation-endpoint]] API.
+    :   <dfn>theme</dfn> (optional)
+    ::  A set of [=Theme JSON=] options.
 </dl>
 
+The <dfn>Theme JSON</dfn> is expected to have the following properties:
+
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_theme">
+    :   <dfn>background-color</dfn> (optional)
```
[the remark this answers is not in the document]
dj2: Elsewhere being things like the manifest spec? Or elsewhere being in this spec?

spec/index.bs
dj2: So the "dark" specifies this is to go on top of a "dark" background? Hm, `background-brightness` isn't a great key fro that. `intended_background` is clearer.

spec/index.bs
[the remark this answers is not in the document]
samuelgoto: I'm inclined to try to pick terms that are compatible with CSS ... so that we don't re-invent CSS ...

spec/index.bs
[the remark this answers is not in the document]
samuelgoto: Right. 

The "branding" scope to me seems like it would give the right context.

I think this works better:

```json
"branding" {
  "background-color": "blue"
}
```

spec/index.bs
```diff
@@ -507,8 +507,32 @@ The file is parsed expecting the following properties:
...
+    ::  The [=IDP=]'s foreground [=color=] of text.
+    :   <dfn>icons</dfn> (optional)
+    ::  A list of [=Icon JSON=] objects.
+</dl>
+
+The <dfn>Icon JSON</dfn> is expected to have the following properties:
+
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_theme_icons">
+    :   <dfn>url</dfn> (required)
+    ::  The url pointing to the icon image.
```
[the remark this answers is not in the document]
samuelgoto: ah, good catch, done.

spec/index.bs
```diff
@@ -507,8 +507,38 @@ The file is parsed expecting the following properties:
...
+
+Note: The branding preferences are deliberaly designed to be high level / abstract (rather than opinionated about a specific UI structure), to allow different [=User Agents=] to offer different UI experiences and for them to evolve independently over time.
+
+It is expected to have the following properties:
+
+<dl dfn-type="argument" dfn-for="accounts_endpoint_response_branding">
+    :   <dfn>background-color</dfn> (optional)
+    ::  The [=IDP=]'s preferred background [=color=].
+    :   <dfn>color</dfn> (optional)
+    ::  The [=IDP=]'s preferred foreground [=color=] of text.
```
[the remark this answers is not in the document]
samuelgoto: Same as above. I think this being inside a "branding" JSON gives the right context.

spec/index.bs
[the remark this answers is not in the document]
samuelgoto: Ah, that's totally a made up word, which was mean to say "places in which the browser could choose to use these branding options".

dj2:
Can this be closed in favour of #202?

samuelgoto:
> Can this be closed in favour of https://github.com/fedidcg/FedCM/pull/202?

Yes, reverting this PR. Thanks @pkotwicz !!

PR #183, opened 2022-02-02
Specify that the URLs passed as "providers" are to be used as a direc…

samuelgoto:
…tory prefix to the fedcm.json configuration file

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1036,7 +1036,7 @@ To <dfn>explicit sign-in</dfn> the user run this algorithm:
 
 To <dfn>fetch the manifest</dfn> configuration file, run this algorithm:
 
-1. Let the |manifest url| be the relative url "fedcm.json" of |provider|["{{FederatedIdentityProvider/url}}"]
+1. Let the |manifest url| be the file "fedcm.json" relative to the directory prefix passed as |provider|["{{FederatedIdentityProvider/url}}"]
 
 1. Return the result of fetching the |manifest url| with the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]] header but without the [=Identity Provider=]'s cookies.
 


dj2 (approved):

spec/index.bs
```diff
@@ -1036,7 +1036,7 @@ To <dfn>explicit sign-in</dfn> the user run this algorithm:
 
 To <dfn>fetch the manifest</dfn> configuration file, run this algorithm:
 
-1. Let the |manifest url| be the relative url "fedcm.json" of |provider|["{{FederatedIdentityProvider/url}}"]
+1. Let the |manifest url| be the file "fedcm.json" relative to the directory prefix passed as |provider|["{{FederatedIdentityProvider/url}}"]
```
yi-gu, line 1039: On the one hand we say "To fetch the manifest configuration file" which implies that the audience being the RP.
On the other hand we explain where to put "fedcm.json" which implies that the audience being the IDP.
Did I get it right?

Another problem is that here we say `manifest url` but such "url" does not explicitly "exist" in the flow. e.g. the IDP only needs to put it under the provider url without referring it anywhere. It's different from the endpoints in which case the IDP does need to have a url "/accounts.php" somewhere.

dj2: The audience is the Browser, not the RP in the "fetch the manifest configuration file". And this is specifying where the browser would find the file.

I'm not sure I understand the second point, we say `Let the |manifest_url|` so, we're defining that it is equal to `prefix/fedcm.json`. So, the `url` exists because we just defined it.

yi-gu: Forgot that the spec is for UA. LGTM.

yi-gu (approved):

samuelgoto:
Thanks all, merging!

PR #184, opened 2022-02-03
Make the client_id_metadata_endpoint optional as well as the terms of…

samuelgoto:
… service and privacy policies

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -499,13 +499,13 @@ Sec-FedCM-CSRF: random_value
 The file is parsed expecting the following properties:
 
 <dl dfn-type="argument" dfn-for="Manifest">
-    :   <dfn>accounts_endpoint</dfn>
+    :   <dfn>accounts_endpoint</dfn> (required)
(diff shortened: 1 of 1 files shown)


dj2 (approved):

spec/index.bs
```diff
@@ -499,13 +499,13 @@ Sec-FedCM-CSRF: random_value
 The file is parsed expecting the following properties:
 
 <dl dfn-type="argument" dfn-for="Manifest">
-    :   <dfn>accounts_endpoint</dfn>
+    :   <dfn>accounts_endpoint</dfn> (required)
     ::  A URL that points to an HTTP API that complies with the [[#idp-api-accounts-endpoint]] API.
-    :   <dfn>client_id_metadata_endpoint</dfn>
+    :   <dfn>client_id_metadata_endpoint</dfn> (optional)
```
cbiesinger, line 504: hmm we need to update our implementation... we currently expect to have a URL in here. we don't care if it's a 404 but we fail if there's no URL. I'll file a bug.

PR #194, opened 2022-02-14
Updating JS API

dj2:
This PR does a first pass at updating the JS API to have split
get/login calls. The `authorize` method is also added for IDP usage.

Issues: #145

***
<a href="https://pr-preview.s3.amazonaws.com/dj2/FedCM/pull/194.html" title="Last updated on Feb 22, 2022, 8:48 PM UTC (dd1b436)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/194/c2680b5...dj2:dd1b436.html" title="Last updated on Feb 22, 2022, 8:48 PM UTC (dd1b436)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -430,6 +430,10 @@ provides the main security boundary on the web.
 :: A general purpose API, as opposed to a [=high-level API=]. See also
     [high level vs low level](https://w3ctag.github.io/design-principles/#high-level-low-level)
 
+: <dfn>minting</dfn>
+: <dfn>minted</dfn>
+:: The act of a new token being creating
+
 : <dfn>out-of-band</dfn>
 :: Outside of the user agent's context.
 
@@ -745,11 +749,13 @@ It will also contain the following parameters in the request body `application/x
 
 <dl dfn-type="argument" dfn-for="id_token_endpoint_request">
     :   <dfn>client_id</dfn>
-    :: The [=RP=]'s client it
+    ::  The [=RP=]'s client it
+    :   <dfn>nonce</dfn>
+    ::  The request nonce
     :   <dfn>account_id</dfn>
-    :: The account identifier that was selected.
+    ::  The account identifier that was selected.
     :   <dfn>consent_acquired</dfn>
-    :: Whether the privacy policy and the terms of service were displayed to the user.
+    ::  Whether the privacy policy and the terms of service were displayed to the user.
 </dl>
 
 For example:
@@ -1108,10 +1114,10 @@ To <dfn noexport>fetch the client id metadata</dfn> run this algorithm:
     1. Return |policies|
 
 The <dfn>Privacy Policy</dfn> are the policies described at
-{{client_id_metadata_endpont_response/privacy_policy_url}}.
+{{client_id_metadata_endpoint_response/privacy_policy_url}}.
 
 The <dfn>Terms of Service</dfn> are the policies described at
-{{client_id_metadata_endpont_response/terms_of_service_url}}.
+{{client_id_metadata_endpoint_response/terms_of_service_url}}.
 
 <!-- ============================================================ -->
 ## The Login API ## {#browser-api-login}
@@ -1131,6 +1137,8 @@ rather calls the same API indistinguishably.
 If all goes well, the [=Relying Party=] receives back a |tokens| object which
 contains an [=id token=] in the form of a signed [[JWT]] which it can use to
 authenticate the user.
+
+The Login API will request the [=id token=] to be [=minted=] by the IDP.
 
 <div class=example>
 ```js
@@ -1177,15 +1185,15 @@ When this method is invoked, the user agent MUST execute the following algorithm
     |manifest|["{{Manifest/id_token_endpoint}}"] with:
     * Let the {{id_token_endpoint_request/account_id}} be
         |account|["{{accounts_endpoint_response_accounts/account_id}}"].
-    * Let the {{id_token_endpoint_request/request}} be a new object with
-       * Let {{id_token_endpoint_request_request/client_id}} be
+    * Let the |id_token_endpoint_request| be a new object with
+       * Let {{id_token_endpoint_request/client_id}} be
           |provider|["{{FederatedIdentityProvider/clientId}}"].
-       * Let the {{id_token_endpoint_request_request/nonce}} be
+       * Let the {{id_token_endpoint_request/nonce}} be
           |provider|["{{FederatedAccountLoginRequest/nonce}}"].
     * This request MUST NOT follow [[RFC7231#header.location|HTTP redirects]]
         and instead abort with an error if there are any.
 1. Return a new {{FederatedTokens}} as:
-    * {{FederatedCredential/idToken}} with |token|
+    * {{FederatedTokens/idToken}} as |token|
 
 
 <!-- ============================================================ -->
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -977,7 +977,7 @@ FedCM API.
 const credential = await navigator.credentials.get({
   federated: {
     providers: [{
-      url: "https://idp.example.com",
+      url: "https://idp.example",
       clientId: "123",
       hint: "dan@example.com"
     }]
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -268,7 +268,7 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
   <script>
   let account = undefined;
 
-  const conn = await navigator.credentials.get({
+  const federation = await navigator.credentials.get({
     federated: {
       providers: [{
         url: "https://idp.example",
@@ -288,6 +288,10 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
     /* Site specific logic to store the accountId */
   }
 
+  function getIdFromIdToken(token) {
+    /* Site logic to decode and retrieve `sub` field from id token */
+  }
+
   function logout() {
     if (account === undefined) {
       return;
@@ -303,23 +307,23 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
   }
 
   function login() {
-    do_login(getLoggedInAccountId());
+    doLogin(getLoggedInAccountId());
   }
 
   function switch() {
-    do_login(undefined);
+    doLogin(undefined);
   }
 
   function try_login() {
-    do_login(getLoggedInAccountId());
+    doLogin(getLoggedInAccountId());
   }
 
-  function do_login(hint) {
-    account = await conn.login({
+  function doLogin(hint) {
+    account = await federation.login({
       hint: hint,
       nonce: "456"
     });
-    setLoggedInAccountId(account.idToken['id']);
+    setLoggedInAccountId(getIdFromIdToken(account.idToken));
   }
   </script>
 </body>
@@ -896,7 +900,7 @@ retrieve {{FederatedAccounts}}.
 
 <div class=example>
 ```js
-const conn = await navigator.credentials.get({
+const federation = await navigator.credentials.get({
   federated: {
     providers: [{
       url: "https://idp.example",
@@ -923,18 +927,13 @@ dictionary FederatedIdentityProvider {
   required USVString url;
   required USVString clientId;
 };
-
-dictionary FederatedCredential {
-  required USVString relyingPartyUrl;
-  required USVString url;
-};
 </xmp>
 
-<dl dfn-type="argument" dfn-for="FederatedCredential">
-    :   <dfn>relyingPartyOrigin</dfn>
-    ::  The origin of the relying party.
+<dl dfn-type="argument" dfn-for="FederatedIdentityProvider">
     :   <dfn>url</dfn>
     ::  The URL of the identity provider origin.
+    :   <dfn>clientId</dfn>
+    ::  The client ID provided to the [=RP=] out of band by the [=IDP=]
 </dl>
 
 This specification overrides the {{FederatedCredential}}'s
@@ -1008,7 +1007,7 @@ authenticate the user.
 
 <div class=example>
 ```js
-const account = conn.login({
+const account = federation.login({
   hint: "dan@example.com",
   signal: signal,
   nonce: "my_nonce"
@@ -1019,7 +1018,7 @@ const account = conn.login({
 
 <xmp class=idl>
 [Exposed=Window, SecureContext]
-dictionary FederatedAccountRequest {
+dictionary FederatedAccountLoginRequest {
   USVString hint;
   USVString nonce;
   bool silent;
@@ -1028,20 +1027,20 @@ dictionary FederatedAccountRequest {
 
 [Exposed=Window, SecureContext]
 partial interface FederatedCredential {
-  Promise<FederatedAccount> login(optional FederatedAccountRequest request = {});
+  Promise<FederatedAccount> login(optional FederatedAccountLoginRequest request = {});
 };
 
 [Exposed=Window, SecureContext]
 dictionary FederatedAccount {
    required USVString id;
-   required USVString idToken;
+   USVString idToken;
 };
 </xmp>
 
-<dl dfn-type="argument" dfn-for="FederatedAccountRequest">
+<dl dfn-type="argument" dfn-for="FederatedAccountLoginRequest">
     :   <dfn>hint</dfn>
     ::  The accountId to auto-select, or |undefined|. If |undefined| is provided
-          then the account chooser will be shown.
+          then the account chooser MAY be shown.
 
           Default: undefined
     :   <dfn>nonce</dfn>
@@ -1162,7 +1161,7 @@ To <dfn>mint</dfn> a new id token, run the following algorithm:
            * Let {{id_token_endpoint_request_request/client_id}} be
               |provider|["{{FederatedIdentityProvider/clientId}}"].
            * Let the {{id_token_endpoint_request_request/nonce}} be
-              |provider|["{{FederatedAccountRequest/nonce}}"].
+              |provider|["{{FederatedAccountLoginRequest/nonce}}"].
         * This request MUST NOT follow [[RFC7231#header.location|HTTP redirects]]
             and instead abort with an error if there are any.
     1. Return a new {{FederatedCredential}} as:
@@ -1185,13 +1184,13 @@ than at the [=RP=].
 
 <div class=example>
 ```js
-const conn = await navigator.credentials.get({
+const federation = await navigator.credentials.get({
   federated: {
     relyingPartyOrigin: “https://rp.example”
   }
 });
 
-conn.authorize({
+federation.authorize({
   name: “dan”
   email “dan@example”
   accountId: 123
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -251,102 +251,149 @@ algorithms to allow the exchange of identity between [=IDP=]s and [=RP=]s.
 When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
 [=RP=] can use to authenticate the user.
 
-<div class=example>
+<!-- ============================================================ -->
+# Single Account Example # {#example-single-user}
+<!-- ============================================================ -->
+<div clas=example>
+Example showing how a website allowing for a single logged in account
+could be implemented.
+
 ```html
 <html>
 <head>
   <title>Welcome to my Website</title>
 </head>
 <body>
-  <button onclick="login()">Login with Foo</button>
-  <button onclick="revoke()">revoke tokens</button>
-  <button onclick="switchAccount()">switch account</button>
-  <button onclick="logout()">logout</button>
+  <button onclick="login()">Login with idp.example</button>
+  <button onclick="revoke()">Revoke tokens</button>
+  <button onclick="logout()">Logout</button>
 
   <script>
-  let tokens = undefined;
-
-  tryLogin();
-
-  function getLoggedInAccountId() {
-    /* Site specifc logic to retrieve the accountId or undefined if none set */
-  }
-  function setLoggedInAccountId(id) {
-    /* Site specific logic to store the accountId */
-  }
-  function idFromIdToken(idToken) {
-    /* Site specific logic to retrieve `sub` field from id token */
-  }
-
-  async function getCredential(opts) {
-    return navigator.credentials.get({
-      mediation: opts.mediation
+  async function login() {
+    const credential = await navigator.credentials.get({
       federated: {
         providers: [{
           url: "https://idp.example",
-          clientId: "123",
-          hint: opts.hint
+          clientId: "123"
+        }]
+      }
+    });
+    return await credential.login({ nonce: "456" });
+  }
+
+  async function logout() {
+    const credential = getLoggedInCredential();
+    if (credential) {
+      return await credential.logout();
+    }
+    return undefined;
+  }
+
+  async function revoke() {
+    const credential = getLoggedInCredential();
+    if (credential) {
+      return await credential.revoke();
+    }
+    return undefined;
+  }
+
+  async function getLoggedInCredential() {
+    return await navigator.credentials.get({
+      mediation: "silent",
+      federated: {
+        providers: [{
+          url: "https://idp.example",
+          clientId: "123"
         }]
       }
     });
   }
+  </script>
+</body>
+</html>
+```
+</div>
 
-  function tryGetCredential() {
-    const id = getLoggedInAccountId();
-    if (id === undefined) {
-      return undefined;
-    }
-    return await getCredential({
-      hint: id,
-      mediation: "silent"
+<!-- ============================================================ -->
+# Multiple Account Example # {#example-multiple-account}
+<!-- ============================================================ -->
+<div class=example>
+Example showing how a website allowing for multiple logged in accounts
+could be implemented.
+
+```html
+<html>
+<head>
+  <title>Welcome to my Website</title>
+</head>
+<body>
+  <button onclick="login()">Login with idp.example</button>
+  <button onclick="revoke()">Revoke tokens</button>
+  <button onclick="multiLogin()">Switch account</button>
+  <button onclick="logout()">Logout</button>
+
+  <script>
+  async function login() {
+    const id = ...; /* site gets stored id if available */
+    const credential = await navigator.credentials.get({
+      mediation: "optional",
+      federated: {
+        providers: [{
+          url: "https://idp.example",
+          clientId: "123",
+          hint: id
+        }]
+      }
     });
+    const tokens = await credential.login({ nonce: "456" });
+    /* site stores id somewhere */
+    return tokens;
   }
 
-  function tryLogin() {
-    const credential = tryGetCredential();
-    if (credential !== undefined) {
-      tokens = doLogin(credential);
-    }
-  }
-
-  function login() {
-    const id = getLoggedInAccountId();
-    const credential = await getCredential({
-      hint: id,
-      mediation: "optional"
+  async function multiLogin() {
+    const credential = await navigator.credentials.get({
+      mediation: "required",
+      federated: {
+        providers: [{
+          url: "https://idp.example",
+          clientId: "123",
+        }]
+      }
     });
-    if (credential !== undefined) {
-      tokens = doLogin(credential);
-    }
-  }
-
-  function switchAccount() {
-    const credential = await getCredential({mode: "required"});
-    if (credential !== undefined) {
-      tokens = doLogin(credential);
-    }
-  }
-
-  function doLogin(credential) {
-    const tokens = await credential.login({
-      nonce: "456"
-    });
-    setLoggedInAccountId(idFromIdToken(tokens.idToken));
+    const tokens = await credential.login({ nonce: "456" });
+    /* site stores id somewhere */
     return tokens;
   }
 
   async function logout() {
-    const credential = tryGetCredential();
-    if (credential !== undefined) {
+    const credential = await getLoggedInCredential();
+    if (credential) {
       credential.logout();
     }
   }
 
   async function revoke() {
-    const credential = tryGetCredential();
-    if (credential !== undefined) {
+    const credential = await getLoggedInCredential();
+    if (credential) {
       credential.revoke();
     }
+  }
+
+  async function getLoggedInCredential() {
+    const id = ...; /* site gets stored id if available */
+    if (id === undefined) {
+      return undefined;
+    }
+    return navigator.credentials.get({
+      mediation: "silent",
+      federated: {
+        providers: [{
+          url: "https://idp.example",
+          clientId: "123",
+          hint: id,
+        }]
+      }
+    });
   }
   </script>
 </body>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -260,7 +260,7 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
 <body>
   <button onclick="login()">Login with Foo</button>
   <button onclick="revoke()">revoke tokens</button>
-  <button onclick="switch()">switch</button>
+  <button onclick="switchAccount()">switch account</button>
   <button onclick="logout()">logout</button>
 
   Welcome <span id="name"></span>.
@@ -268,26 +268,16 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
   <script>
   let tokens = undefined;
 
-  try_login();
+  tryLogin();
 
   function getLoggedInAccountId() {
     /* Site specifc logic to retrieve the accountId or undefined if none set */
-    const parts = document.cookie
-                    .split(';')
-                    .find(row => row.trim().startsWith('id='));
-    if (parts !== undefined) {
-      return parts.split('=')[1];
-    }
-    return undefined;
   }
-
   function setLoggedInAccountId(id) {
     /* Site specific logic to store the accountId */
-    document.cookie = `id=#{id}`
   }
-
   function idFromIdToken(idToken) {
-    return JSON.parse(Base64.decode(idToken.split(“.”)[1])).sub;
+    /* Site specific logic to retrieve `sub` field from id token */
   }
 
   async function getCredential(opts) {
@@ -314,7 +304,7 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
     });
   }
 
-  function try_login() {
+  function tryLogin() {
     const credential = tryGetCredential();
     if (credential !== undefined) {
       tokens = doLogin(credential);
@@ -332,7 +322,7 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
     }
   }
 
-  function switch() {
+  function switchAccount() {
     const credential = await getCredential({mode: "required"});
     if (credential !== undefined) {
       tokens = doLogin(credential);
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -254,7 +254,7 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
 <!-- ============================================================ -->
 # Single Account Example # {#example-single-user}
 <!-- ============================================================ -->
-<div clas=example>
+<div class=example>
 Example showing how a website allowing for a single logged in account
 could be implemented.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -258,7 +258,7 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
   <title>Welcome to my WebSite</title>
 </head>
 <body>
-  <button onclick="login('https://foo.example', '123')">Sign-in with Foo</button>
+  <button onclick="login()">Sign-in with Foo</button>
   <button onclick="revoke()">revoke tokens</button>
   <button onclick="switch()">switch</button>
   <button onclick="logout()">logout</button>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -255,7 +255,7 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
 ```html
 <html>
 <head>
-  <title>Welcome to my WebSite</title>
+  <title>Welcome to my Website</title>
 </head>
 <body>
   <button onclick="login()">Login with Foo</button>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -931,7 +931,7 @@ dictionary FederatedIdentityProvider {
 
 <dl dfn-type="argument" dfn-for="FederatedIdentityProvider">
     :   <dfn>url</dfn>
-    ::  The URL of the identity provider origin.
+    ::  The URL of the identity provider.
     :   <dfn>clientId</dfn>
     ::  The client ID provided to the [=RP=] out of band by the [=IDP=]
 </dl>
@@ -1001,7 +1001,7 @@ rather calls the same API indistinguishably.
 The [=Relying Party=] can control a variety of UX knobs. For
 example an `AbortSignal` in case it needs to abort the flow.
 
-If all goes well, the [=Relying Party=] receives back a |user| object which
+If all goes well, the [=Relying Party=] receives back a |account| object which
 contains an [=id token=] in the form of a signed [[JWT]] which it can use to
 authenticate the user.
 
@@ -1011,7 +1011,6 @@ const account = federation.login({
   hint: "dan@example.com",
   signal: signal,
   nonce: "my_nonce"
-
 });
 ```
 </div>
@@ -1172,86 +1171,6 @@ To <dfn>mint</dfn> a new id token, run the following algorithm:
             |provider|["{{FederatedIdentityProvider/url}}"]
 
 <!-- ============================================================ -->
-### IDP Sign-in API ### {#browser-api-idp-sign-in}
-<!-- ============================================================ -->
-
-The IDP Sign-in API allows an [=IDP=] to create a credential for the user in
-a given [=RP=] that can later be retrieved by the [=RP=].
-
-Its equivalent to the [[#browser-api-rp-sign-in]] in its effect,
-except that it is designed to be called while the user is at the [=IDP=] rather
-than at the [=RP=].
-
-<div class=example>
-```js
-const federation = await navigator.credentials.get({
-  federated: {
-    relyingPartyOrigin: “https://rp.example”
-  }
-});
-
-federation.authorize({
-  name: “dan”
-  email “dan@example”
-  accountId: 123
-});
-```
-</div>
-
-<xmp class=idl>
-[exposed=window, SecureContext]
-partial dictionary FederatedIdentityProvider {
-  USVString relyingPartyOrigin;
-};
-
-[Exposed=Window, SecureContext]
-dictionary FederatedAuthorizationRequest {
-  required USVString name;
-  required USVString email;
-  required USVString picture;
-  required USVString id;
-
-  USVString privacyPolicyUrl;
-  USVString termsOfServiceUrl;
-};
-
-[Exposed=Window, SecureContext]
-partial interface FederatedCredential {
-  Promise<undefined> authorize(FederatedAuthorizationRequest request);
-};
-</xmp>
-
-<dl dfn-type="argument" dfn-for="FederatedIdentityProvider">
-    :   <dfn>relyingPartyOrigin</dfn>
-    ::  The relying party origin for which this credential is being created.
-</dl>
-
-<dl dfn-type="argument" dfn-for="FederatedAuthorizationRequest">
-    :   <dfn>name</dfn>
-    ::  The name to display in the dialog
-    :   <dfn>email</dfn>
-    ::  The email to display in the dialog
-    :   <dfn>picture</dfn>
-        The URL of the users profile photo to display in the dialog
-    :   <dfn>id</dfn>
-    ::  The users account id
-    :   <dfn>privacyPolicyUrl</dfn>
-    ::  The privacy policy url
-    :   <dfn>termsOfServiceUrl</dfn>
-    ::  The terms of service url
-</dl>
-
-When this method is invoked, the user agent MUST execute the following algorithm:
-
-    1. Gather explicit intent to create an account
-    1. Gather explicit agreement with the [=Privacy Policy=]
-    1. Gather explicit agreement with the [=Terms of Service=]
-    1. Set the account's {{StateMachine/Account State}} from [=unregistered=]
-        to [=registered=].
-    1. Set the account's {{StateMachine/Session State}} from [=logged out=]
-        to [=logged in=].
-
-<!-- ============================================================ -->
 ## The Revocation API ## {#browser-api-revocation}
 <!-- ============================================================ -->
 
@@ -1336,12 +1255,13 @@ path: img/mock31.svg
 
 <div class=example>
 ```js
-await FederatedCredential.logoutRPs([
-  { url: "https://rp1.example",
-    accountId: "123" },
-  { url: "https://rpN.example",
-    accountId: "456" }
-]);
+await FederatedCredential.logoutRPs([{
+    url: "https://rp1.example",
+    accountId: "123"
+  }, {
+    url: "https://rpN.example",
+    accountId: "456"
+  }]);
 ```
 </div>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -268,7 +268,7 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
   <script>
   let account = undefined;
 
-  const federation = await navigator.credentials.get({
+  const federated_credential = await navigator.credentials.get({
     federated: {
       providers: [{
         url: "https://idp.example",
@@ -319,7 +319,7 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
   }
 
   function doLogin(hint) {
-    account = await federation.login({
+    account = await federated_credential.login({
       hint: hint,
       nonce: "456"
     });
@@ -900,7 +900,7 @@ retrieve {{FederatedAccounts}}.
 
 <div class=example>
 ```js
-const federation = await navigator.credentials.get({
+const federated_credential = await navigator.credentials.get({
   federated: {
     providers: [{
       url: "https://idp.example",
@@ -1007,7 +1007,7 @@ authenticate the user.
 
 <div class=example>
 ```js
-const account = federation.login({
+const account = federated_credential.login({
   hint: "dan@example.com",
   signal: signal,
   nonce: "my_nonce"
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -90,7 +90,7 @@ used by certain designs of federated identity.
 The Federated Credential Management API aims to bridge the gap for the
 federated identity designs which relied on third-party cookies.
 The API provides the primitives needed to support federated identity when/where
-it depends on third-party cookies, from sign-in to sign-out and revocation.
+it depends on third-party cookies, from login to logout and revocation.
 
 In order to provide the federated identity primitives without the use of
 third-party cookies the API places the User Agent as a mediator between [=RPs=]
@@ -125,7 +125,7 @@ Upon user agreement an [=account=] is created with the [=Relying Party=] and
 the user has a [=session=] initialized.
 
 <!-- ============================================================ -->
-### Sign-in ### {#use-cases-sign-in}
+### Login ### {#use-cases-login}
 <!-- ============================================================ -->
 
 After a user navigates to a [=Relying Party=] in a browser and decides to
@@ -133,42 +133,42 @@ create an account by going through their [[#use-cases-sign-up]] flow,
 there are two ways a user logs into their [=account=] once their [=session=] expires:
 
 <!-- ============================================================ -->
-#### Auto Sign-in #### {#use-cases-auto-sign-in}
+#### Auto login #### {#use-cases-auto-login}
 <!-- ============================================================ -->
 
-Auto Sign-in occurs when the [=Identity Provider=] has already gathered
+Auto login occurs when the [=Identity Provider=] has already gathered
 permissions from the user to share their identity with the [=Relying Party=]
 and automatically signs the user in.
 
 For example, the user has previously executed the [[#use-cases-sign-up]] flow
 and then changes from their phone to their laptop. On the new device the
-user goes to the [=Relying Party=] and selects to sign-in using their
+user goes to the [=Relying Party=] and selects to login using their
 [=Identity Provider=]. The [=Identity Provider=] knows, and proves, the user
 has signed up to the [=Relying Party=] and the [=Relying Party=] creates a
 new [=session=] for the users [=account=].
 
 <!-- ============================================================ -->
-#### Explicit Sign-in #### {#use-cases-explicit-sign-in}
+#### Explicit Login #### {#use-cases-explicit-login}
 <!-- ============================================================ -->
 
-An explicit sign-in occurs when the [=Identity Provider=] believes it is
+An explicit login occurs when the [=Identity Provider=] believes it is
 necessary to gather an explicit permission from the user to sign into a
 [=Relying Party=], typically after the user goes through a
-[[#use-cases-sign-out]] flow.
+[[#use-cases-logout]] flow.
 
-For example, after the user has done the [[#use-cases-sign-out]] flow of the
+For example, after the user has done the [[#use-cases-logout]] flow of the
 [=Relying Party=] they decide to log in again. The user visits the
-[=Relying Party=] and selects their [=Identity Provider=] to sign-in. The
+[=Relying Party=] and selects their [=Identity Provider=] to login. The
 [=Identity Provider=] knows:
     * the user already has an account with the [=Relying Party=].
     * the user has logged out of the [=Relying Party=].
 
-The user is then prompted, "Do you want to sign-in with the [=Relying Party=]?"
+The user is then prompted, "Do you want to login with the [=Relying Party=]?"
 and upon user agreement the [=Relying Party=] creates a new [=session=] with
 the users existing [=account=].
 
 <!-- ============================================================ -->
-### Sign-out ### {#use-cases-sign-out}
+### Logout ### {#use-cases-logout}
 <!-- ============================================================ -->
 
 After a user navigates to a [=Relying Party=] in a browser and decides to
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -254,61 +254,79 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
-  </head>
-  <body>
-    <button onclick="login('https://foo.example', '123')">Sign-in with Foo</button>
-    <button onclick="logout()">logout</button>
+<head>
+  <title>Welcome to my WebSite</title>
+</head>
+<body>
+  <button onclick="login('https://foo.example', '123')">Sign-in with Foo</button>
+  <button onclick="revoke()">revoke tokens</button>
```
samuelgoto, line 262: Unregister account

samuelgoto: or "delete account" as below

spec/index.bs
```diff
@@ -254,61 +254,79 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
 
-    Welcome <span id="name"></span>!!
+  Welcome <span id="name"></span>.
 
-    <button onclick="revoke">delete account</button>
+  <script>
+  let user = undefined;
 
-    <script>
+  const auth = await navigator.credentials.get({
```
samuelgoto, line 271: there is something awkward about "auth" ... i can't come up with anything better right now, but in case you can, this feels awkward at the moment.

spec/index.bs
```diff
@@ -254,61 +254,79 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+    federated: {
+      providers: [{
+        url: "https://idp.example",
+        clientId: "123",
+        nonce: "456"
```
samuelgoto, line 276: can we leave "nonce" to the `login` call?

spec/index.bs
```diff
@@ -254,61 +254,79 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
-      // ... update local storage ...
-
-      user = parse(idToken);
-      document.getElementById("name").innerText = user.name;
+  function logout() {
+    if (user === undefined) {
+      return;
     }
+    let id = user.id_token['id'];
+    localStorage["users"].remove(user);
```
samuelgoto, line 288: this isn't a valid usage of localStorage is it?

spec/index.bs
```diff
@@ -254,61 +254,79 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+
+    user.logout();
```
samuelgoto, line 290: user => account

spec/index.bs
```diff
@@ -254,61 +254,79 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
-  </body>
+    do_login(hint);
+  }
 
+  function switch() {
+    do_login(undefined);
+  }
+
+  function try_login() {
+    if (localStorage["users"].empty()) {
```
samuelgoto, line 313: there is a chance that we could remove all of this `localStorage` business here to make the example simpler/smaller. wdyt?

spec/index.bs
```diff
@@ -873,89 +893,32 @@ path: img/mock42.svg
...
-supports and has pre-registered with (i.e. it has a `clientId`).
-
-Additionally, the [=Relying Party=] can control a variety of UX knobs. For
-example, whether to allow an [[#use-cases-auto-sign-in]] flow and an
-`AbortSignal` in case it needs to abort the flow.
-
-If all goes well, the [=Relying Party=] receives back an [=id token=] in the
-form of a signed [[JWT]] which it can use to authenticate the user.
+Interactions with FedCM are split into two primary objects. The |auth| object
+and the |account| object. The |auth| object allows making calls to create and
```
samuelgoto, line 899: like i said earlier, `auth` feels awkward here ...

spec/index.bs
```diff
@@ -873,89 +893,32 @@ path: img/mock42.svg
...
-      }],
-    }
-  });
-
-  return idToken;
-}
+const auth = await navigator.credentials.get({
+  federated: {
+    providers: [{
+      identity_provider_url: "https://idp.example",
```
samuelgoto, line 907: This is already an FederatedIdentityProvider object, inside the `providers` array. Shorten this up and call it a `url` (or `manifest`?) instead?

spec/index.bs
```diff
@@ -873,89 +893,32 @@ path: img/mock42.svg
...
+      clientId: "123",
+      nonce: "456"
```
samuelgoto, line 909: like i said earlier, i think `nonce` should be part of `login()`.

spec/index.bs
```diff
@@ -967,54 +930,147 @@ dictionary FederatedIdentityProvider {
   required USVString clientId;
   USVString nonce;
 };
+
+dictionary FederatedAuthCredential {
```
samuelgoto, line 934: Why not reuse the `FederatedCredential` object? This seems off to me. Check with Mike West and/or Nina or @kenrb ?

spec/index.bs
```diff
@@ -873,89 +893,32 @@ path: img/mock42.svg
...
-        clientId: "123",
-        nonce: "456"
-      }],
-    }
-  });
-
-  return idToken;
-}
+const auth = await navigator.credentials.get({
+  federated: {
```
samuelgoto, line 905: `federated` is an property that is reserved for the `FederatedCredential` object type, I believe.

spec/index.bs
```diff
@@ -967,54 +930,147 @@ dictionary FederatedIdentityProvider {
   required USVString clientId;
   USVString nonce;
 };
+
+dictionary FederatedAuthCredential {
+  required USVString relying_party_url;
+  required USVString identity_provider_url;
```
samuelgoto, line 936: dictionaryKeysAreCamelCasedForJS

spec/index.bs
```diff
@@ -967,54 +930,147 @@ dictionary FederatedIdentityProvider {
...
+  required bool is_consented;
```
samuelgoto, line 937: What is this?

spec/index.bs
```diff
@@ -967,54 +930,147 @@ dictionary FederatedIdentityProvider {
   required USVString clientId;
   USVString nonce;
 };
+
+dictionary FederatedAuthCredential {
+  required USVString relying_party_url;
```
samuelgoto, line 935: relyingPartyUrl and identityProviderUrl? or, for short, rpUrl and idpUrl?

spec/index.bs
```diff
@@ -967,54 +930,147 @@ dictionary FederatedIdentityProvider {
...
+  required bool is_consented;
+};
 </xmp>
 
-And finally, this specification overrides the {{FederatedCredential}}'s
-<code><dfn for="FederatedCredential" method>\[[DiscoverFromExternalSource]](
-origin, options, sameOriginWithAncestors)</dfn></code> Method.
+<dl dfn-type="argument" dfn-for="FederatedAuthCredential">
+    :   <dfn>relying_party_url</dfn>
+    ::  The URL of the relying party origin.
```
samuelgoto, line 943: add an example when/how this is expected to be used?

spec/index.bs
```diff
@@ -967,54 +930,147 @@ dictionary FederatedIdentityProvider {
...
+    :   <dfn>identity_provider_url</dfn>
+    ::  The URL of the identity provider origin.
+    :   <dfn>is_consented</dfn>
```
samuelgoto, line 946: isn't this a parameter that is passed, rather than returned?

spec/index.bs
```diff
@@ -967,54 +930,147 @@ dictionary FederatedIdentityProvider {
...
+    ::  Boolean if the communication between the two origins is consented.
+
+          Default: `false`
+</dl>
+
+This specification overrides the {{FederatedCredential}}'s
```
samuelgoto, line 952: Are you overriding or are you exposing a new type, FederatedAuthCredential?

spec/index.bs
```diff
@@ -967,54 +930,147 @@ dictionary FederatedIdentityProvider {
...
+contains an [=id token=] in the form of a signed [[JWT]] which it can use to
+authenticate the user.
+
+<div class=example>
+```js
+user = await auth.login({
+  hint: "dan@example.com",
+  signal: signal
+});
+```
```
samuelgoto, line 1027: const idToken = user.getIdToken() later on?

spec/index.bs
```diff
@@ -967,54 +930,147 @@ dictionary FederatedIdentityProvider {
...
-        Note: This restriction aims to address the concern raised in [[Security-Origin-Confusion]].
-    1.  Assert: |options|["{{CredentialRequestOptions/federated}}"]["{{FederatedCredentialRequestOptions/providers}}"] [=map/exists=].
-    1.  Assert: |options|["{{CredentialRequestOptions/federated}}"]["{{FederatedCredentialRequestOptions/providers}}"] [=list/size=] is 1.
+If all goes well, the [=Relying Party=] receives back a |user| object which
+contains an [=id token=] in the form of a signed [[JWT]] which it can use to
+authenticate the user.
+
+<div class=example>
+```js
+user = await auth.login({
```
samuelgoto, line 1023: const user

spec/index.bs
```diff
@@ -967,54 +930,147 @@ dictionary FederatedIdentityProvider {
...
+
+[Exposed=Window, SecureContext]
+partial interface FederatedAuthCredential {
+  Promise<FederatedAccount> login(optional FederatedAccountRequest request);
+};
+
+[Exposed=Window, SecureContext]
+dictionary FederatedAccount {
+   required USVString id;
+   required USVString id_token;
```
samuelgoto, line 1046: idToken as propertiesAreCamelCased?

spec/index.bs
```diff
@@ -967,54 +930,147 @@ dictionary FederatedIdentityProvider {
...
+
+          Default: false
+    :   <dfn>signal</dfn>
+    ::  Abort signal
+</dl>
+
+<dl dfn-type="argument" dfn-for="FederatedAccount">
+    :   <dfn>id</dfn>
+    ::  The account id as provided by the [=IDP=]
+    :   <dfn>id_token</dfn>
```
samuelgoto, line 1068: idToken

spec/index.bs
```diff
@@ -1027,7 +1083,6 @@ When this method is invoked, the user agent MUST execute the following algorithm
         1. Return the result of running the [=explicit sign-in=] algorithm.
     1. Return the result of running the [=auto sign-in=] algorithm.
```
samuelgoto, line 1084: these algorithm are a bit messy right now, we should clean them up at some point later on.

spec/index.bs
```diff
@@ -1036,12 +1091,16 @@ To <dfn>sign-up</dfn> the user run this algorithm:
         to [=registered=].
     1. Set the account's {{StateMachine/Session State}} from [=logged out=]
         to [=logged in=].
-    1. Return the result of [=minting=] a new [=id token=].
+    1. Set the {{FederatedAuthCredential}} to |consented|
```
samuelgoto, line 1094: so `consented` is per profile, right? not per `FederatedAccount` right?

samuelgoto, line 1094: there is something awkward about `consented` ... what have we used in the past? any precedent to pull from? would `approved` or `allowed` work better here?

spec/index.bs
```diff
@@ -1133,20 +1193,74 @@ than at the [=RP=].
...
-  const credential = new FederatedCredential({
-    id: id,
-    provider: 'https://idp.example',
-    name: name,
-    iconURL: iconUrl,
-    client_id: "1234",
-  });
+const auth = await navigator.credentials.get({
+  federated: {
+    relying_party_origin: “https://rp.example”
```
samuelgoto, line 1198: `rpUrl` or `relyingPartyUrl`?

spec/index.bs
```diff
@@ -1133,20 +1193,74 @@ than at the [=RP=].
```
samuelgoto, line 1193: The introduction to this section is out of date.

spec/index.bs
```diff
@@ -1133,20 +1193,74 @@ than at the [=RP=].
...
-  });
+const auth = await navigator.credentials.get({
+  federated: {
+    relying_party_origin: “https://rp.example”
+  }
+});
 
-  return await navigator.credentials.store(credential);
-}
+auth.authorize({
```
samuelgoto, line 1202: `authorize` felt awkward at first ... aren't there better options here? a few come to mind, in no particular order:

- `store`?
- `approve`?
- `forward`?
- anything else?

These all seem like bad choices, but nothing else is occurring to me now.

spec/index.bs
```diff
@@ -1133,20 +1193,74 @@ than at the [=RP=].
...
+  email “dan@example”
+  account_id: 123
+});
 ```
 </div>
 
+<xmp class=idl>
+[exposed=window, SecureContext]
+partial dictionary FederatedIdentityProvider {
+  USVString relying_party_origin;
```
samuelgoto, line 1213: relyingParty?

spec/index.bs
```diff
@@ -1133,20 +1193,74 @@ than at the [=RP=].
...
+};
+
+[Exposed=Window, SecureContext]
+dictionary FederatedAuthorizationRequest {
+  required USVString name;
+  required USVString email;
+  required USVString profile_picture;
```
samuelgoto, line 1220: I think I'd make things consistent with this array:

https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims

So, `picture` here instead.

spec/index.bs
```diff
@@ -1133,20 +1193,74 @@ than at the [=RP=].
...
+  required USVString id;
+
+  USVString privacy_policy_url;
```
samuelgoto, line 1223: privacyPolicy

spec/index.bs
```diff
@@ -1133,20 +1193,74 @@ than at the [=RP=].
...
+  USVString terms_of_service_url;
```
samuelgoto, line 1224: termsOfService

spec/index.bs
```diff
@@ -1133,20 +1193,74 @@ than at the [=RP=].
...
+<dl dfn-type="argument" dfn-for="FederatedAuthorizationRequest">
+    :   <dfn>name</dfn>
+    ::  The name to display in the dialog
+    :   <dfn>email</dfn>
+    ::  The email to display in the dialog
+    :   <dfn>profile_picture</dfn>
+        The URL of the users profile photo to display in the dialog
+    :   <dfn>id</dfn>
+    ::  The users account id
+    :   <dfn>privacy_policy_url</dfn>
```
samuelgoto, line 1247: privacyPolicy

spec/index.bs
```diff
@@ -1133,20 +1193,74 @@ than at the [=RP=].
...
+    ::  The privacy policy url
+    :   <dfn>terms_of_service_url</dfn>
```
samuelgoto, line 1249: termsOfService

spec/index.bs
```diff
@@ -1133,20 +1193,74 @@ than at the [=RP=].
...
+  Promise<undefined> authorize(FederatedAuthorizationRequest request);
+};
+</xmp>
+
+<dl dfn-type="argument" dfn-for="FederatedIdentityProvider">
+    :   <dfn>relying_party_origin</dfn>
+    ::  The relying party origin for which this credential is being created.
+</dl>
+
+<dl dfn-type="argument" dfn-for="FederatedAuthorizationRequest">
```
samuelgoto, line 1238: can't we reuse the `FederatedAccount` here somehow? it feels like duplicating things here to represent the same concept.

e.g.:

auth.authorize({
  account: new FederatedAccount({id: 1, email: foo, name: bar}),
  privacyPolicy: ...,
  termsOfService: ...
});

WDYT?

spec/index.bs
```diff
@@ -1133,20 +1193,74 @@ than at the [=RP=].
...
+When this method is invoked, the user agent MUST execute the following algorithm:
+
+    1. Gather explicit intent to create an account
+    1. Gather explicit agreement with the [=Privacy Policy=]
+    1. Gather explicit agreement with the [=Terms of Service=]
+    1. Set the account's {{StateMachine/Account State}} from [=unregistered=]
+        to [=registered=].
+    1. Set the account's {{StateMachine/Session State}} from [=logged out=]
+        to [=logged in=].
+    1. Set the {{FederatedAuthCredential}} to |consented|
```
samuelgoto, line 1262: `approved`

spec/index.bs
```diff
@@ -1190,18 +1302,20 @@ When a user wants to log out of their session in the [=Relying Party=],
...
 flow to be disabled, so that the user doesn't get into an infinite loop.
 
-It does so by calling the `preventSilentAccess()` API, which clear the
-user's current account session and takes them to an [[#use-cases-explicit-sign-in]] flow
-whenever the user wants to log back in.
-
 <div class=example>
 ```js
-function logout() {
-  navigator.credentials.preventSilentAccess();
```
samuelgoto, line 1200: ah, yes, much better now.

spec/index.bs
```diff
@@ -1232,15 +1346,12 @@ path: img/mock31.svg
...
-async function logout() {
-  await FederatedCredential.logoutRPs([{
-      url: "https://rp1.example",
-      accountId: "123",
-    }, {
-      url: "https://rpN.example",
-      accountId: "456",
-    }]);
-}
+FederatedCredential.logoutRPs([
```
samuelgoto, line 1349: this is still an `await`, right?

spec/index.bs

spec/index.bs
dj2: What about |conn| for connection?

spec/index.bs

spec/index.bs
dj2: Done, thoughts?

spec/index.bs
dj2: ack

spec/index.bs
dj2: Removed

spec/index.bs
dj2: We've been overriding what `federated` contains in the spec, I'm just continine that?

spec/index.bs

spec/index.bs

spec/index.bs
dj2: Went with `relyingPartyUrl`. As above `identity_provider_url` -> `url`

spec/index.bs
dj2: that_makes_me_sad. done.

spec/index.bs
dj2: Implementation detail I was thinking about but that doesn't make sense in the spec. Removed.

spec/index.bs
dj2: There is one in the `IDP Sign-in API` section.

spec/index.bs
dj2: Implementation detail, removed.

spec/index.bs
dj2: We override the `[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)` method regardless of what we return.

spec/index.bs

spec/index.bs
dj2: I'm not sure what this means? I was just showing how to get the `account` object, not sure if the id token bit is needed?

spec/index.bs

spec/index.bs

spec/index.bs
dj2: ack. They need to be gone over and make sure they match up to the new structure as well. I went over them, but they probably still aren't quite correct.

spec/index.bs
dj2: Yes, the consented is per profile (at least in Chrome). But I'm removing this bit as it's an implementation detail I think.

dj2: removed.

spec/index.bs
dj2: The `RP Sign-in API` section? How so, it seems correct to me?

spec/index.bs
dj2: `request`?

spec/index.bs
dj2: `relyingPartyOrigin`

spec/index.bs

spec/index.bs
dj2: `privacyPolicyUrl`

spec/index.bs
dj2: `termsOfServiceUrl`

spec/index.bs
dj2: The `FederatedAccount` doesn't have `name`, `email`, etc in it. It has an `idToken`. Having the request object lets us extend to include extra things in the future if we need without shoe-horning them into the `FederatedAccount` object.

spec/index.bs
dj2: ack

spec/index.bs
dj2: ack

spec/index.bs
dj2: removed

spec/index.bs
dj2: I wasn't sure the right framing for things at the time. What do you think about:

* `get` returns a `FederatedCredential` as that's what tells you if you have consent to the communication (that's our credential)
* `login` returns a `FederatedAccount` which has the `accountId` associated with it for future calls. The account also has the `idToken`

spec/index.bs
samuelgoto: For IDPs, `revoke` is equivalent to `delete account` or `disconnect`. e.g. calling this API [1] gets reflected as actually `disconnecting the RP` as far as the IDP is concerned. 

We should also expect RPs to call `revoke` only when the user is about to delete their account on the RP too.

[1] https://developers.google.com/identity/gsi/web/guides/revoke

spec/index.bs
samuelgoto: I was making a comment about the object type that this returned. If we converge on having the `get` return a `FederatedCredential`, I don't feel strongly what we call the JS variable here.

spec/index.bs
samuelgoto: SGTM

spec/index.bs
samuelgoto: LGTM

This is what I was imagining too, upon first glance.

spec/index.bs
samuelgoto: > The FederatedAccount doesn't have name, email, etc in it. It has an idToken. 

Should we create perhaps a FederatedProfile type and reuse it between FederatedAccount and the authorize here?

> Having the request object lets us extend to include extra things in the future if we need without shoe-horning them into the FederatedAccount object.

Can you give me a concrete example of what you'd expect we would need to shoe horn that we wouldn't be able to insert into the top level object?

```js
auth.authorize({
  account: new FederatedProfile({id: 1, email: foo, name: bar}),
  privacyPolicy: ...,
  termsOfService: ...
  otherRandomStuff: ...
});
```

I don't actually know if this is a good idea, I just wanted to bring up and make sure it is deliberate, but I don't feel strongly about this and I can see where you are coming from.

spec/index.bs
```diff
@@ -266,64 +266,60 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
-        clientId: "123",
-        nonce: "456"
+        clientId: "123"
       }]
     }
   });
 
   try_login();
 
+  function get_logged_in_account_id() {
```
samuelgoto, line 282: getLoggedInAccountId

spec/index.bs
```diff
@@ -266,64 +266,60 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+    /* Site specifc logic to retrieve the accountId or undefined if none set */
+    return undefined;
+  }
+
+  function set_logged_in_account_id() {
```
samuelgoto, line 287: setLoggedInAccountId

spec/index.bs
```diff
@@ -893,20 +887,19 @@ path: img/mock42.svg
 </dl>
 
 <!-- ============================================================ -->
-## The Authorization Object
+## The Connection Object
```
samuelgoto, line 890: Connection -> FederatedCredential?

spec/index.bs
```diff
@@ -1193,35 +1184,35 @@ than at the [=RP=].
 
 <div class=example>
 ```js
-const auth = await navigator.credentials.get({
+const conn = await navigator.credentials.get({
   federated: {
-    relying_party_origin: “https://rp.example”
+    relyingPartyOrigin: “https://rp.example”
```
samuelgoto, line 1189: do we really need all of these `Origin` and `Url` suffixes for determining the types? can't we assume that `relyingParty` will be an origin and termsOfService and privacyPolicy are going to be `url`? 

i don't feel strongly, but this seems like more verbose/longer than it needs to be for readers.

e.g. if we need to make them take other things in the future we can overload it with an object? e.g.

```js
{
  relyingParty: {
    somethingElseThatIsntAnOrigin: .....,
    anotherSomethingElse: .....
  }
}
```

samuelgoto:
Looking good! Just a few more clarification questions, and adding @cbiesinger and @yi-gu to review it too.

samuelgoto:
I left some comments but they are showing as `pending` to me here ... I'm trying to figure out how to publish them, in case you didn't get them.

spec/index.bs
dj2: Hah, my editor was un-synced from the line numbers in the diff, so linked to the wrong section.

spec/index.bs
dj2: I tend towards explicit so it's clear exactly what we're expecting in a given location. Curious what others think.

spec/index.bs
dj2: That would require inspecting the `idToken` to pull out the required information which we have been avoiding doing up to this point.

spec/index.bs
dj2: Clarified, better?

spec/index.bs

spec/index.bs

spec/index.bs
samuelgoto: LGTM

spec/index.bs
```diff
@@ -254,61 +254,75 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+    if (account === undefined) {
+      return;
     }
-    </script>
-  </body>
+    account.revoke();
+  }
 
+  function login() {
+    do_login(getLoggedInAccountId());
```
samuelgoto, line 306: doLogin()

spec/index.bs
```diff
@@ -254,61 +254,75 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+  function try_login() {
+    do_login(getLoggedInAccountId());
+  }
+
+  function do_login(hint) {
+    account = await conn.login({
+      hint: hint,
+      nonce: "456"
+    });
+    setLoggedInAccountId(account.idToken['id']);
```
samuelgoto, line 322: IdTokens are encoded as a JWT:

https://jwt.io/

They look like this:

```
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c
```

So, you can't do `idToken["id"]` because they are a string, not an object.

spec/index.bs
```diff
@@ -965,56 +922,148 @@ partial dictionary FederatedCredentialRequestOptions {
 dictionary FederatedIdentityProvider {
   required USVString url;
   required USVString clientId;
-  USVString nonce;
+};
+
+dictionary FederatedCredential {
+  required USVString relyingPartyUrl;
```
samuelgoto, line 928: Aren't these things that should be part of the `FederatedCredentialRequestOptions` instead?

spec/index.bs
```diff
@@ -965,56 +922,148 @@ partial dictionary FederatedCredentialRequestOptions {
...
+  required USVString url;
```
samuelgoto, line 929: What is this for?

spec/index.bs
```diff
@@ -965,56 +922,148 @@ partial dictionary FederatedCredentialRequestOptions {
...
+  signal: signal,
+  nonce: "my_nonce"
+
+});
+```
+</div>
+
+<xmp class=idl>
+[Exposed=Window, SecureContext]
+dictionary FederatedAccountRequest {
```
samuelgoto, line 1022: FederatedAccountLoginRequest? <= I'm guessing there are going to be more methods in FederatedAccount that are going to take objects going forward?

spec/index.bs
```diff
@@ -965,56 +922,148 @@ partial dictionary FederatedCredentialRequestOptions {
...
+
+[Exposed=Window, SecureContext]
+partial interface FederatedCredential {
+  Promise<FederatedAccount> login(optional FederatedAccountRequest request = {});
+};
+
+[Exposed=Window, SecureContext]
+dictionary FederatedAccount {
+   required USVString id;
+   required USVString idToken;
```
samuelgoto, line 1037: Should this be optional?

spec/index.bs
```diff
@@ -965,56 +922,148 @@ partial dictionary FederatedCredentialRequestOptions {
...
+};
+</xmp>
+
+<dl dfn-type="argument" dfn-for="FederatedAccountRequest">
+    :   <dfn>hint</dfn>
+    ::  The accountId to auto-select, or |undefined|. If |undefined| is provided
+          then the account chooser will be shown.
```
samuelgoto, line 1044: "the account chooser will be shown" isn't entirely correct, right? it won't show if the user is not logged in to the IDP right?

spec/index.bs
```diff
@@ -254,61 +254,75 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
 
-    Welcome <span id="name"></span>!!
+  Welcome <span id="name"></span>.
 
-    <button onclick="revoke">delete account</button>
+  <script>
+  let account = undefined;
 
-    <script>
+  const conn = await navigator.credentials.get({
```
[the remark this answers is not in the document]
dj2: Sure, don't feel strongly either way. Done.

spec/index.bs

spec/index.bs
dj2: Shifted to a helper `getIdFromIdToken`

spec/index.bs
dj2: They're part of the `FederatedIdentityProvider` which is passed to `get`. But, `relyingPartUrl` is actually already covered earlier in the spec when dealing with IDP login.

spec/index.bs
dj2: Duplicate for the `url` in `FederatedIdentityProvider` above. Got confused which object I was putting these into.

spec/index.bs

spec/index.bs
dj2: I'm not sure. Do we want to require an id_token to always be returned? I guess we don't? Made optional.

spec/index.bs
dj2: True. Changed `will` to `MAY`, better?

spec/index.bs
samuelgoto: optional sgtm.

spec/index.bs
samuelgoto: better

samuelgoto (commented):
LGTM

@cbiesinger @yi-gu want to take a second pass before I merge?

spec/index.bs
```diff
@@ -834,7 +850,7 @@ the exchange of the user's identity.
 For [=RP=]s, it allows them to:
 
 1. The [[#browser-api-sign-in]] allows [=RP=]'s users to [[#use-cases-sign-up]]
-    and [[#use-cases-sign-up]]
+    and [[#use-cases-sign-in]]
```
samuelgoto, line 853: I'm inclined to call these use cases `login` and `logout` in the future, so that they match the API structure. Not for this CL, of course, but just wanted to make a mental note.

spec/index.bs
```diff
@@ -965,56 +926,143 @@ partial dictionary FederatedCredentialRequestOptions {
...
+to intermediate the relationship with the [=Identity Provider=] and the
+provisioning of an [=id token=].
+
+The [=Relying Party=] makes no delineation between Sign-up and Sign-in, but
+rather calls the same API indistinguishably.
+
+The [=Relying Party=] can control a variety of UX knobs. For
+example an `AbortSignal` in case it needs to abort the flow.
+
+If all goes well, the [=Relying Party=] receives back a |user| object which
```
samuelgoto, line 1004: |user| => |account|

spec/index.bs
```diff
@@ -965,56 +926,143 @@ partial dictionary FederatedCredentialRequestOptions {
...
+
+[Exposed=Window, SecureContext]
+partial interface FederatedCredential {
+  Promise<FederatedAccount> login(optional FederatedAccountLoginRequest request = {});
+};
+
+[Exposed=Window, SecureContext]
+dictionary FederatedAccount {
+   required USVString id;
+   USVString idToken;
```
samuelgoto, line 1036: right, this feels right to be optional, because we expect this to be undefined when later we introduce `scopes`.

spec/index.bs
```diff
@@ -1232,15 +1336,12 @@ path: img/mock31.svg
...
-  await FederatedCredential.logoutRPs([{
-      url: "https://rp1.example",
-      accountId: "123",
-    }, {
-      url: "https://rpN.example",
-      accountId: "456",
-    }]);
-}
+await FederatedCredential.logoutRPs([
+  { url: "https://rp1.example",
```
samuelgoto, line 1340: nit picking, but I prefer the other indentation style, with the objects starting and ending:

```js
async function logout() {
  await FederatedCredential.logoutRPs([{
      url: "https://rp1.example",
      accountId: "123",
    }, {
      url: "https://rpN.example",
      accountId: "456",
    }]);
}
```

samuelgoto (commented):
LGTM

Taking another pass at the full text. None of this blocks this review.

spec/index.bs
```diff
@@ -254,61 +254,79 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
 
-    Welcome <span id="name"></span>!!
+  Welcome <span id="name"></span>.
 
-    <button onclick="revoke">delete account</button>
+  <script>
+  let account = undefined;
 
-    <script>
+  const federation = await navigator.credentials.get({
```
yi-gu, line 271: "federation" seems a little bit abstract. Is this intentional?

samuelgoto: yeah, we couldn't figure out a better term :) agreed, not ideal. the other two that occurred to was were:

- `auth` and
- `connection`

cbiesinger: Why not name this credential / federated_credential..?

spec/index.bs
cbiesinger: probably a mild preference to include Origin, so that callers don't provide a full path

spec/index.bs
```diff
@@ -1124,29 +1175,82 @@ To <dfn>mint</dfn> a new id token, run the following algorithm:
...
+  required USVString picture;
+  required USVString id;
+
+  USVString privacyPolicyUrl;
+  USVString termsOfServiceUrl;
+};
+
+[Exposed=Window, SecureContext]
+partial interface FederatedCredential {
+  Promise<undefined> authorize(FederatedAuthorizationRequest request);
```
cbiesinger, line 1220: What is this for...? I didn't think we had a use case for such a function? And why are TOS/PP provided by the RP and not by the IDP's response?

spec/index.bs
```diff
@@ -1171,8 +1273,8 @@ static method:
 
 <xmp class=idl>
 [Exposed=Window, SecureContext]
-partial interface FederatedCredential {
-  static Promise<undefined> revoke(USVString accountId, FederatedIdentityProvider provider);
+partial interface FederatedAccount {
```
cbiesinger, line 1276: Why move this to the Account? It seems like you should be able to call this on the credential?

spec/index.bs
```diff
@@ -965,56 +926,143 @@ partial dictionary FederatedCredentialRequestOptions {
...
-  USVString nonce;
 };
 </xmp>
 
-And finally, this specification overrides the {{FederatedCredential}}'s
-<code><dfn for="FederatedCredential" method>\[[DiscoverFromExternalSource]](
-origin, options, sameOriginWithAncestors)</dfn></code> Method.
+<dl dfn-type="argument" dfn-for="FederatedIdentityProvider">
+    :   <dfn>url</dfn>
+    ::  The URL of the identity provider origin.
```
yi-gu, line 934: "Origin" does not include path. e.g. for URL `https://www.example.com:443/foo`, the "origin" is `https://www.example.com:443`.
Perhaps just "The URL of the identity provider"?

spec/index.bs
```diff
@@ -965,56 +926,143 @@ partial dictionary FederatedCredentialRequestOptions {
...
+contains an [=id token=] in the form of a signed [[JWT]] which it can use to
+authenticate the user.
+
+<div class=example>
+```js
+const account = federation.login({
+  hint: "dan@example.com",
+  signal: signal,
+  nonce: "my_nonce"
+
```
yi-gu, line 1014: nit: remove empty line

spec/index.bs
```diff
@@ -1232,15 +1336,12 @@ path: img/mock31.svg
 
 <div class=example>
 ```js
-async function logout() {
-  await FederatedCredential.logoutRPs([{
```
samuelgoto, line 1236: Something occurred to me ... is there a reason this needs to be a static method? should we also put this under the FederatedCredential prototype instead? e.g.

```js
const federation = await navigator.credentials.get({
  federated: {
     // maybe that's a bad idea, because what would go in here?
   }
});

federation.logoutRPs(...);
```

spec/index.bs
```diff
@@ -1124,29 +1175,82 @@ To <dfn>mint</dfn> a new id token, run the following algorithm:
 ### IDP Sign-in API ### {#browser-api-idp-sign-in}
 <!-- ============================================================ -->
 
-The IDP Sign-in API allows an [=IDP=] to sign the user into a [=RP=] by
-storing a {{FederatedCredential}} that can later be retrieved by a [=RP=].
+The IDP Sign-in API allows an [=IDP=] to create a credential for the user in
+a given [=RP=] that can later be retrieved by the [=RP=].
 
-Its effect is equivalent to the [[#browser-api-rp-sign-in]] in its effect,
+Its equivalent to the [[#browser-api-rp-sign-in]] in its effect,
```
yi-gu, line 1181: nit: It's

spec/index.bs
samuelgoto: @dj2 maybe it is best if we remove the `authorize()` method from this initial PR, so that it unblocks the origin trial?

spec/index.bs
samuelgoto: Should we remove this entirely from this PR so that we can look at the `authorize()` call separately from what we need for origin trials?

spec/index.bs
yi-gu: +1 to the pp/tos concern. They should always be provided by the IDP.
As for `authorize` (and in general everything that is mentioned in the spec but not planned for OT), we need to put together a list of such features if any for devrels to better communicate the scope of OT.

dj2: The `authorize` call is meant to be called from the IDP, not the RP. I'll remove into another PR.

spec/index.bs
dj2: The credential doesn't have the `accountId`, the account does.

So, we either put this on the `FederatedCredential` and require `accountId` as a parameter or put to on `FederatedAccount` and have no parameter. Putting it on the latter seems like the better match?

spec/index.bs
dj2: You don't need the `clientId` so you're basically creating an empty `FederatedCredential` for no purpose?

spec/index.bs
dj2: To be fair, I don't like either but the one I do like takes too much space, heh. Reverted.

spec/index.bs

spec/index.bs

spec/index.bs

spec/index.bs

spec/index.bs
dj2: ack

spec/index.bs
dj2: Removed the IDP sign-in section.

spec/index.bs
dj2: I'm not sure if `credential` / `federated_credential` work much better.

spec/index.bs
```diff
@@ -873,89 +891,32 @@ path: img/mock42.svg
...
-</xmp>
+The most important parameter to the API call is the set of
+[=Identity Provider=]s that the [=Relying Party=] supports and has
+pre-registered with (i.e. it has a `clientId`).
 
-<dl dfn-type="attribute" dfn-for="FederatedCredential">
-    :   {{idToken}}
-    ::  The [=id token=].
-
-    :   {{approvedBy}}
```
samuelgoto, line 951: Are we not using this at all in implementation (i think the answer is "no", but check?)? if yes, don't you need to move this into the `FederatedAccount` object?

samuelgoto (commented):
LGTM++

Thanks for simplifying the patch here by removing the `authorize()` part.

spec/index.bs
dj2: @yi-gu @cbiesinger do we set `approvedBy` anywhere in Chrome currently? I'd removed it from here as we don't have auto-signin at the moment.

yi-gu: We only have it in the idl but not yet used it anywhere.

samuelgoto: Removing LGTM.

spec/index.bs
samuelgoto: Yeah, agreed, a bit awkward ... will revisit this later.

yi-gu (commented):
LGTM

samuelgoto (dismissed):

spec/index.bs
cbiesinger: The problem with putting it on the latter is that you need to call login() to revoke a token and I don't think RPs will (want to) do that.

spec/index.bs
cbiesinger: But the object is of type FederatedCredential...

spec/index.bs
```diff
@@ -965,56 +926,142 @@ partial dictionary FederatedCredentialRequestOptions {
...
+
+<!-- ============================================================ -->
+## The Sign-in API ## {#browser-api-sign-in}
+<!-- ============================================================ -->
+
+<!-- ============================================================ -->
+### RP Sign-in API ### {#browser-api-rp-sign-in}
+<!-- ============================================================ -->
+
+The Sign-up and Sign-in APIs used by the [=Relying Party=]s to ask the browser
```
cbiesinger, line 994: Just so I understand better -- is the expectation that the browser shows a prompt for get() or for login()? (and if the former, why not put the account id on the credential...)

samuelgoto: The latter.

spec/index.bs
samuelgoto: They can use login({silent: true})

spec/index.bs
dj2: True, was more from a length perspective.

spec/index.bs
dj2: Why not call `login` before revoke? You're going to need to have called `login` to get the tokens in the first place and it can be called silently as @samuelgoto pointed out.

spec/index.bs
dj2: done.

spec/index.bs
cbiesinger: But even with silent, it has to make an HTTP request to get the token, and there is no reason why you would need a token to revoke a credential...

samuelgoto: Why would it need to make an HTTP request?

dj2: Because login will request an `id_token`.

samuelgoto: Oh yeah, I agree that we shouldn't be making an HTTP request in order to revoke. I was imagining that `silent` would fail if there wasn't an account that was already available. 

Maybe we need to expose something like 

```
partial interface FederatedCredential {
  promise<bool> isLoggedIn(int accountId);
  promise<bool> getAccount(int accountId);
}
```

or something along those lines at the `FederatedCredential` level?

dj2: Done.

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+          hint: opts.hint
+        }]
+      }
+    });
+  }
 
-    var user;
+  function tryGetCredential() {
+    const id = getLoggedInAccountId();
+    if (id === undefined) {
```
samuelgoto, line 308: if (!id) {
  return undefined;
}

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
-      });
-
-      // ... do stuff ...
-      // ... send to server ...
-      // ... validate ...
-      // ... update local storage ...
-
-      user = parse(idToken);
-      document.getElementById("name").innerText = user.name;
+  function try_login() {
```
samuelgoto, line 317: tryLogin() {

}

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+          url: "https://idp.example",
+          clientId: "123",
+          hint: opts.hint
+        }]
+      }
+    });
+  }
 
-    var user;
+  function tryGetCredential() {
```
samuelgoto, line 306: call this `isLoggedIn()` and make it return a boolean instead (and make `getCredential` take hint?)?

these `try*` methods are throwing me off.

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
-      // ... do stuff ...
-      // ... send to server ...
-      // ... validate ...
-      // ... update local storage ...
-
-      user = parse(idToken);
-      document.getElementById("name").innerText = user.name;
+  function try_login() {
+    const credential = tryGetCredential();
+    if (credential !== undefined) {
```
samuelgoto, line 319: if (credential) {
  doLogin();
}

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+    }
+  }
+
+  function login() {
+    const id = getLoggedInAccountId();
+    const credential = await getCredential({
+      hint: id,
+      mediation: "optional"
+    });
+    if (credential !== undefined) {
```
samuelgoto, line 330: if (!credential) {

}

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+      tokens = doLogin(credential);
```
samuelgoto, line 331: can we ignore other tokens in this PR and call of these tokens `idToken` instead?

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
     }
+  }
 
-    async function logout() {
-      await navigator.credentials.preventSilentAccess();
+  function switch() {
```
samuelgoto, line 335: `switch` is a reserved JS keyword, this won't compile.

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+
+  Welcome <span id="name"></span>.
+
+  <script>
+  let tokens = undefined;
+
+  try_login();
+
+  function getLoggedInAccountId() {
+    /* Site specifc logic to retrieve the accountId or undefined if none set */
```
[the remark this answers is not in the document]
dj2: done.

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
 
-    Welcome <span id="name"></span>!!
+  function setLoggedInAccountId(id) {
+    /* Site specific logic to store the accountId */
+    document.cookie = `id=#{id}`
+  }
 
-    <button onclick="revoke">delete account</button>
+  function idFromIdToken(idToken) {
+    return JSON.parse(Base64.decode(idToken.split(“.”)[1])).sub;
```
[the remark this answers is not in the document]
dj2: done.

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+  }
 
-    <script>
+  async function getCredential(opts) {
+    return navigator.credentials.get({
+      mediation: opts.mediation
+      federated: {
+        providers: [{
+          url: "https://idp.example",
+          clientId: "123",
```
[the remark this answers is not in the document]
dj2: The `get` call has to get the `client_id_metadata` in order to display the user dialog. So, it has to be here.

spec/index.bs
dj2: `tryGet` actually returns a `credential` instead of a `bool`. `getCredential` does take a hint. This is wrapping up the logic to determine the hint and set `mediation: "silent"`

spec/index.bs

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+    const credential = tryGetCredential();
+    if (credential !== undefined) {
+      tokens = doLogin(credential);
+    }
+  }
+
+  function login() {
+    const id = getLoggedInAccountId();
+    const credential = await getCredential({
+      hint: id,
```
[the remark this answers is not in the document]
dj2: If you have multiple accounts signed in, this would return that one directly if available.

spec/index.bs
dj2: I'd prefer to have this expandable to allow other tokens without changing the signature in the followup.

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+
+  try_login();
+
+  function getLoggedInAccountId() {
+    /* Site specifc logic to retrieve the accountId or undefined if none set */
+    const parts = document.cookie
+                    .split(';')
+                    .find(row => row.trim().startsWith('id='));
+    if (parts !== undefined) {
+      return parts.split('=')[1];
```
yi-gu, line 279: Perhaps providing an example of `parts` for better readability

spec/index.bs
dj2: doh, fixed.

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+
+  <script>
+  let tokens = undefined;
+
+  try_login();
+
+  function getLoggedInAccountId() {
+    /* Site specifc logic to retrieve the accountId or undefined if none set */
+    const parts = document.cookie
+                    .split(';')
```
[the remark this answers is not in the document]
dj2: Removed.

spec/index.bs
dj2: Removed.

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+<head>
+  <title>Welcome to my WebSite</title>
+</head>
+<body>
+  <button onclick="login()">Login with Foo</button>
+  <button onclick="revoke()">revoke tokens</button>
+  <button onclick="switch()">switch</button>
+  <button onclick="logout()">logout</button>
+
+  Welcome <span id="name"></span>.
```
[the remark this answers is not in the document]
dj2: Good catch, removed.

spec/index.bs
```diff
@@ -125,77 +125,77 @@ Upon user agreement an [=account=] is created with the [=Relying Party=] and
 the user has a [=session=] initialized.
 
 <!-- ============================================================ -->
-### Sign-in ### {#use-cases-sign-in}
```
samuelgoto, line 128: It feels awkward now that we still have the "Sign-up" use case. Is there a specific reason why you changed the use cases from Sign-in to Login and Sign-out to Logout?

spec/index.bs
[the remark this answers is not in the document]
samuelgoto: ah, interesting ... right ... maybe we'll have to make `client_id` to be an `openid` agnostic term then ...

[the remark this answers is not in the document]
yi-gu: We need `clientId` to fetch client_metadata in `get`, right?

spec/index.bs
samuelgoto: Yep, i understand that `tryGet()` returns a credential rather than a bool, but I'm having a hard time parsing all of the `try*` methods and what they do. can you make them more explicit with their intent (i.e. name them what is it that they are trying to do)?

My suggestion is something along the lines of:

```js
function isLoggedIn() {
  // returns whether the user has an account or not
}

function getAccount() {
  // returns the user's account. throws if one doesn't exist.
}
```

spec/index.bs
samuelgoto: I don't feel strongly, but it seems substantially shorter.

I'm trying to make the example as simple as it can possibly be so that readers don't over index on the specifics between `==` and `===` and the signature of `getLoggedInAccountId()` and rather focus on the parts that matter most (i.e. `FederatedCredential` and the `navigator.credentials.get`)?

samuelgoto: I think the same applies to the `try*` methods: we want this to be as verbose and simple as possible, rather than concise and dependent on parameters/arguments.

spec/index.bs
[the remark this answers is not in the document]
samuelgoto: `getCredential` is super hard to read, it does a lot of magic inside and covers too many cases ...

spec/index.bs
samuelgoto: This isn't a signature right? This is a variable name?

spec/index.bs
```diff
@@ -254,61 +254,114 @@ When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
...
+    if (credential !== undefined) {
+      tokens = doLogin(credential);
     }
+  }
 
-    async function revoke() {
-      await FederatedCredential.revoke(user.id, identityProvider);
+  function doLogin(credential) {
+    const tokens = await credential.login({
+      nonce: "456"
```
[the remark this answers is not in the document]
dj2: It's optional in the IDL, I don't think the example needs it marked as such.

spec/index.bs
dj2: Because having the sections named `sign-in` and `sign-out` and the api calls being `login` and `logout` was confusing. This makes it more consistent with the API.

spec/index.bs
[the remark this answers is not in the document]
dj2: I don't think we should change the term as it's the same as the clientId. Using a different name will just make things more confusing.

spec/index.bs
dj2: I could add more words, but `getCredential` gets the credential and `tryGetCredential` tries to get a credential if the user has one already. I don't like `isLoggedIn` because the answer isn't boolean.

spec/index.bs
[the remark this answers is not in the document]
dj2: Why? It calls `credential.get` and optionally passes a hint and the mediation type? I could remove it but it adds a bunch of duplicated content and makes the example harder to read.

spec/index.bs
dj2: What isn't a signature? `tokens` is a variable defined above.

spec/index.bs
dj2: We always combined them into a single api call, it was just called `get` before.

spec/index.bs
cbiesinger: these don't look like the right quote marks

spec/index.bs
```diff
@@ -833,17 +884,19 @@ the exchange of the user's identity.
 
 For [=RP=]s, it allows them to:
 
-1. The [[#browser-api-sign-in]] allows [=RP=]'s users to [[#use-cases-sign-up]]
-    and [[#use-cases-sign-up]]
+1. The [[#browser-api-login]] allows [=RP=]'s users to [[#use-cases-sign-up]]
```
cbiesinger, line 928: btw why the search&replace of sign-in -> login?

spec/index.bs
dj2: That's been removed.

spec/index.bs
dj2: Was trying to match the API methods. Having the API for `login` in a section called `Browser sign in` seemed strange.

samuelgoto: I don't mean to make an imposition here, but I think it would make this patch cleaner if you reverted those changes. I made these same comments myself earlier. It is a big API change, and trying to review editorial vs substantial changes makes it harder for us to find convergence.

spec/index.bs
```diff
@@ -965,56 +1001,84 @@ partial dictionary FederatedCredentialRequestOptions {
 dictionary FederatedIdentityProvider {
   required USVString url;
   required USVString clientId;
-  USVString nonce;
+  USVString hint;
```
samuelgoto, line 1004: You can probably change this in a different patch, but introducing "hint" makes the name of this dictionary a bit more awkward (than it already was). I'd do that in a separate CL, but just wanted to point out here.

samuelgoto (approved):
LGTM++

@cbiesinger @yi-gu  any last comments before I merge?

spec/index.bs
```diff
@@ -251,64 +251,152 @@ algorithms to allow the exchange of identity between [=IDP=]s and [=RP=]s.
...
+  async function login() {
+    const credential = await navigator.credentials.get({
+      federated: {
+        providers: [{
+          url: "https://idp.example",
+          clientId: "123"
+        }]
+      }
+    });
+    return await credential.login({ nonce: "456" });
```
yi-gu, line 281: Is `credential.login` defined in this PR? I only see it gets mentioned in the IDL.

dj2: That's the `RP Sign-in API` section

spec/index.bs
```diff
@@ -873,89 +967,31 @@ path: img/mock42.svg
...
-      }],
-    }
-  });
-
-  return idToken;
-}
+const credential = await navigator.credentials.get({
+  federated: {
+    providers: [{
+      url: "https://idp.example.com",
```
yi-gu, line 980: nit: remove `.com` here and below

yi-gu: I think we have been consistently using `idp.example` in the spec, right?

spec/index.bs
yi-gu: Right. Sorry for the noise

spec/index.bs

spec/index.bs
```diff
@@ -251,64 +251,152 @@ algorithms to allow the exchange of identity between [=IDP=]s and [=RP=]s.
...
+    });
+    return await credential.login({ nonce: "456" });
+  }
+
+  async function logout() {
+    const credential = getLoggedInCredential();
+    if (credential) {
+      return await credential.logout();
+    }
+    return undefined;
```
samuelgoto, line 289: i think this is implied if you leave the function without a return

spec/index.bs
```diff
@@ -251,64 +251,152 @@ algorithms to allow the exchange of identity between [=IDP=]s and [=RP=]s.
...
+  }
+
+  async function revoke() {
+    const credential = getLoggedInCredential();
+    if (credential) {
+      return await credential.revoke();
     }
+    return undefined;
```
samuelgoto, line 297: same here

spec/index.bs
```diff
@@ -251,64 +251,152 @@ algorithms to allow the exchange of identity between [=IDP=]s and [=RP=]s.
...
+      }
+    });
+    const tokens = await credential.login({ nonce: "456" });
+    /* site stores id somewhere */
+    return tokens;
+  }
+
+  async function multiLogin() {
+    const credential = await navigator.credentials.get({
+      mediation: "required",
```
samuelgoto, line 355: add a comment somewhere saying that the "mediation: required" is the thing that does the trick?

spec/index.bs
```diff
@@ -251,64 +251,152 @@ algorithms to allow the exchange of identity between [=IDP=]s and [=RP=]s.
...
+    const credential = getLoggedInCredential();
+    if (credential) {
+      return await credential.revoke();
     }
+    return undefined;
+  }
+
+  async function getLoggedInCredential() {
+    return await navigator.credentials.get({
+      mediation: "silent",
```
samuelgoto, line 302: add a comment somewhere here that the trick here is to use "mediation: silent", which doesn't prompt and returns null in case there isn't already an existing account?

samuelgoto (approved):
LGTM++ % nits

@cbiesinger and @yi-gu could you please approve this too before I merge?

samuelgoto:
I'm going to merge this to unblock implementation and leave all of the remaining open bits as editoral things to be done independently.

Squashing and merging, thanks you all!

PR #202, opened 2022-02-16
Move branding information to fedcm.json manifest

pkotwicz:
This PR subsumes https://github.com/fedidcg/FedCM/pull/182

This PR:
- reflects the soon to be landed changes to the branding information. It does not include dark-mode parameters.
- adds more details about the nature of icons listed in the manifest. They should adhere to the spec for "maskable icons" in the Web Manifest spec and should be single-resolution (not .ico files)

Note that I use 'background_color' instead of 'background-color' to be consistent with JSON naming conventions rather than being consistent with CSS.

Marked the change as non-substantial because the W3C check was giving me trouble. (I did connect my github and W3C account)

***
<a href="https://pr-preview.s3.amazonaws.com/pkotwicz/FedCM/pull/202.html" title="Last updated on Feb 17, 2022, 11:24 PM UTC (c7cc098)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/202/9764bae...pkotwicz:c7cc098.html" title="Last updated on Feb 17, 2022, 11:24 PM UTC (c7cc098)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -553,7 +553,7 @@ The <dfn>Icon JSON</dfn> may have the following properties:
         [maskable](https://www.w3.org/TR/appmanifest/#icon-masks) specification.
     :   <dfn>size</dfn> (optional)
     ::  The size of the icon. The icon is assumed to be square and single resolution
-        (not a multi-resolution .ico). The size should be ommitted if the icon is in
+        (not a multi-resolution .ico). The size may be ommitted if the icon is in
         a vector graphic format (like SVG).
 </dl>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -524,16 +524,22 @@ The file is parsed expecting the following properties:
     ::  A set of [=Branding JSON=] options.
 </dl>
 
-The <dfn>Branding JSON</dfn> enables an [=IDP=] to express their branding preferences, which may be used by User Agents to customize the user agent consent prompt.
+The <dfn>Branding JSON</dfn> enables an [=IDP=] to express their branding
+preferences, which may be used by User Agents to customize the user agent
+consent prompt.
 
-Note: The branding preferences are deliberately designed to be high level / abstract (rather than opinionated about a specific UI structure), to enable different [=User Agents=] to offer different UI experiences and for them to evolve independently over time.
+Note: The branding preferences are deliberately designed to be high level
+/ abstract (rather than opinionated about a specific UI structure), to
+enable different [=User Agents=] to offer different UI experiences and
+for them to evolve independently over time.
 
 It may have the following properties:
 <dl dfn-type="argument" dfn-for="manifest_branding">
     :   <dfn>background_color</dfn> (optional)
-    ::  background [=color=] for [=IDP=]-branded widgets such as buttons.
+    ::  Background [=color=] for [=IDP=]-branded widgets such as buttons.
     :   <dfn>color</dfn> (optional)
-    ::  [=color=] for text on widgets with [=background_color=].
+    ::  [=color=] for text on [=IDP=] branded widgets using
+        [=background_color=].
     :   <dfn>icons</dfn> (optional)
     ::  A list of [=Icon JSON=] objects.
 </dl>
@@ -543,8 +549,8 @@ The <dfn>Icon JSON</dfn> may have the following properties:
 <dl dfn-type="argument" dfn-for="manifest_branding_icons">
     :   <dfn>url</dfn> (required)
     ::  The url pointing to the icon image.
-        The icon needs to comply with the [maskable](https://www.w3.org/TR/appmanifest/#icon-masks)
-        specification.
+        The icon needs to comply with the
+        [maskable](https://www.w3.org/TR/appmanifest/#icon-masks) specification.
     :   <dfn>size</dfn> (optional)
     ::  The size of the icon. The icon is assumed to be square and single resolution
         (not a multi-resolution .ico). The size should be ommitted if the icon is in
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -553,7 +553,7 @@ The <dfn>Icon JSON</dfn> may have the following properties:
         [maskable](https://www.w3.org/TR/appmanifest/#icon-masks) specification.
     :   <dfn>size</dfn> (optional)
     ::  The size of the icon. The icon is assumed to be square and single resolution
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -519,9 +519,40 @@ The file is parsed expecting the following properties:
     :   <dfn>id_token_endpoint</dfn> (required)
     ::  A URL that points to an HTTP API that complies with the [[#idp-api-id-token-endpoint]] API.
     :   <dfn>revocation_endpoint</dfn> (required)
-    :: A URL that points to an HTTP API that complies with the [[#idp-api-revocation-endpoint]] API.
+    ::  A URL that points to an HTTP API that complies with the [[#idp-api-revocation-endpoint]] API.
+    :   <dfn>branding</dfn> (optional)
+    ::  A set of [=Branding JSON=] options.
 </dl>
 
+The <dfn>Branding JSON</dfn> enables an [=IDP=] to express their branding preferences, which may be used by User Agents to customize the user agent consent prompt.
```
dj2, line 527: Can you wrap these to ~80 columns to be consistent with the rest of the spec?

spec/index.bs
```diff
@@ -519,9 +519,40 @@ The file is parsed expecting the following properties:
...
+
+Note: The branding preferences are deliberately designed to be high level / abstract (rather than opinionated about a specific UI structure), to enable different [=User Agents=] to offer different UI experiences and for them to evolve independently over time.
+
+It may have the following properties:
+<dl dfn-type="argument" dfn-for="manifest_branding">
+    :   <dfn>background_color</dfn> (optional)
+    ::  background [=color=] for [=IDP=]-branded widgets such as buttons.
+    :   <dfn>color</dfn> (optional)
+    ::  [=color=] for text on widgets with [=background_color=].
```
dj2, line 536: maybe: `[=color=] for text on [=IDP=] branded widgets using [=background_color=].` ?

spec/index.bs
```diff
@@ -519,9 +519,40 @@ The file is parsed expecting the following properties:
...
+The <dfn>Icon JSON</dfn> may have the following properties:
+
+<dl dfn-type="argument" dfn-for="manifest_branding_icons">
+    :   <dfn>url</dfn> (required)
+    ::  The url pointing to the icon image.
+        The icon needs to comply with the [maskable](https://www.w3.org/TR/appmanifest/#icon-masks)
+        specification.
+    :   <dfn>size</dfn> (optional)
+    ::  The size of the icon. The icon is assumed to be square and single resolution
+        (not a multi-resolution .ico). The size should be ommitted if the icon is in
```
dj2, line 550: Could we just say `The size MAY be omitted if the icon is in a vector graphic format` ? Like, could another UA choose to honour the size if provided?

samuelgoto (approved):
LGTM

Thanks Peter!

I'll merge this once @dj2 approves too.

spec/index.bs
pkotwicz: Done

spec/index.bs
pkotwicz: Done

spec/index.bs
```diff
@@ -519,9 +519,40 @@ The file is parsed expecting the following properties:
...
 </dl>
 
+The <dfn>Branding JSON</dfn> enables an [=IDP=] to express their branding preferences, which may be used by User Agents to customize the user agent consent prompt.
+
+Note: The branding preferences are deliberately designed to be high level / abstract (rather than opinionated about a specific UI structure), to enable different [=User Agents=] to offer different UI experiences and for them to evolve independently over time.
+
+It may have the following properties:
+<dl dfn-type="argument" dfn-for="manifest_branding">
+    :   <dfn>background_color</dfn> (optional)
+    ::  background [=color=] for [=IDP=]-branded widgets such as buttons.
```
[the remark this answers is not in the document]
pkotwicz: Done

spec/index.bs
pkotwicz: The correct wording is based on the meaning of "size"

Meaning 1)
IDP accurately reports the size of the icon.

The "size" is used by the user agent to select an icon. The user agent requires the IDP to accurately report the size of the icon. If the IDP claims that a bitmap is 24x24px but the bitmap when downloaded is a different size, the user agent can choose to ignore the bitmap.

In this meaning of "size", SVG images should not specify a size in the branding JSON because SVGs do not have a set size.

Meaning 2)
The "size" key is a hint from the IDP to the user agent.

For instance, it would be legal for an IDP to provide two icons in the branding JSON
[{
  "url": "icon_144x144px.png"
  "size": "32"
 },
 {
  "url": "icon_16x16px",
  "size": "16"
}]

In the above scenario the IDP is setting the size of the 144x144px bitmap as 32x32px in the JSON so that if the user agent wants a 32x32px bitmap the user agent downsamples the 144x144px bitmap instead of upsampling the 16x16px bitmap.

My opinion is that we should go with "Meaning 1". The Web Manifest spec - https://www.w3.org/TR/appmanifest/ - uses "Meaning 1". Dan what do you think?

dj2: Looking at the linked manifest spec they specifically allow SVG sizes to be provided (https://www.w3.org/TR/appmanifest/#declaring-multiple-icons):

```
The developer wants to use an SVG for greater than or equal to 257x257px. They've found that the SVG file looks too blurry at small sizes, even on high-density screens. To deal with this problem, the developer includes an SVG icon that is only used when the dimensions are at least 257px. 
```

spec/index.bs
```diff
@@ -553,7 +553,7 @@ The <dfn>Icon JSON</dfn> may have the following properties:
         [maskable](https://www.w3.org/TR/appmanifest/#icon-masks) specification.
     :   <dfn>size</dfn> (optional)
     ::  The size of the icon. The icon is assumed to be square and single resolution
-        (not a multi-resolution .ico). The size should be ommitted if the icon is in
+        (not a multi-resolution .ico). The size may be ommitted if the icon is in
```
dj2, line 556: nit: omitted

samuelgoto (approved):
LGTM++

Merging

samuelgoto:
@dj2 based on your previous comments, I'm assuming that you are cool with these changes. LMK if I missed something and I'll work with @pkotwicz to address any remaining problems.

I'm squashing and merging.

dj2 (commented):
lgtm

PR #207, opened 2022-02-23
Add IDP sign-in methods.

dj2:
This PR adds methods to allow an IDP to create the credential
from the IDP top-level origin. The `authorize` call will display
the prompt to the user and will create the RPs credential if
accepted.

***
<a href="https://pr-preview.s3.amazonaws.com/dj2/FedCM/pull/207.html" title="Last updated on Feb 23, 2022, 2:15 PM UTC (88e988c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/207/ecfa1d3...dj2:88e988c.html" title="Last updated on Feb 23, 2022, 2:15 PM UTC (88e988c)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1257,6 +1257,86 @@ When this method is invoked, the user agent MUST execute the following algorithm
 
 
 <!-- ============================================================ -->
+### IDP Sign-in API ### {#browser-api-idp-sign-in}
+<!-- ============================================================ -->
+
+The IDP Sign-in API allows an [=IDP=] to create a credential for the user in
+a given [=RP=] that can later be retrieved by the [=RP=].
+
+It's equivalent to the [[#browser-api-rp-sign-in]] in its effect,
+except that it is designed to be called while the user is at the [=IDP=] rather
+than at the [=RP=].
+
+<div class=example>
+```js
+const federation = await navigator.credentials.get({
+  federated: {
+    relyingPartyOrigin: “https://rp.example”
+  }
+});
+
+federation.authorize({
+  name: “dan”
+  email “dan@example.com”
+  accountId: 123
+});
+```
+</div>
+
+<xmp class=idl>
+[exposed=window, SecureContext]
+partial dictionary FederatedIdentityProvider {
+  USVString relyingPartyOrigin;
+};
+
+[Exposed=Window, SecureContext]
+dictionary FederatedAuthorizationRequest {
+  required USVString name;
+  required USVString email;
+  required USVString picture;
+  required USVString id;
+
+  USVString privacyPolicyUrl;
+  USVString termsOfServiceUrl;
+};
+
+[Exposed=Window, SecureContext]
+partial interface FederatedCredential {
+  Promise<undefined> authorize(FederatedAuthorizationRequest request);
+};
+</xmp>
+
+<dl dfn-type="argument" dfn-for="FederatedIdentityProvider">
+    :   <dfn>relyingPartyOrigin</dfn>
+    ::  The relying party origin for which this credential is being created.
+</dl>
+
+<dl dfn-type="argument" dfn-for="FederatedAuthorizationRequest">
+    :   <dfn>name</dfn>
+    ::  The name to display in the dialog
+    :   <dfn>email</dfn>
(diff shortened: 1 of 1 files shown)


dj2:
Why? We can update the spec in the future if we need to change direction but this lets the wider community know what we're currently aiming at and invites broader feedback.

spec/index.bs
```diff
@@ -1256,6 +1256,86 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+
+<div class=example>
+```js
+const federation = await navigator.credentials.get({
+  federated: {
+    relyingPartyOrigin: “https://rp.example”
+  }
+});
+
+federation.authorize({
```
samuelgoto, line 1278: Something feels a bit asymmetric compared to `login()` is that `authorize()` prompts the user whereas in the `login()` case `navigator.credentials.get()` prompts the user.

I don't know how to factor that into here, other than using `navigator.credentials.store()` instead. Here is what it could look like (not clear to me yet what are the merits of each approach, just wanted to note the asymmetry and a few other options):

```js
await navigator.credentials.store(new FederatedCredential({
  relyingPartyOrigin: "https://rp.example",
  name: "dan",
  email: "dan@example.com",
  id: "123",
}));
```

WDYT?

dj2: In my head `get` doesn't prompt `login` and `authorize` do. The `get` request doesn't have all the needed context in order to handle prompting as it doesn't know tokens or scopes being requested. So, unless we start pulling more stuff up into `get` I think the prompting happens in `login` and `authorize`.

samuelgoto: Resolving this comment, since we sent the other PR to clarify the semantics here.

spec/index.bs
```diff
@@ -1256,6 +1256,86 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+
+It's equivalent to the [[#browser-api-rp-sign-in]] in its effect,
+except that it is designed to be called while the user is at the [=IDP=] rather
+than at the [=RP=].
+
+<div class=example>
+```js
+const federation = await navigator.credentials.get({
+  federated: {
+    relyingPartyOrigin: “https://rp.example”
```
samuelgoto, line 1274: relyingPartyOrigin feels like a mouthful ... wouldn't something more concise like "relyingParty" (and assume that it is an origin passed -- throw in case it isn't) be shorter?

spec/index.bs
samuelgoto, line 1278: "authorize" is really overloaded in the context of OIDC (in that there are "authorization servers" and the distinction between "authorization" vs "authentication"), so i'm trying to avoid it. 

A couple of random observations:

- IIUC here, the semantics here are more along the line of "put a prompt in front a user to capture / observe the login to the RP with the IDP account".
- In the StateMachine, as you saw in the other PR, we used the term `registered` and `unregistered`. The other term that I think you are using to represent the same concept has been `consented` vs `unconsented` (which, like I said earlier, feels too legally charged). The other terms that occurred to me too were along the lines of `approved` vs `unapproved`.

A few ideas here:

Given (1), in that what we want is to prompt the user to "login", is there a reason why `login` wouldn't be appropriate here? e.g.

```js
const credential = await navigator.credentials.get({
  federated: {
    relyingParty: "https://rp.example"
  }
});
await credential.login({
  name: "Dan",
  email: "dan@dan.example",
  ...
});
```

The other idea was perhaps to pick a better name, if we don't feel like overloading `login` because it is being called in a different context. Perhaps:


```js
const credential = await navigator.credentials.get({
  federated: {
    relyingParty: "https://rp.example"
  }
});
await credential.register({ // or maybe approve()?
  name: "Dan",
  email: "dan@dan.example",
  ...
});
```

spec/index.bs
dj2: It would be shorter, but the `Origin` on the end makes it a lot clearer for folks reading code to know exactly what to provide.

spec/index.bs
dj2: I don't think I'd overload `login` but `register` could work.

Issue #83, opened 2021-06-16
delegation-oriented api requires "consequence-free" - must be free of entropy bits which could enable global tracking when IDPs and RPs collude


will-bartlett:
In the explainer, it says:

> The delegation-oriented model is very compelling as an end-state for WebID because it solves both tracking problems. Because it does, it doesn't have any unintended tracking consequence.

> Because it is consequence-free, we expect it to be permission-less (the best permission is that one that isn't needed) and comparably better for users with formulations that impose on the user to make a determination on the consequences of being tracked.

**NB:** "Colluding" IDPs and RPs are the norm in many circumstances because IDPs distribute JavaScript SDKs that RPs run on their sites. The code running on the two "colluding" parties is often owned by the same organization.

I am not sure the proposed designs and diagrams for the delegation-oriented flow achieve the ambitions in the explainer. I worry that, even in the most privacy-preserving forms (directed identifiers) they enable colluding IDPs and RPs to undermine the proposed privacy guarantees and obtain global identifiers for the user. If so, enabling a delegation-oriented model on a permission-less basis would undermine rather than promote user privacy.

The key problem is the entropy. Baselines:

 - 128 or 256 bits of entropy is what is commonly recommended for "un-guessable" values (e.g. onetime codes) by security experts and cryptographers
 - 32 bits is enough to identify each human online uniquely
 - 8 bits is useful for tracking, because it can be combined with other values.

The delegated model relies on the user agent presenting the relying party a token that contains a certificate signed by the issuer that contains the directed user identifier. However, some digital signature schemes (e.g. RSAPSS, standardized in [JWA](https://datatracker.ietf.org/doc/html/rfc7518#section-3.5) as PS256) rely on random data. By using a digital signature scheme with random data, but including a global user identifier as some part of the random data instead of fully random random data, and by making this pattern known to RPs, colluding IDPs and RPs could share global user identifiers. Outside of signatures, a IDP interested in collusion could also embed data e.g. in the order of fields in the certificate (e.g. if an order-independent format like a JWT is used) or in the least significant bits of a high precision `valid-until` timestamp. Finally, digital signature schemes must support rollover - keys expire and get replace - so IDPs could embed entropy in the choice of which key they use to sign a given certificate.

[...]

PR #208, opened 2022-02-23
Add tokens and scopes.

dj2:
This PR adds the concept of tokens and scopes into the `login` request.

***
<a href="https://pr-preview.s3.amazonaws.com/dj2/FedCM/pull/208.html" title="Last updated on Feb 23, 2022, 8:16 PM UTC (f053e5a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/208/ecfa1d3...dj2:f053e5a.html" title="Last updated on Feb 23, 2022, 8:16 PM UTC (f053e5a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -248,8 +248,8 @@ entries provided by the [=Identity Provider=].
 
 This specification extends the {{FederatedCredential}} type and internal
 algorithms to allow the exchange of identity between [=IDP=]s and [=RP=]s.
-When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
-[=RP=] can use to authenticate the user.
+When it succeeds, it returns to the [=RP=] tokens. The [=ID Token=] can be
+used by the [=RP=] to authenticate the user.
 
 <!-- ============================================================ -->
 # Single Account Example # {#example-single-user}
@@ -573,7 +573,7 @@ by exposing a series of HTTP endpoints:
 1. A [[#idp-api-manifest]] endpoint in an agreed upon location that points to
 1. An [[#idp-api-accounts-endpoint]] endpoint
 1. A [[#idp-api-client-id-metadata-endpoint]] endpoint
-1. An [[#idp-api-id-token-endpoint]] endpoint
+1. A [[#idp-api-token-endpoint]] endpoint
 1. A [[#idp-api-revocation-endpoint]] endpoint
 
 <!-- ============================================================ -->
@@ -608,8 +608,8 @@ The file is parsed expecting the following properties:
     ::  A URL that points to an HTTP API that complies with the [[#idp-api-accounts-endpoint]] API.
     :   <dfn>client_metadata_endpoint</dfn> (required)
     ::  A URL that points to an HTTP API that complies with the [[#idp-api-client-id-metadata-endpoint]] API.
-    :   <dfn>id_token_endpoint</dfn> (required)
-    ::  A URL that points to an HTTP API that complies with the [[#idp-api-id-token-endpoint]] API.
+    :   <dfn>token_endpoint</dfn> (required)
+    ::  A URL that points to an HTTP API that complies with the [[#idp-api-token-endpoint]] API.
     :   <dfn>revocation_endpoint</dfn> (required)
     ::  A URL that points to an HTTP API that complies with the [[#idp-api-revocation-endpoint]] API.
     :   <dfn>branding</dfn> (optional)
@@ -658,7 +658,7 @@ For example:
 {
   "accounts_endpoint": "/accounts.php",
   "client_metadata_endpoint": "/metadata.php",
-  "id_token_endpoint": "/idtokens.php",
+  "token_endpoint": "/tokens.php",
   "revocation_endpoint": "/revocation.php",
   "branding": {
     "background_color": "green",
@@ -791,12 +791,12 @@ For example:
 </div>
 
 <!-- ============================================================ -->
-## ID Token ## {#idp-api-id-token-endpoint}
+## Tokens ## {#idp-api-token-endpoint}
 <!-- ============================================================ -->
 
-The ID Token endpoint is responsible for [=minting=] a new [=id token=] for the user.
+The Token endpoint is responsible for [=minting=] new tokens for the user.
 
-The ID Token endpoint is fetched
+The Token endpoint is fetched
 (a) as a **POST** request,
 (b) **with** [=IDP=] cookies,
 (c) **with** a [[RFC7231#header.referer|Referer]] header indicating the [=RP=]'s origin
@@ -807,7 +807,7 @@ The ID Token endpoint is fetched
 
 It will also contain the following parameters in the request body `application/x-www-form-urlencoded`:
 
-<dl dfn-type="argument" dfn-for="id_token_endpoint_request">
+<dl dfn-type="argument" dfn-for="token_endpoint_request">
     :   <dfn>client_id</dfn>
     ::  The [=RP=]'s client it
     :   <dfn>nonce</dfn>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -608,8 +608,8 @@ The file is parsed expecting the following properties:
     ::  A URL that points to an HTTP API that complies with the [[#idp-api-accounts-endpoint]] API.
     :   <dfn>client_metadata_endpoint</dfn> (required)
     ::  A URL that points to an HTTP API that complies with the [[#idp-api-client-id-metadata-endpoint]] API.
-    :   <dfn>id_token_endpoint</dfn> (required)
-    ::  A URL that points to an HTTP API that complies with the [[#idp-api-id-token-endpoint]] API.
+    :   <dfn>token_endpoint</dfn> (required)
```
samuelgoto, line 611: A couple of points:

(a) this is going to be backwards incompatible ... which isn't great and
(b) it would be useful to somewhat get inspiration on terminology / requests / responses / etc from the OIDC [token endpoint](https://openid.net/specs/openid-connect-core-1_0.html#TokenEndpoint)

It would be interesting to find a formulation in which (b) is 100% backwards compatible to the OIDC endpoint, because then we could re-use their existing deployments. I don't think it is a mandatory / hard requirement, but if everything fits into that model, it wouldn't be bad to reuse the existing deployment.

Also, would be useful to anticipate the kinds of security affordances that were added to that endpoint so that we don't run into the same walls.

There is also the [authorization endpoint](https://openid.net/specs/openid-connect-core-1_0.html#AuthorizationEndpoint) which I don't think I quite understand the distinction between that and the token endpoint, but worth pointing out.

dj2: Backwards incompatible with what? FedCM or OIDC?

bc-pi: The authorization endpoint in OAuth https://www.rfc-editor.org/rfc/rfc6749.html#section-3.1 is where the client sends the user for interactive authentication and approval of the requested access. 

The token endpoint in OAuth https://www.rfc-editor.org/rfc/rfc6749.html#section-3.2 is used directly by the client to get tokens - typically by presenting an authorization code obtained from the authorization endpoint or a refresh token.

bc-pi: Which is to say that, while I don't know that backwards compatible is even possible here, it would be nice if different terms were used for things which are different than established things in these established protocols.

will-bartlett:
I'm a bit unsure of the need to even enumerate access_token and refresh_token explicitly - these are just names, but in the context of FedCM, they seem to carry little semantics. Which is to say - if an IDP were to return an access_token as a refresh_token or a refresh_token as an access_token, or an access_token inside the id_token - the API would function the same. The names are semantic-less.

It seems like the spec is looking for the following attributes:
1. An id_token is mandatory, a JWT, and contains the OpenID Connect minimal claimset for an id_token.
2. The browser can distinguish a minimally-privacy-revealing request from an unknown-privacy-revealing request - and show a scarier warning in the later case. "The |tokens| and |scopes| MAY affect the dialog presented to the user to makes
clear that more power is being granted in the communication."

The first makes sense for interoperability.

The second, I have doubts about, captured in #83 . Effectively, if the IDP can hide 128 bits of data anywhere, the IDP can be perfectly privacy non-preserving. I believe the IDP can already hide 128 bits of data in an id_token and so distinguishing requests as "less-privacy-preserving" because they use access_token or refresh_token is mostly hurting well-behaved companies, while not impeding bad actors.

There is also a body of existing practice of the scope parameter indicating the response token type. In particular, OpenID Connect defines that the "openid" scope requests that an additional token (the id_token) be returned and that the "offline_access" scope requests that an additional token (the refresh_token) be returned.

With all that in mind, my recommendation would be to drop the "tokens" parameter, drop the accessToken and refreshToken members from FederatedTokens response object, rename "FederatedTokens" to "FederatedTokenResponse" (to indicate that it can contain either token or non-token content) and simply state that the entire JSON response (including unknown parameters) is serialized into a javascript object that must match the FederatedTokenResponse interface. Relying parties can continue to indicate additional tokens with the scope parameter, as per their existing practice. And, if the browser wants to make distinctions about the privacy or requests, the browser can distinguish id_token vs id_token+opaquedata - which I think is the only meaningful technical distinction anyway.

I think recommendation this follows the OAuth pattern. Across the ecosystem, there are various extensions where IDPs can indicate the lifetime of a refresh token - some IDPs (Microsoft) use "refresh_token_expires_in":7776000, others use "refresh_token_expires_on", "rt_expiry", "rt_expires", "rt_expires_in", etc. OAuth made access_token mandatory and basically made everything else optional and out-of-scope - which gave the Identity ecosystem freedom to grow and experiment. Taking a similar choice here (making id_token mandatory and making everything else optional and out-of-scope). Just scanning Microsoft's OAuth implementation, we have optional response parameters for geographic affinity, optional response parameters for extended duration resiliency during outages, optional response parameters for cache keys to improve client caching - it'd be nice to strike a balance where this stuff "just works", but doesn't need to be covered in the spec.

will-bartlett:
A bit of a nit, but scope should be used in the singular here, consistent with [RFC6749](https://datatracker.ietf.org/doc/html/rfc6749). The use of scope is similar to "scope of work" for a project. You might say "the scope of the project that we will complete in the first quarter are text, audio, and video", but you wouldn't say "the **scopes** of the project that we will complete in the first quarter are text, audio, and video" - similarly, we write scope=text+audio+video, not scopes=text+audio+video. The individual items that are space delimited in the scope parameter are called "permissions" by Microsoft and called "strings" and/or "values" by RFC 6749.

spec/index.bs
will-bartlett: The authorization endpoint is used for cases where the user needs to interact with the IDP (in a browser window). If I understand your scoping correctly (https://github.com/fedidcg/FedCM/blob/main/explainer.md), you want to exclude such interactive requests currently (under "out of scope", "Sign into the IDP"). There are two main cases where the user might need to interact with the IDP: 1) initial sign-in to the IDP 2) additional authentication required (e.g. TFA). As a concrete example of additional authentication required, Microsoft requires any sign-in (e.g. password) to sync the browser history in Microsoft Edge. However, to access the [OneDrive personal vault](https://www.microsoft.com/en-us/microsoft-365/onedrive/personal-vault), the sign-in must include TFA. I think excluding interactive sign-in (and thus the authorization endpoint) is reasonable - you may want to add "additional authentication required" as an out of scope case in the explainer as well.

Connecting to OAuth token. Microsoft defines an artifact called the [primary refresh token](https://docs.microsoft.com/en-us/azure/active-directory/devices/concept-primary-refresh-token). It's similar to an OAuth refresh token, but useable for all clients, not just one clients, because it is owned by the IDP. Here, you've effectively done the same thing, but with cookies.

Consider if you start from an OAuth refresh token request that would be valid for a public client under [RFC 6749 Section 6](https://datatracker.ietf.org/doc/html/rfc6749#section-6):

>      POST /token HTTP/1.1
>      Host: server.example.com
>      Content-Type: application/x-www-form-urlencoded
> 
>      grant_type=refresh_token&refresh_token=tGzv3JOkF0XG5Qx2TlKWIA
>      &client_id=s6BhdRkqt3&scope=profile_short

You move the refresh_token from form data to cookie, and treat the cookie as if it is valid for all clients, not just one client. That gets you to something like:

>      POST /token HTTP/1.1
>      Host: server.example.com
>      Content-Type: application/x-www-form-urlencoded
>      **Cookie: tGzv3JOkF0XG5Qx2TlKWIA**
> 
>      grant_type=**cookie**&client_id=s6BhdRkqt3&scope=profile_short

If you drop grant_type and switch from /token to /tokens.php, and add the security header, you have a FedCM request:

>      POST **/tokens.php** HTTP/1.1
>      Host: server.example.com
>      Content-Type: application/x-www-form-urlencoded
>      **Cookie: tGzv3JOkF0XG5Qx2TlKWIA**
>      Sec-FedCM-CSRF: ?1
> 
>      client_id=s6BhdRkqt3&scope=profile_short

So, to the extent that you want to make this more OAuth-y (which I tend to think is a **good** thing), I think the only change you would really need to make is to add a grant_type parameter and to ensure you carefully follow OAuth semantics (e.g. returning unknown response parameters, not interpreting request scope parameter, etc.).

PR #214, opened 2022-03-02
Update algorithms to match examples.

dj2:
This PR updates the various algorithms to match the intent expressed
in the example usages.

Fixes https://github.com/fedidcg/FedCM/issues/209

***
<a href="https://pr-preview.s3.amazonaws.com/dj2/FedCM/pull/214.html" title="Last updated on Apr 13, 2022, 7:13 PM UTC (67d7bbb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/214/b439d86...dj2:67d7bbb.html" title="Last updated on Apr 13, 2022, 7:13 PM UTC (67d7bbb)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -31,6 +31,8 @@ spec: ecma262; urlPrefix: https://tc39.github.io/ecma262/
 spec: credential-management-1; urlPrefix: https://w3c.github.io/webappsec-credential-management/
     type: dictionary
         text: CredentialRequestOptions; url: dictdef-credentialrequestoptions
+    type: enum
+        text: CredentialMediationRequirement; url: enumdef-credentialmediationrequirement
     for: Credential
         type: method
             text: [[CollectFromCredentialStore]](origin, options, sameOriginWithAncestors)
@@ -272,31 +274,31 @@ could be implemented.
   async function login() {
     const credential = getFederatedCredential();
 
-    // This will prompt when !credential[“consented”], but
-    // it won’t when credential[“consented”].
+    // This will prompt when !credential[“approved”], but
+    // it won’t when credential[“approved”].
     //
-    // If !credential["consented"] and mediation == silent will reject.
+    // If !credential["approved"] and mediation == silent will reject.
     //
-    // If the user selects an account updates the credential["consented"]
+    // If the user selects an account updates the credential["approved"]
     // state and stores any needed account information.
     return await credential.login({ nonce: "456" });
   }
 
   async function logout() {
     const credential = getFederatedCredential();
-    // This never prompts, rejects the promise in case !credential[“consented”]
+    // This never prompts, rejects the promise in case !credential[“approved”]
     return credential.logout();
   }
 
   async function revoke() {
     const credential = getFederatedCredential();
-    // This never prompts, rejects the promise in case !credential[“consented”]
-    return credential.revoke(credential.id);
+    // This never prompts, rejects the promise in case !credential[“approved”]
+    return credential.revoke();
   }
 
   async function getFederatedCredential() {
     // This never prompts. A FederatedCredential object will always be returned.
-    // The object will be either consented or unconsented and store information
+    // The object will be either approved or unapproved and store information
     // on if the user should be prompted based on the mediation flag.
     return await navigator.credentials.get({
       mediated: “optional”, // “optional” is the default
@@ -1022,19 +1024,19 @@ dictionary FederatedIdentityProvider {
 The {{CredentialRequestOptions/mediation}} flag is used to determine if the user
 is prompted to select an account.
 
-  :  silent
+  :  {{CredentialMediationRequirement/silent}}
   :: * If 1 credential exists, return it
      * If > 1 credentials exist, and hint matches one of them return it
-     * Otherwise return unconsented FederatedCredential
-     * Never prompts the user
-  :  optional
+     * Otherwise return an unapproved {{FederatedCredential}}
+     * Rejects the promise of any API call that requires the user to be prompted
+  :  {{CredentialMediationRequirement/optional}}
   :: * If 1 credential exists, return it
      * If > 1 credentials exist, and hint matches one of them return it
-     * Otherwise return unconsented FederatedCredential
-     * Prompts the user if the credential is unconsented
-  :  required
-  :: * Always returns an unconsented FederatedCredential
-      * Always prompts the user
+     * Otherwise return unapproved {{FederatedCredential}}
+     * Prompts the user if the credential is unapproved
+  :  {{CredentialMediationRequirement/required}}
+  :: * Always returns an unapproved {{FederatedCredential}}
+     * Always prompts the user
 
 The {{CredentialRequestOptions/signal}} is used as an abort signal for the
 requests.
@@ -1052,7 +1054,12 @@ requests.
 This specification overrides the {{FederatedCredential}}'s
 <code><dfn for="FederatedCredential" method>
  \[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)</dfn></code>
-method.
+method. As well, several members are defined for {{FederatedCredential}} which
+are internal (non-accessible) properties, these include:
+
+* `approved`
+* `signedIn`
+* `prompts`
 
 This algorithm runs in parallel inside the [[CM#algorithm-request]] to request
 credentials and returns a set of {{FederatedCredential}} for the requested
@@ -1096,20 +1103,20 @@ When this method is invoked, the user agent MUST execute the following algorithm
   1. Let |hint| be |provider|["{{FederatedIdentityProvider/hint}"] or "undefined" if none provided
   1. Let |mode| be |options|["{{CredentialRequestOptions/mediation"] or "optional" if none provided
   1. if |mode| is "required"
-     1. Let |credential| be a new FederatedCredential
-     1. |credential|["consented"] = false
-     1. |credential|["mayPrompt"] = true
+     1. Let |credential| be a new {{FederatedCredential}}
+     1. |credential|["approved"] = false
+     1. |credential|["prompts"] = true
      1. return |credential|
   1. If there is 1 existing |credential|
      1. return |credential|
   1. If there is > 1 existing |credential| and |hint| matches |credential|["id"]
      1. return |credential|
-  1. Let |credential| be a new FederatedCredential
-     1. |credential|["consented"] = false
+  1. Let |credential| be a new {{FederatedCredential}}
+     1. |credential|["approved"] = false
      1. If |mode| is "silent"
-         1. |credential|["mayPrompt"] = false
+         1. |credential|["prompts"] = false
      1. If |mode| is "optional"
-         1. |credential|["mayPrompt"] = true
+         1. |credential|["prompts"] = true
      1. return |credential|
 
 
@@ -1136,7 +1143,7 @@ The Sign-in API will request the [=id token=] to be [=minted=] by the IDP.
 
 <div class=example>
 ```js
-const tokens = credential.login({
+const tokens = await credential.login({
   signal: signal,
   nonce: "my_nonce"
 });
@@ -1174,9 +1181,9 @@ dictionary FederatedTokens {
 </dl>
 
 When this method is invoked, the user agent MUST execute the following algorithm:
-  1. If |credential|["mayPrompt"] is false
+  1. If |credential|["approved"] is false and |credential|["prompts"] is false
      1. Reject promise
-  1. If |credential|["consented"] is false
+  1. If |credential|["approved"] is false
      1. Request user consent by running the [=request-consent=] algorithm.
   1. Return the result of running the [=create-tokens=] algorithm.
 
@@ -1307,7 +1314,7 @@ IDP; for example, it can be an email address.
 
 When this method is invoked, the user agent MUST execute the following algorithm:
 
-  1. If |credential|["consented"] is "false"
+  1. If |credential|["approved"] is "false"
     1. Reject
   1. Let the |revocation_endpoint| url be the relative url
         |manifest|["{{Manifest/revocation_endpoint}}"] of
@@ -1316,6 +1323,8 @@ When this method is invoked, the user agent MUST execute the following algorithm
         cookies.  This request MUST NOT follow
         [[RFC7231#header.location|HTTP redirects]] and instead abort with an
         error if there are any.
+    1. If the HTTP response code is not 2xx
+       1. Reject
     1. Resolve with |undefined|
 
 <!-- ============================================================ -->
@@ -1348,7 +1357,7 @@ partial interface FederatedCredential {
 
 When this method is invoked, the user agent MUST execute the following algorithm:
 
-  1. If |credential|["consented"] is "false"
+  1. If |credential|["approved"] is "false"
     1. Reject
   1. Mark the |credential| to dis-allow Front-channel logout
   1. Mark the |credential| to dis-allow auto-signin
@@ -1414,7 +1423,7 @@ When this method is invoked, the user agent MUST execute the following algorithm
         {{FederatedCredentialLogoutRpsRequest/url}} and
         {{FederatedCredentialLogoutRpsRequest/accountId}}
     1. If no matching |credential| continue
-    1. If |credential|["consented"] = false; continue
+    1. If |credential|["approved"] = false; continue
     1. If |credential|["signedIn"] = false; continue
     1. POST to the {{FederatedCredentialLogoutRpsRequest/url}} with the
         [=Relying Parties=] cookies.  This request MUST NOT follow
@@ -1814,7 +1823,7 @@ invisible to the browser, in which case it could contain anything.
 
 The potential for IDPs profiling users based on their visits to RPs
 [[#attack-scenarios-by-idp-secondary-use]] can be partially mitigated by hiding
-the RP from the IDP until after the user has consented to that tracking risk.
+the RP from the IDP until after the user has agreed to that tracking risk.
 
 However, there is residual risk in cases where the RP and IDP are colluding.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -275,7 +275,7 @@ could be implemented.
     // This will prompt when !credential[“consented”], but
     // it won’t when credential[“consented”].
     //
-    // If !credential["consented"] and mediation == silent will `reject`.
+    // If !credential["consented"] and mediation == silent will reject.
     //
     // If the user selects an account updates the credential["consented"]
     // state and stores any needed account information.
@@ -528,6 +528,14 @@ provides the main security boundary on the web.
     allocated by that site that is used to identify a [=user=] on that site. A
     single user will generally have many user IDs that refer to them, and a single
     site may or may not know that multiple user identifiers refer to the same user.
+
+:  <dfn>Privacy Policy</dfn>
+:: The policies described at
+    {{client_id_metadata_endpoint_response/privacy_policy_url}}.
+
+:  <dfn>Terms of Service</dfn>
+:: The policies described at
+    {{client_id_metadata_endpoint_response/terms_of_service_url}}.
 
 <!-- ============================================================ -->
 # High Level Design # {#high-level-design}
@@ -847,7 +855,6 @@ For example:
 ## Revocation ## {#idp-api-revocation-endpoint}
 <!-- ============================================================ -->
 
-
 The revocation endpoint is responsible for revoking all [=id token=]s for the
 specified client ID for the user.
 
@@ -1015,6 +1022,20 @@ dictionary FederatedIdentityProvider {
 The {{CredentialRequestOptions/mediation}} flag is used to determine if the user
 is prompted to select an account.
 
+  :  silent
+  :: * If 1 credential exists, return it
+     * If > 1 credentials exist, and hint matches one of them return it
+     * Otherwise return unconsented FederatedCredential
+     * Never prompts the user
+  :  optional
+  :: * If 1 credential exists, return it
+     * If > 1 credentials exist, and hint matches one of them return it
+     * Otherwise return unconsented FederatedCredential
+     * Prompts the user if the credential is unconsented
+  :  required
+  :: * Always returns an unconsented FederatedCredential
+      * Always prompts the user
+
 The {{CredentialRequestOptions/signal}} is used as an abort signal for the
 requests.
 
@@ -1025,6 +1046,7 @@ requests.
     ::  The client ID provided to the [=RP=] out of band by the [=IDP=]
     :   <dfn>hint</dfn>
     ::  Optional |accountId| hint. Maybe used by the UA when displaying the
+         account chooser dialog.
 </dl>
 
 This specification overrides the {{FederatedCredential}}'s
@@ -1056,9 +1078,6 @@ This [=internal method=] accepts three arguments:
          [=same-origin with its ancestors=]. It is [FALSE] if caller is cross-origin.
 </dl>
 
-NOTE: This algorithm is currently out of date and will be updated to match the
-examples.
-
 When this method is invoked, the user agent MUST execute the following algorithm:
 
   1.  If <var ignore>sameOriginWithAncestors</var> is `false`, return a
@@ -1074,107 +1093,25 @@ When this method is invoked, the user agent MUST execute the following algorithm
   1. Let |provider| be |options|["{{CredentialRequestOptions/federated}}"]["{{FederatedCredentialRequestOptions/providers}}"][0].
   1. Assert: |provider|["{{FederatedIdentityProvider/url}}"] [=map/exists=].
   1. Assert: |provider|["{{FederatedIdentityProvider/clientId}}"] [=map/exists=].
+  1. Let |hint| be |provider|["{{FederatedIdentityProvider/hint}"] or "undefined" if none provided
   1. Let |mode| be |options|["{{CredentialRequestOptions/mediation"] or "optional" if none provided
-  1. if |mode| is {{CredentialMediationRequirement/required}}
-      1. Return the result of running the [=create-credential=] algorithm.
+  1. if |mode| is "required"
+     1. Let |credential| be a new FederatedCredential
+     1. |credential|["consented"] = false
+     1. |credential|["mayPrompt"] = true
+     1. return |credential|
+  1. If there is 1 existing |credential|
+     1. return |credential|
+  1. If there is > 1 existing |credential| and |hint| matches |credential|["id"]
+     1. return |credential|
+  1. Let |credential| be a new FederatedCredential
+     1. |credential|["consented"] = false
+     1. If |mode| is "silent"
+         1. |credential|["mayPrompt"] = false
+     1. If |mode| is "optional"
+         1. |credential|["mayPrompt"] = true
+     1. return |credential|
 
-  1. if |mode| is {{CredentialMediationRequirement/silent}}
-      1. If |options|["{{FederatedIdentityProvider/hint}}"] does not [=map/exists=]
-         1. Reject
-      1. If a {{FederatedCredential}} with id == |options|["{{FederatedIdentityProvider/hint}}"] exists
-         1. Return credential
-      1. Reject
-
-  1. If a single {{FederatedCredential}} exists, return it
-  1. Otherwise, return the result of running teh [=create-credential=] algorithm.
-
-To <dfn>create-credential</dfn> run this algorithm:
-    1. Let |manifest| be the result of running the [=fetch the manifest=]
-        algorithm with the |provider|.
-    1. Let |accounts list| be the result of running the
-        [=fetch the accounts list=] algorithm.
-    1. Let |account| be the result of running the [=select an account=] from the
-        |accounts list| algorithm.
-    1. If the {{StateMachine/Account State}} is [=unregistered=]
-        1. Return the result of running the [=sign-up=] algorithm.
-    1.  If the {{StateMachine/Session State}} is  [=logged out=] then
-        1. Return the result of running the [=explicit sign-in=] algorithm.
-    1. Return the result of running the [=auto sign-in=] algorithm.
-
-To <dfn>sign-up</dfn> the user run this algorithm:
-    1. Gather explicit intent to create an account
-    1. Gather explicit agreement with the [=Privacy Policy=]
-    1. Gather explicit agreement with the [=Terms of Service=]
-    1. Set the account's {{StateMachine/Account State}} from [=unregistered=]
-        to [=registered=].
-    1. Set the account's {{StateMachine/Session State}} from [=logged out=]
-        to [=logged in=].
-    1. Return a {{FederatedCredential}} object
-
-To <dfn>select an account</dfn> from the accounts list run this algorithm:
-    1. If |accounts list|'s [=list/size=] is 1
-        1. Let |account| be |accounts list|[0]
-        1. Return |account|
-    1. If the |hint| is provided and an account matches |hint|
-        1. Return |account|
-    1. Display an account chooser
-    1. Let |account| be an account that the user manually selects from the
-        accounts chooser
-    1. Return |account|
-
-To <dfn>sign-in</dfn> the user run this algorithm:
-    1. Assert that the {{StateMachine/Account State}} is [=registered=]
-    1. Set the account's {{StateMachine/Session State}} from [=logged out=] to
-        [=logged in=].
-    1. Return a {{FederatedCredential}} object
-
-To <dfn>auto sign-in</dfn> the user run this algorithm:
-    1. Return the result of running the [=sign-in=] algorithm.
-
-To <dfn>explicit sign-in</dfn> the user run this algorithm:
-    1. Gather explicit confirmation that the user wants to sign-in to
-        [=Relying Party=] with their [=Identity Provider=]'s |account|
-    1. Return the result of running the [=sign-in=] algorithm.
-
-To <dfn>fetch the manifest</dfn> configuration file, run this algorithm:
-    1. Run a [[!CSP]] check with a [[CSP#directive-connect-src|connect-src]]
-        directive on the URL passed as |provider|["{{FederatedIdentityProvider/url}}"]
-        and abort this algorithm if the check fails.
-    1. Let the |manifest url| be the file "fedcm.json" relative to the directory
-        prefix passed as |provider|["{{FederatedIdentityProvider/url}}"]
-    1. Return the result of fetching the |manifest url| with the
-        [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]] header but without the
-        [=Identity Provider=]'s cookies.  This request MUST NOT follow
-        [[RFC7231#header.location|HTTP redirects]] and instead abort with an
-        error if there are any.
-
-To <dfn>fetch the accounts list</dfn> run this algorithm:
-    1. Let the |accounts_endpoint| url be the relative url
-        |manifest|["{{Manifest/accounts_endpoint}}"] of
-        |provider|["{{FederatedIdentityProvider/url}}"]
-    1. Let the |accounts list| be the result of fetching the |accounts_endpoint|
-        with the [=Identity Provider=]'s cookies.  This request MUST NOT follow
-        [[RFC7231#header.location|HTTP redirects]] and instead abort with an
-        error if there are any.
-    1. Return the |accounts list|
-
-To <dfn noexport>fetch the client metadata</dfn> run this algorithm:
-    1. Let the |client_metadata_endpoint| url be the relative url
-        |manifest|["{{Manifest/client_metadata_endpoint}}"] of
-        |provider|["{{FederatedIdentityProvider/url}}"]
-    1. Let the |policies| be the result of fetching the
-        |client_metadata_endpoint| with the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]]
-        header but without the [=Identity Provider=]'s cookies.  This request
-        MUST NOT follow [[RFC7231#header.location|HTTP redirects]] and instead
-        abort with
-        an error if there are any.
-    1. Return |policies|
-
-The <dfn>Privacy Policy</dfn> are the policies described at
-{{client_id_metadata_endpoint_response/privacy_policy_url}}.
-
-The <dfn>Terms of Service</dfn> are the policies described at
-{{client_id_metadata_endpoint_response/terms_of_service_url}}.
 
 <!-- ============================================================ -->
 ## The Sign-in API ## {#browser-api-sign-in}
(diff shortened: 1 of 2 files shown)


dj2:
Please give this a close read over. I think I got the algorithms correct, but I thought that last time too ....

spec/index.bs
```diff
@@ -1014,6 +1021,20 @@ dictionary FederatedIdentityProvider {
 The {{CredentialRequestOptions/mediation}} flag is used to determine if the user
 is prompted to select an account.
 
+  :  silent
```
samuelgoto, line 1024: Linkify this to something like [CredentialRequestOption.silent](https://www.w3.org/TR/credential-management-1/#dom-credentialrequestoptions-mediation) so that it is clear that this is a parameter in the CredentialRequestOptions object rather than the FederatedCredentialRequestOptions or FederatedCredential object?

spec/index.bs
```diff
@@ -1014,6 +1021,20 @@ dictionary FederatedIdentityProvider {
...
+  :: * If 1 credential exists, return it
+     * If > 1 credentials exist, and hint matches one of them return it
+     * Otherwise return unconsented FederatedCredential
```
samuelgoto, line 1027: I know that we used those terms "consented" and "unconsented" in our discussions, but it feels like a really charged term with legal meaning.

Maybe `approve` or `unapproved` would be lighter yet still accurate?

samuelgoto, line 1027: linkify FederatedCredential?

samuelgoto, line 1027: As a comparison to line #1033, maybe add:

> Throws / rejects the promises of any API call that requires the user to be prompted.

spec/index.bs
```diff
@@ -1073,32 +1092,105 @@ When this method is invoked, the user agent MUST execute the following algorithm
   1. Let |provider| be |options|["{{CredentialRequestOptions/federated}}"]["{{FederatedCredentialRequestOptions/providers}}"][0].
   1. Assert: |provider|["{{FederatedIdentityProvider/url}}"] [=map/exists=].
   1. Assert: |provider|["{{FederatedIdentityProvider/clientId}}"] [=map/exists=].
+  1. Let |hint| be |provider|["{{FederatedIdentityProvider/hint}"] or "undefined" if none provided
```
samuelgoto, line 1095: Isn't this already the case? Does it need to be said explicilty?

spec/index.bs
```diff
@@ -1073,32 +1092,105 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
   1. Let |mode| be |options|["{{CredentialRequestOptions/mediation"] or "optional" if none provided
-  1. if |mode| is {{CredentialMediationRequirement/required}}
-      1. Return the result of running the [=create-credential=] algorithm.
+  1. if |mode| is "required"
+     1. Let |credential| be a new FederatedCredential
```
samuelgoto, line 1098: linkify FederatedCredential here and elsewhere?

spec/index.bs
```diff
@@ -1073,32 +1092,105 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+     1. |credential|["consented"] = false
```
samuelgoto, line 1099: Is there some spec kung-fu that we could use to say that ["consented"] is an internal (non-accessible) property or "credential"? 

maybe define that somewhere and explain what it is and what it does and how it is used?

spec/index.bs
```diff
@@ -1073,32 +1092,105 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+     1. |credential|["mayPrompt"] = true
```
samuelgoto, line 1100: nipicking: is there something shorter/less camel-casy so that it reads more consistently with `consented`? maybe `prompts` rather than `mayPrompt`.

spec/index.bs
samuelgoto, line 1098: don't we need to save somewhere in the |credential| object certain properties that are going to be used later in `login()`?

spec/index.bs
```diff
@@ -1073,32 +1092,105 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+
+If all goes well, the [=Relying Party=] receives back a |tokens| object which
+contains an [=id token=] in the form of a signed [[JWT]] which it can use to
+authenticate the user.
+
+The Sign-in API will request the [=id token=] to be [=minted=] by the IDP.
+
+<div class=example>
+```js
+const tokens = credential.login({
```
samuelgoto, line 1138:
```js
const tokens = await credential.log({
...
});
```

spec/index.bs
```diff
@@ -1073,32 +1092,105 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+<xmp class=idl>
+[Exposed=Window, SecureContext]
+dictionary FederatedAccountLoginRequest {
+  AbortSignal signal;
+  USVString nonce;
+};
+
+[Exposed=Window, SecureContext]
+partial interface FederatedCredential {
+  Promise<FederatedTokens> login(optional FederatedAccountLoginRequest request = {});
```
samuelgoto, line 1154: Maybe we should call this FederatedAccountLoginResponse instead?

spec/index.bs
```diff
@@ -1073,32 +1092,105 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+<dl dfn-type="argument" dfn-for="FederatedTokens">
+    :   <dfn>idToken</dfn>
+    ::  The minted ID token
+</dl>
 
-  1. If a single {{FederatedCredential}} exists, return it
-  1. Otherwise, return the result of running teh [=create-credential=] algorithm.
+When this method is invoked, the user agent MUST execute the following algorithm:
+  1. If |credential|["mayPrompt"] is false
+     1. Reject promise
```
samuelgoto, line 1177: ?

If |credential|["consented"] wouldn't this be able to continue without prompting?

spec/index.bs
```diff
@@ -1073,32 +1092,105 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+  1. If |credential|["consented"] is false
+     1. Request user consent by running the [=request-consent=] algorithm.
```
samuelgoto, line 1179: like i said earlier, `consent` feels a bit legally heavy / binding to me ... maybe `approval`?

spec/index.bs
```diff
@@ -1073,32 +1092,105 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
 
-To <dfn>create-credential</dfn> run this algorithm:
+To <dfn>request-consent</dfn> run this algorithm:
     1. Let |manifest| be the result of running the [=fetch the manifest=]
         algorithm with the |provider|.
     1. Let |accounts list| be the result of running the
         [=fetch the accounts list=] algorithm.
     1. Let |account| be the result of running the [=select an account=] from the
         |accounts list| algorithm.
     1. If the {{StateMachine/Account State}} is [=unregistered=]
```
samuelgoto, line 1189: i have an intuition that the {{StateMachine}} stuff is probably not useful anymore or at least should be revisited as you update these algorithms too ...

at least, [=unregistered=] and [=registered=] in AccountState, is really close to what you are calling "consented" or not.

spec/index.bs
samuelgoto: ah, speaking of which, `registered` and `unregistered` could work too.

samuelgoto: but i think it has something to do with what we discussed in the past as "the browser observed the sign-up".

spec/index.bs
```diff
@@ -1279,7 +1301,18 @@ partial interface FederatedCredential {
 };
 </xmp>
 
-Note: go over the revocation API.
+When this method is invoked, the user agent MUST execute the following algorithm:
+
+  1. If |credential|["consented"] is "false"
```
samuelgoto, line 1306: I think "true" and "false" is defined somewhere.

spec/index.bs
```diff
@@ -1279,7 +1301,18 @@ partial interface FederatedCredential {
...
+    1. Reject
+  1. Let the |revocation_endpoint| url be the relative url
+        |manifest|["{{Manifest/revocation_endpoint}}"] of
+        |provider|["{{FederatedIdentityProvider/url}}"]
+    1. Fetch the |revocation_endpoint| with the [=Identity Provider=]'s
+        cookies.  This request MUST NOT follow
+        [[RFC7231#header.location|HTTP redirects]] and instead abort with an
+        error if there are any.
+    1. Resolve with |undefined|
```
samuelgoto, line 1315: Should this resolve with the HTTP response so that the RP can know whether it 200 or 500 or 400?

spec/index.bs
```diff
@@ -1309,7 +1342,13 @@ partial interface FederatedCredential {
 };
 </xmp>
 
-Note: go over the RP sign-out API.
+When this method is invoked, the user agent MUST execute the following algorithm:
+
+  1. If |credential|["consented"] is "false"
+    1. Reject
+  1. Mark the |credential| to dis-allow Front-channel logout
```
samuelgoto, line 1349: it disallows front-channel logout for this specific account id, though, not globally for this entire RP, right?

spec/index.bs
```diff
@@ -1309,7 +1342,13 @@ partial interface FederatedCredential {
...
+  1. Mark the |credential| to dis-allow auto-signin
```
samuelgoto, line 1350: something that i think is confusing is whether |credential| is an object that is "stored" or "remembered" or "can be marked" or rather that's just an in-memory object that has no persistent state. 

I think you are implying here more of the former rather than the latter, but i don't think we've made that separation consistently.

spec/index.bs
```diff
@@ -1364,7 +1403,22 @@ partial interface FederatedCredential {
...
-Note: go over how this is implemented.
+When this method is invoked, the user agent MUST execute the following algorithm:
+
+  1. For each request in |logoutRequests|
+    1. Let |credential| be a credential matching
+        {{FederatedCredentialLogoutRpsRequest/url}} and
+        {{FederatedCredentialLogoutRpsRequest/accountId}}
+    1. If no matching |credential| continue
+    1. If |credential|["consented"] = false; continue
+    1. If |credential|["signedIn"] = false; continue
```
samuelgoto, line 1414: ah, there is another "state" in |credential| ... yeah, i think these different states need to be defined and written down somewhere where we can see all of them holistically, what values they take, and what each of these states mean.

like i said before, i think there is some intersection with the state machine states, so i think it is probably the right time to look at it again and unify here.

spec/index.bs
dj2: Do you mean the "undefined" if none provided part? The definition of `|hint|` isnt' already the case

spec/index.bs
dj2: It's a login Request, not a Response though?

spec/index.bs
dj2: Good catch, will fix.

spec/index.bs
dj2: Maybe reject if !204 response code?

spec/index.bs
dj2: Yes, this specific credential.

spec/index.bs
dj2: There is persistant state for a credential, things like if there is an active session that existed prior to this change. We need to collect them up and record them better though.

spec/index.bs
samuelgoto: Possibly, but maybe any !2xx would work?

spec/index.bs

spec/index.bs
dj2: Don't know how to do it in bikeshed, added a blurb above the algorithm. It's not great, but a start at least.

spec/index.bs

spec/index.bs
dj2: done

spec/index.bs
dj2: ack

spec/index.bs
```diff
@@ -1073,32 +1099,105 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-  1. if |mode| is {{CredentialMediationRequirement/required}}
-      1. Return the result of running the [=create-credential=] algorithm.
+  1. if |mode| is "required"
+     1. Let |credential| be a new {{FederatedCredential}}
+     1. |credential|["approved"] = false
+     1. |credential|["prompts"] = true
+     1. return |credential|
+  1. If there is 1 existing |credential|
+     1. return |credential|
+  1. If there is > 1 existing |credential| and |hint| matches |credential|["id"]
```
samuelgoto, line 1111: do we set |credential|["id"] somewhere?

spec/index.bs
samuelgoto: I meant wrt to `FederatedTokens`. Major nit, but should it be calledn `FederatedAccountLoginResponse`?

spec/index.bs
samuelgoto: sgtm

spec/index.bs
```diff
@@ -1169,88 +1266,20 @@ To <dfn noexport>fetch the client metadata</dfn> run this algorithm:
...
+            |account|["{{accounts_endpoint_response_accounts/account_id}}"].
+        * Let the |id_token_endpoint_request| be a new object with
+           * {{id_token_endpoint_request/client_id}} as
+              |provider|["{{FederatedIdentityProvider/clientId}}"].
+           * {{id_token_endpoint_request/nonce}} as
+              |provider|["{{FederatedAccountLoginRequest/nonce}}"].
+        * This request MUST NOT follow [[RFC7231#header.location|HTTP redirects]]
+            and instead abort with an error if there are any.
+    1. Return a new {{FederatedTokens}} as:
+        * {{FederatedTokens/idToken}} as |token|
```
samuelgoto, line 1282: The result of the POST request is of the form:

{
  "id_token": "foo bar"
}

I think you want to say something like:

{{FederatedTokens/idToken}} as |token|["id_token"]

spec/index.bs
```diff
@@ -1279,7 +1308,20 @@ partial interface FederatedCredential {
 };
 </xmp>
 
-Note: go over the revocation API.
+When this method is invoked, the user agent MUST execute the following algorithm:
+
+  1. If |credential|["approved"] is "false"
```
samuelgoto, line 1313: do you need to check the |credential|["loggedin"] property to?

spec/index.bs
samuelgoto: I guess, what's the mechanism here? I was expecting to see something like:

1. Set |credential|["loggedin"] to false

So that it is later used in the front-channel logout algorithm:

```
1. For each request in |logoutRequests|
    1. Let |credential| be a credential matching
        {{FederatedCredentialLogoutRpsRequest/url}} and
        {{FederatedCredentialLogoutRpsRequest/accountId}}
    1. If |credential|["signedIn"] = false; continue
```

spec/index.bs
samuelgoto: total bikeshedding here, but the camel-casing is bothering me here compared to "approved" and "prompts". can we call this "signed-in" instead (or "signedin" or "logged-in" or "loggedin") ?

spec/index.bs
```diff
@@ -1014,6 +1023,20 @@ dictionary FederatedIdentityProvider {
 The {{CredentialRequestOptions/mediation}} flag is used to determine if the user
 is prompted to select an account.
 
+  :  {{CredentialMediationRequirement/silent}}
+  :: * If 1 credential exists, return it
+     * If > 1 credentials exist, and hint matches one of them return it
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -1024,12 +1047,18 @@ requests.
...
     ::  Optional |accountId| hint. Maybe used by the UA when displaying the
+         account chooser dialog.
 </dl>
 
 This specification overrides the {{FederatedCredential}}'s
 <code><dfn for="FederatedCredential" method>
  \[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)</dfn></code>
-method.
+method. As well, several members are defined for {{FederatedCredential}} which
+are internal (non-accessible) properties, these include:
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -1024,12 +1047,18 @@ requests.
...
+
+* `approved`
```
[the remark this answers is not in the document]

spec/index.bs

spec/index.bs
dj2: The ID is part of the `FederatedCredentialInit` object used to create the `FederatedCredential` (It's inherited from `CredentialData`) so it has to be set in order to create the credential.

spec/index.bs
dj2: Maybe? I lean towards `FederatedTokens` as it's clear what it holds/represents.

spec/index.bs
dj2: Good catch. Fixed.

spec/index.bs

spec/index.bs
dj2: Yup, fixed.

spec/index.bs
dj2: "authenticated"? "active"?

samuelgoto: Either one of those work for me.

spec/index.bs
samuelgoto: This isn't developer-visible, so I don't feel strongly about it. Resolving to try to get this merged sooner.

spec/index.bs
```diff
@@ -1073,32 +1099,105 @@ When this method is invoked, the user agent MUST execute the following algorithm
   1. Let |provider| be |options|["{{CredentialRequestOptions/federated}}"]["{{FederatedCredentialRequestOptions/providers}}"][0].
   1. Assert: |provider|["{{FederatedIdentityProvider/url}}"] [=map/exists=].
   1. Assert: |provider|["{{FederatedIdentityProvider/clientId}}"] [=map/exists=].
+  1. Let |hint| be |provider|["{{FederatedIdentityProvider/hint}"]
   1. Let |mode| be |options|["{{CredentialRequestOptions/mediation"] or "optional" if none provided
-  1. if |mode| is {{CredentialMediationRequirement/required}}
-      1. Return the result of running the [=create-credential=] algorithm.
+  1. if |mode| is "required"
+     1. Let |credential| be a new {{FederatedCredential}}
```
samuelgoto, line 1105: don't you need to save into the |credential| the |provider| such that you can later retrieve from it?

spec/index.bs
```diff
@@ -1169,88 +1266,20 @@ To <dfn noexport>fetch the client metadata</dfn> run this algorithm:
...
-       * Let {{id_token_endpoint_request/client_id}} be
-          |provider|["{{FederatedIdentityProvider/clientId}}"].
-       * Let the {{id_token_endpoint_request/nonce}} be
-          |provider|["{{FederatedAccountLoginRequest/nonce}}"].
-    * This request MUST NOT follow [[RFC7231#header.location|HTTP redirects]]
-        and instead abort with an error if there are any.
-1. Return a new {{FederatedTokens}} as:
-    * {{FederatedTokens/idToken}} as |token|
-
+To <dfn>create tokens</dfn> run this algorithm:
```
samuelgoto, line 1269: s/<dfn>create token</dfn>/<dfn>create-token</dfn>

<= the [=create-token=] link above is broken unless you do so i believe.

samuelgoto: on a related note, check your bikeshed log/warnings for more broken links?

spec/index.bs
```diff
@@ -1073,32 +1099,105 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
         [=fetch the accounts list=] algorithm.
     1. Let |account| be the result of running the [=select an account=] from the
         |accounts list| algorithm.
     1. If the {{StateMachine/Account State}} is [=unregistered=]
-        1. Return the result of running the [=sign-up=] algorithm.
+        1.Run the [=sign-up=] algorithm.
     1.  If the {{StateMachine/Session State}} is  [=logged out=] then
-        1. Return the result of running the [=explicit sign-in=] algorithm.
-    1. Return the result of running the [=auto sign-in=] algorithm.
+        1. Run the [=explicit sign-in=] algorithm.
```
samuelgoto, line 1199: where are you setting the credential["approved"] bit? i don't see that in the [=explicit sign-in=] algorithm or elsewhere?

same question for credential["signed-in"].

samuelgoto, line 1199: doesn't this need to "return" something? i.e. should we use the term "return the result of running the algorithm" rather than "run"?

samuelgoto, line 1199: i was following the [=explicit sign-in=] algorithm, and it feels entirely broken too.

Step (3), specifically, seems to be returning an "account" if there is just one, which seems totally broken.

```
1. If accounts list’s [size](https://infra.spec.whatwg.org/#list-size) is 1
2. Let account be accounts list[0] 
3. Return account 
4. If the hint is provided and an account matches hint 
5. Return account
6. Display an account chooser
7. Let account be an account that the user manually selects from the accounts chooser
8. Return account
```

spec/index.bs
```diff
@@ -1169,88 +1266,20 @@ To <dfn noexport>fetch the client metadata</dfn> run this algorithm:
...
+            |account|["{{accounts_endpoint_response_accounts/account_id}}"].
+        * Let the |id_token_endpoint_request| be a new object with
+           * {{id_token_endpoint_request/client_id}} as
+              |provider|["{{FederatedIdentityProvider/clientId}}"].
+           * {{id_token_endpoint_request/nonce}} as
+              |provider|["{{FederatedAccountLoginRequest/nonce}}"].
+        * This request MUST NOT follow [[RFC7231#header.location|HTTP redirects]]
+            and instead abort with an error if there are any.
+    1. Return a new {{FederatedTokens}} as:
+        * {{FederatedTokens/idToken}} as |token|["id_token"]
```
samuelgoto, line 1282: don't we need to set in the |credential|["id"] = |account|["id"] so that later the other algorithms work?

as well as |credential|["approved"] = true and |credential|["signed-in"] = true?

samuelgoto (commented):
Took another pass. Still missing a lot of the "state writes" so that the "state reads" make sense.

spec/index.bs
```diff
@@ -1237,21 +1189,113 @@ dictionary FederatedTokens {
...
+
+To <dfn>request consent</dfn> run this algorithm:
+    1. Let |manifest| be the result of running the [=fetch the manifest=]
+        algorithm with the |credential|["provider"].
+    1. Let |accounts list| be the result of running the
+        [=fetch the accounts list=] algorithm.
+    1. Let |account| be the result of running the [=select an account=] from the
+        |accounts list| algorithm.
+    1. If the {{StateMachine/Account State}} is [=unapproved=]
+        1. Run the [=sign-up=] algorithm for |account|.
```
npm1, line 1210: If this is the last step we need to do here, we need to add ', then return', otherwise let's get rid of the single-substeps here as well.

spec/index.bs
```diff
@@ -1237,21 +1189,113 @@ dictionary FederatedTokens {
...
+    1. Set the account's {{StateMachine/Session State}} from [=inactive=]
+        to [=active=].
+    1. Set |credential|["id"] to |account|["id"}]
 
+To <dfn>select an account</dfn> from the accounts list run this algorithm:
+    1. If |accounts list|'s [=list/size=] is 1
+        1. Let |account| be |accounts list|[0]
+        1. Return |account|
+    1. If the |hint| is provided and an account matches |hint|
+        1. Return |account|
```
npm1, line 1233: This is not C++, you can merge the single substep returns here

spec/index.bs
```diff
@@ -1237,21 +1189,113 @@ dictionary FederatedTokens {
...
+              |provider|["{{FederatedAccountLoginRequest/nonce}}"].
+        * This request MUST NOT follow [[RFC7231#header.location|HTTP redirects]]
+            and instead abort with an error if there are any.
+    1. Return a new {{FederatedTokens}} with:
+        * {{FederatedTokens/idToken}} as |token|["id_token"]
+
+To <dfn>fetch the manifest</dfn> configuration file, run this algorithm:
+    1. Run a [[!CSP]] check with a [[CSP#directive-connect-src|connect-src]]
+        directive on the URL passed as |provider|["{{FederatedIdentityProvider/url}}"]
+        and abort this algorithm if the check fails.
```
npm1, line 1269: s/abort this algorithm/return

spec/index.bs
```diff
@@ -1283,7 +1327,22 @@ partial interface FederatedCredential {
 The hint parameter is passed directly to the IDP. Its meaning is defined by the
 IDP; for example, it can be an email address.
 
-Note: go over the revocation API.
+When this method is invoked, the user agent MUST execute the following algorithm:
+
+  1. If {{StateMachine/Account State}} is [=unapproved=] or
+      {{StateMachine/Session State}} is [=inactive=]
+    1. Reject promise
```
npm1, line 1334: merge

spec/index.bs
```diff
@@ -1283,7 +1327,22 @@ partial interface FederatedCredential {
...
+  1. Let the |revocation_endpoint| url be the relative url
+        |manifest|["{{Manifest/revocation_endpoint}}"] of
+        |provider|["{{FederatedIdentityProvider/url}}"]
+    1. Fetch the |revocation_endpoint| with the [=Identity Provider=]'s
+        cookies.  This request MUST NOT follow
+        [[RFC7231#header.location|HTTP redirects]] and instead abort with an
+        error if there are any.
+    1. If the HTTP response code is not 2xx
+       1. Reject
```
npm1, line 1343: Maybe merge (>2 indentation not working anyways)

spec/index.bs
```diff
@@ -1313,7 +1372,12 @@ partial interface FederatedCredential {
 };
 </xmp>
 
-Note: go over the RP sign-out API.
+When this method is invoked, the user agent MUST execute the following algorithm:
+
+  1. If {{StateMachine/Account State}} is [=unapproved=]
+    1. Reject
+  1. Set {{StateMachine/Session State}} to [=inactive=]
```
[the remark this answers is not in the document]
dj2: done

spec/index.bs
```diff
@@ -1368,7 +1432,22 @@ partial interface FederatedCredential {
 };
 </xmp>
 
-Note: go over how this is implemented.
+When this method is invoked, the user agent MUST execute the following algorithm:
+
+  1. For each request in |logoutRequests|
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -1368,7 +1432,22 @@ partial interface FederatedCredential {
...
+    1. Let |credential| be a credential matching
+        {{FederatedCredentialLogoutRpsRequest/url}} and
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -1368,7 +1432,22 @@ partial interface FederatedCredential {
...
+        {{FederatedCredentialLogoutRpsRequest/accountId}}
+    1. If no matching |credential| continue
+    1. If the |credential| {{StateMachine/Account State}} is [=unapproved=] or
+        {{StateMachine/Session State}} is [=inactive=]
+    1. POST to the {{FederatedCredentialLogoutRpsRequest/url}} with the
```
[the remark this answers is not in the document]
dj2: Missing a `continue` above.

spec/index.bs
```diff
@@ -1368,7 +1432,22 @@ partial interface FederatedCredential {
...
+        [=Relying Parties=] cookies.  This request MUST NOT follow
+        [[RFC7231#header.location|HTTP redirects]] and instead abort with an
+        error if there are any.
+    1. Set the |credential| {{StateMachine/Session State}} to [=inactive=]
+  1. Resolve with |undefined|
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -1074,107 +1115,25 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
-        |provider|["{{FederatedIdentityProvider/url}}"]
-    1. Let the |policies| be the result of fetching the
-        |client_metadata_endpoint| with the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]]
-        header but without the [=Identity Provider=]'s cookies.  This request
-        MUST NOT follow [[RFC7231#header.location|HTTP redirects]] and instead
-        abort with
-        an error if there are any.
-    1. Return |policies|
+  1. Let |hint| be |provider|["{{FederatedIdentityProvider/hint}}"]
+  1. Let |mode| be |options|["{{CredentialRequestOptions/mediation}}"] or "optional" if none provided
```
npm1, line 1119: You can remove the last part (`or "optional" if none provided`), as that is already what happens from the IDL.

spec/index.bs
```diff
@@ -1074,107 +1115,25 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
+     1. Let |credential| be a new {{FederatedCredential}}
+     1. Set |credential|["provider"] = |provider|
+     1. Set {{StateMachine/Account State}} to [=approved=]
+     1. Return |credential|
+  1. If there is 1 existing credential in the credential store
+     1. Return the credential from the credential store
+  1. If there is > 1 existing credential in the credential store and |hint| matches the credential["id"]
+     1. Return the credential from the credential store
+  1. Let |credential| be a new {{FederatedCredential}}
+     1. Set |credential|["provider"] = |provider|
```
npm1, line 1132: Two things: I think this notation is not used because this isn't a map. Maybe `Set |credential|'s {{FederatedCredential/provider}} to |provider|`. But also isn't the left a `UVString` and the right a `DOMString or FederatedIdentityProvider`? How do you assign this?

spec/index.bs
```diff
@@ -1237,21 +1196,109 @@ dictionary FederatedTokens {
 </dl>
 
 When this method is invoked, the user agent MUST execute the following algorithm:
+  1. If {{StateMachine/Account State}} is [=unapproved=] and
+      {{StateMachine/Prompt State}} is [=disabled=] then reject promise
```
npm1, line 1200: nit: similar to below, create a promise, reject it, and return it.

spec/index.bs
```diff
@@ -1237,21 +1196,109 @@ dictionary FederatedTokens {
...
+    1. Set the account's {{StateMachine/Account State}} from [=unapproved=]
+        to [=approved=].
+    1. Set the account's {{StateMachine/Session State}} from [=inactive=]
+        to [=active=].
+    1. Set |credential|["id"] to |account|["id"}]
+
+To <dfn>select an account</dfn> from the accounts list:
+    1. If |accounts list|'s [=list/size=] is 1:
+        1. Let |account| be |accounts list|[0] and return |account|
+    1. If the |hint| is provided and an account matches |hint| then return |account|
```
npm1, line 1236: Unclear where |hint| comes from. It looks like it needs to be plumbed from somewhere

spec/index.bs
```diff
@@ -1237,21 +1189,113 @@ dictionary FederatedTokens {
...
+    1. Let |tokens| be the result of making a POST request to the
+        |manifest|["{{Manifest/id_token_endpoint}}"] with:
+        * Let the {{id_token_endpoint_request/account_id}} be
+            |account|["{{accounts_endpoint_response_accounts/account_id}}"].
+        * Let the |id_token_endpoint_request| be a new object with
+           * {{id_token_endpoint_request/client_id}} as
+              |provider|["{{FederatedIdentityProvider/clientId}}"].
+           * {{id_token_endpoint_request/nonce}} as
+              |provider|["{{FederatedAccountLoginRequest/nonce}}"].
+        * This request MUST NOT follow [[RFC7231#header.location|HTTP redirects]]
```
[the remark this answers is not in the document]
npm1: I guess this is probably related to fetch integration

spec/index.bs
```diff
@@ -1237,21 +1196,109 @@ dictionary FederatedTokens {
...
-          |provider|["{{FederatedIdentityProvider/clientId}}"].
-       * Let the {{id_token_endpoint_request/nonce}} be
-          |provider|["{{FederatedAccountLoginRequest/nonce}}"].
-    * This request MUST NOT follow [[RFC7231#header.location|HTTP redirects]]
-        and instead abort with an error if there are any.
-1. Return a new {{FederatedTokens}} as:
-    * {{FederatedTokens/idToken}} as |token|
+To <dfn>fetch the accounts list</dfn>:
+    1. Let the |accounts_endpoint| url be the relative url
+        |manifest|["{{Manifest/accounts_endpoint}}"] of
```
npm1, line 1283: Needs |manifest| and |provider| inputs?

spec/index.bs
```diff
@@ -1313,7 +1375,12 @@ partial interface FederatedCredential {
 };
 </xmp>
 
-Note: go over the RP sign-out API.
+When this method is invoked, the user agent MUST execute the following algorithm:
```
npm1, line 1378: ditto promise problem: this algorithm must return a promise.

spec/index.bs
```diff
@@ -953,10 +979,10 @@ an internal [[#browser-api-state-machine]].
 ## The State Machine ## {#browser-api-state-machine}
 <!-- ============================================================ -->
 
-Internally, the account management and session management APIs allows the user
-to explicitly move through the different stages of their accounts. At each
-stage, the state machine manages the [=RP=]'s and the [=IDP=]'s access to the
-appropriate browser capabilities.
+Internally, the FedCM API manages a set of |credential|s which are associated
+with an [=RP=] and [=IDP=] pair. Those |credential| objects have different
+stages of there accounts. At each stage, the state machine manages the |credential|s
```
npm1, line 984:
```suggestion
stages of their accounts. At each stage, the state machine manages the |credential|'s
```

spec/index.bs
```diff
@@ -1237,21 +1189,113 @@ dictionary FederatedTokens {
...
+
+To <dfn>explicit sign-in</dfn> the user run this algorithm:
+    1. Gather explicit confirmation that the user wants to sign-in to
+        [=Relying Party=] with their [=Identity Provider=]'s |account|
+    1. Run the [=sign-in=] algorithm.
+
+To <dfn>create tokens</dfn> run this algorithm:
+    1. Assert {{StateMachine/Account State}} is [=approved=]
+    1. Assert {{StateMachine/Session State}} is [=active=]
+    1. Let |tokens| be the result of making a POST request to the
```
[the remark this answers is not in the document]
dj2: better?

spec/index.bs
```diff
@@ -1237,21 +1196,109 @@ dictionary FederatedTokens {
...
+        accounts chooser
+    1. Return |account|
+
+To <dfn>sign-in</dfn> the user:
+    1. Assert that the {{StateMachine/Account State}} is [=approved=]
+    1. Set the account's {{StateMachine/Session State}} from [=inactive=] to
+        [=active=].
+
+To <dfn>explicit sign-in</dfn> the user:
+    1. Gather explicit confirmation that the user wants to sign-in to
```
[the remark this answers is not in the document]
dj2: Fixed

spec/index.bs
```diff
@@ -1237,21 +1196,109 @@ dictionary FederatedTokens {
...
+            and instead abort with an error if there are any.
+    1. Return a new {{FederatedTokens}} with:
+        * {{FederatedTokens/idToken}} as |tokens|["id_token"]
+
+To <dfn>fetch the manifest</dfn> configuration file:
+    1. Run a [[!CSP]] check with a [[CSP#directive-connect-src|connect-src]]
+        directive on the URL passed as |provider|["{{FederatedIdentityProvider/url}}"]
+        and return if the check fails.
+    1. Let the |manifest url| be the file "fedcm.json" relative to the directory
+        prefix passed as |provider|["{{FederatedIdentityProvider/url}}"]
```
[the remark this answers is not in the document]
dj2: done

spec/index.bs
dj2: Is there a way those inputs are expected to be spec'd?

spec/index.bs
dj2: done

npm1 (approved):
@samuelgoto if you're ok with it I think we can land this since @dj2 is moving on. We can work on more improvements on separate PRs but this seems like an improvement so I propose landing.

npm1:
Ping @samuelgoto?

spec/index.bs
```diff
@@ -965,11 +984,14 @@ path: img/mock42.svg
 
 <dl dfn-type="argument" dfn-for="StateMachine">
     :   <dfn>Account State</dfn>
-    ::  Keeps track of whether the user has an account or not. Can be
-         [=registered=] or (by default) [=unregistered=].
+    ::  Keeps track of whether the user has an account or not.
```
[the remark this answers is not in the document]
samuelgoto: ping

spec/index.bs
```diff
@@ -952,10 +971,10 @@ an internal [[#browser-api-state-machine]].
...
-Internally, the account management and session management APIs allows the user
-to explicitly move through the different stages of their accounts. At each
-stage, the state machine manages the [=RP=]'s and the [=IDP=]'s access to the
-appropriate browser capabilities.
+Internally, the FedCM API manages a set of |credential|s which are associated
+with an [=RP=] and [=IDP=] pair. Those |credential| objects have different
+stages of there accounts. At each stage, the state machine manages the |credentials|
+access to the appropriate browser capabilities.
 
 <div class='image'>
```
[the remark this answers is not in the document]
samuelgoto: ping

spec/index.bs
```diff
@@ -965,11 +984,14 @@ path: img/mock42.svg
...
-         [=registered=] or (by default) [=unregistered=].
+    ::  Keeps track of whether the user has an account or not.
+         Can be [=approved=] or [=unapproved=] (by default).
     :   <dfn>Session State</dfn>
-    ::  Keeps track of whether the user has an open or closed session. Can be
-         [=logged in=] or (by default) [=logged out=].
+    ::  Keeps track of whether the user has an open or closed session.
+         Can be [=active=] or [=inactive=] (by default).
+    :   <dfn>Prompt State</dfn>
+    ::  Keeps track if prompting the user was requested.
```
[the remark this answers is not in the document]
samuelgoto: ping

spec/index.bs
```diff
@@ -1015,6 +1044,20 @@ dictionary FederatedIdentityProvider {
...
 
+  :  {{CredentialMediationRequirement/silent}}
+  :: * If 1 credential exists, return it
+     * If > 1 credentials exist, and hint matches a credential `id` return it
+     * Otherwise return an unapproved {{FederatedCredential}}
+     * Rejects the promise of any API call that requires the user to be prompted
+  :  {{CredentialMediationRequirement/optional}}
+  :: * If 1 credential exists, return it
+     * If > 1 credentials exist, and hint matches one of them return it
+     * Otherwise return unapproved {{FederatedCredential}}
```
samuelgoto, line 1055: Linkify [=unapproved=] to point to the definition in the state machine?

spec/index.bs
```diff
@@ -1015,6 +1044,20 @@ dictionary FederatedIdentityProvider {
...
+     * Prompts the user if the credential is unapproved
```
samuelgoto, line 1056: Linkify [=unapproved=] to point to the definition in the state machine?

spec/index.bs
```diff
@@ -1015,6 +1044,20 @@ dictionary FederatedIdentityProvider {
...
+  :  {{CredentialMediationRequirement/required}}
+  :: * Always returns an unapproved {{FederatedCredential}}
```
samuelgoto, line 1058: Linkify [=unapproved=] to point to the definition in the state machine?

spec/index.bs
```diff
@@ -1015,6 +1044,20 @@ dictionary FederatedIdentityProvider {
 The {{CredentialRequestOptions/mediation}} flag is used to determine if the user
 is prompted to select an account.
 
+  :  {{CredentialMediationRequirement/silent}}
+  :: * If 1 credential exists, return it
+     * If > 1 credentials exist, and hint matches a credential `id` return it
+     * Otherwise return an unapproved {{FederatedCredential}}
```
samuelgoto, line 1050: Linkify [=unapproved=] to point to the definition in the state machine?

spec/index.bs
```diff
@@ -1237,21 +1194,115 @@ dictionary FederatedTokens {
 </dl>
 
 When this method is invoked, the user agent MUST execute the following algorithm:
+  1. Let |promise| be a new {{Promise}}.
+  1. If {{StateMachine/Account State}} is [=unapproved=] and
+      {{StateMachine/Prompt State}} is [=disabled=] then <a for=Promise>Reject</a>
```
samuelgoto, line 1199: "disable" doesn't seem quite right ... should we call this "silent" and "explicit" instead?

spec/index.bs
```diff
@@ -1237,21 +1194,115 @@ dictionary FederatedTokens {
...
 
+To <dfn>explicit sign-in</dfn> the user:
+    1. Gather explicit confirmation that the user wants to sign-in to
+        [=Relying Party=] with their [=Identity Provider=]'s |account|
+    1. if the user does not provide consent then create a {{Promise}},
+         <a for=Promise>reject</a> it and return it.
+    1. Run the [=sign-in=] algorithm.
+
+To <dfn>create tokens</dfn>:
+    1. Assert {{StateMachine/Account State}} is [=approved=]
```
samuelgoto, line 1257: On Account State being "approved" and "unapproved" also doesn't seem right here ... should this be "registered" and "unregistered"?

On Session State, should we call this "loggedin" and "loggedout" instead of "active" or "inactive"?

samuelgoto (commented):
Apologies for the delay, lots of moving parts the last few weeks :( Will try to be more responsive here going forward.

Just a small round of clarification questions on the intent of the state machine and the specific attributes. 

This PR significantly improves the algorithms, so I'm really looking forward to merging it.

spec/index.bs
npm1: Done

spec/index.bs
npm1: Done

spec/index.bs
npm1: Done

spec/index.bs
npm1: Done

spec/index.bs
npm1: Done

spec/index.bs
npm1: I think Account State is fine, as it is about whether the account can be used in the RP or not... I changed Session State.

npm1 (commented):
Addressing comments in https://github.com/fedidcg/FedCM/pull/256

PR #220, opened 2022-03-05
Make certain fields optional

yi-gu:
1. `revocation_endpoint` in the manifest
2. `given_name` in `accounts_endpoint`


yi-gu:
@dj2 @cbiesinger FYI

PR #221, opened 2022-03-07
Rename explainer to explorations

dj2:
The current explainer folder holds an exploration into the design space
that FedCM is considering. It takes about a lot of possible things that
could be done and ways to do that.
    
This is different from the explainer that will need to be produced which
explains the FedCM API specifically.
    
This PR moves the `explainer` folder to an `explorations` folder to make
room for the future explainer.


samuelgoto:
I don't know if "explorations" is the right term, nor if the separation is the right formation.

There is a massive amount of things in this, that you are not calling an explainer, that is pertinent for the short term design.

I can take almost any markdown file, but a few that stand out upon quick introspection in this PR:

- activation.md is highly influential and not forward looking
- prior.md is highly influential and not forward looking
- alternatives.md the same

So, I think "exploration" seems to be a "ignore all of these things here in the corner of the repo" rather than a set of rich documents that have informed the design.

I have an intuition that we are better off just changing the README.md of this directory to point to a summarized / more up to date version of the README.md, but keep all of its content as "part of" the explainer.

npm1:
When I went over this it seemed to me like this set of documents contains mainly two different things:

- Explainer
- Detailed design with forward looking ideas

We need a clear, concise explainer. We also want to keep the design discussion and future work for interested folks, but that group is going to be much smaller than the one just interested in the explainer. Unfortunately they are very intertwined right now.

dj2:
I think the separation makes things a lot clearer. Having an explainer folder with a bunch of stuff unrelated to the explainer is confusing. This provides the context for folks that want it, it's still available in the repo and accessibly from the README file.

I don't think it says to ignore them anywhere? This is where we documented our explorations, there is lots of useful stuff in there.

samuelgoto:
LGTM++

PR #226, opened 2022-03-09
Minor suggestions for explainer

npm1:
This PR fixes some typos and removes the "Example". The explainer has too much code as it is and it seems sufficient to have the code snippets for each of the key scenarios.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -121,18 +121,21 @@ async function login() {
   // state and store any needed account information.
   return await credential.login({ nonce: "456" });
 }
+
 async function logout() {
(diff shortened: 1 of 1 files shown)


npm1:
It is literally repeated across the key scenarios sections. So we should either remove the Example code or remove the redundant snippets in the relevant key scenario sections.

dj2:
The redundancy is in context thought. The example is a broad overview and then each of the key scenarios section goes over it in a bit more detail.

samuelgoto:
Minor editorial suggestions LGTM. Squashing and merging.

PR #228, opened 2022-03-10
Update feature detection in HOWTO

yi-gu:
We should use `login` for feature detection with the new API shape.

### DIFF
--- a/explorations/HOWTO.md
+++ b/explorations/HOWTO.md
@@ -57,9 +57,9 @@ snippet:
 ```js
 // Feature detection: Since `FederatedCredential` is already available in Chrome
 // for the old Credential Management API, additional
-// `FederatedCredential.revoke` check is required.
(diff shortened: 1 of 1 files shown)


npm1:
LGTM

explorations/HOWTO.md
```diff
@@ -57,9 +57,9 @@ snippet:
 ```js
 // Feature detection: Since `FederatedCredential` is already available in Chrome
 // for the old Credential Management API, additional
-// `FederatedCredential.revoke` check is required.
+// `FederatedCredential.prototype.login` check is required.
 function isFedCMEnabled() {
-  return !(!window.FederatedCredential || !FederatedCredential.revoke);
+  return !(!window.FederatedCredential || !FederatedCredential.prototype.login);
```
cbiesinger, line 62: uh, why not:
  return window.FederatedCredential && FederatedCredential.prototype.login;

PR #232, opened 2022-03-14
Add a hint to the revoke call.

### LINKED ISSUE #229: `revoke()` should take a `hint` that is independent from the `accountId`
The revocation endpoint takes "hints" that are different than the "account_id" that it passed to FedCM during the "accounts_endpoint". So, I think we should expose some form of parameter in the revoke method to allow that to be passed to the HTTP endpoint.

Here is a draft of what I was imagining, but open to any other formulation that allows this to be done:

```js
const cred = await navigator.credentials.get({
  federated: {
    providers: [{
      url: "idp.example",
      clientId: "123",
      // no hint needed here, see below
    }]
  }
});
cred.revoke({
  hint: "user@google.com"
})
```

cbiesinger:
Fixes #229

***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/232.html" title="Last updated on Mar 14, 2022, 7:25 PM UTC (22a4064)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/232/b3ae0c8...cbiesinger:22a4064.html" title="Last updated on Mar 14, 2022, 7:25 PM UTC (22a4064)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -291,7 +291,7 @@ could be implemented.
   async function revoke() {
     const credential = getFederatedCredential();
     // This never prompts, rejects the promise in case !credential[“consented”]
-    return credential.revoke();
+    return credential.revoke(credential.id);
   }
 
   async function getFederatedCredential() {
@@ -372,7 +372,7 @@ could be implemented.
 
   async function revoke() {
     const credential = await getLoggedInCredential();
-    return credential.revoke();
+    return credential.revoke(credential.id);
   }
 
   async function getLoggedInCredential() {
@@ -863,10 +863,10 @@ The request is made:
 It will also contain the following parameters in the request body `application/x-www-form-urlencoded`:
 
 <dl dfn-type="argument" dfn-for="revocation_endpoint_request">
-    :   <dfn>account_id</dfn>
-    :: The account identifier for the user (e.g. email).
     :   <dfn>client_id</dfn>
     :: The [=RP=]'s client id
+    :   <dfn>hint</dfn>
+    :: The hint provided to the JS call.
 </dl>
 
 For example:
@@ -879,7 +879,7 @@ Referer: https://rp.example/
 Content-Type: application/x-www-form-urlencoded
 Cookie: 0x23223
 Sec-FedCM-CSRF: ?1
-account_id=123&client_id=client1234
+client_id=client1234&hint=user@idp.example
 ```
 </div>
 
@@ -1265,7 +1265,7 @@ Upon return to the [=Relying Party=], the user goes through a
 
 <div class=example>
 ```js
-credential.revoke();
+credential.revoke('user@idp.example');
 ```
 </div>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -867,8 +867,12 @@ It will also contain the following parameters in the request body `application/x
     :: The account identifier for the user (e.g. email).
     :   <dfn>client_id</dfn>
     :: The [=RP=]'s client id
+    :   <dfn>hint</dfn>
+    :: The hint provided to the JS call.
```
dj2, line 871: Should we remove `account_id` and just have `hint`? When would I use one over the other?

spec/index.bs
```diff
@@ -1339,7 +1350,7 @@ path: img/mock31.svg
 
 <div class=example>
 ```js
-await FederatedCredential.logoutRPs([{
+await FederatedCredential.logoutRps([{
```
dj2, line 1353: This seems unrelated to this PR.

spec/index.bs
cbiesinger: Yeah, probably better to remove it until we know we need it (I was thinking some IDPs may just use their account IDs instead of a different hint, but they can just pass the id as the argument)

spec/index.bs

spec/index.bs
cbiesinger: To be clear, I changed it to just hint and made that argument required.

dj2 (approved):

spec/index.bs
```diff
@@ -1275,10 +1275,13 @@ method:
 <xmp class=idl>
 [Exposed=Window, SecureContext]
 partial interface FederatedCredential {
-  Promise<undefined> revoke();
+  Promise<undefined> revoke(USVString hint);
```
samuelgoto, line 1278: Should this take a dictionary instead of a string? Don't we anticipate revoke to need more parameters as we go along?

spec/index.bs
```diff
@@ -291,7 +291,7 @@ could be implemented.
   async function revoke() {
     const credential = getFederatedCredential();
     // This never prompts, rejects the promise in case !credential[“consented”]
-    return credential.revoke();
+    return credential.revoke(credential.id);
```
samuelgoto, line 294: isn't `credential.id` precisely the account identifier which isn't useful? that is, isn't the case that, in practice, this isn't going to be thing to be pass? wouldn't a better example be:

```js
return credential.revoke(email);
```

samuelgoto, line 294: isn't `credential.id` precisely the account identifier which isn't useful? that is, isn't the case that, in practice, this isn't going to be thing to be pass? wouldn't a better example be:

```js
return credential.revoke(email);
```

cbiesinger: Well, I figured it would work for some IDPs, and in another example I did use an email address

spec/index.bs
cbiesinger: To summarize the discussion elsewhere: If we later on find a need for another parameter, we can easily change this function to take string|dictionary

PR #233, opened 2022-03-16
Reference Permissions Policy for Cross-Origin iframe Support

johannhof:
I think there's a misconception in the explainer that PP would only be useful for per-page control. It actually supports both iframe and header controls. Assuming that the header feature is not harmful to FedCM, PP is the recommended way of adding new `allow` features, makes it easier for implementers to support and more versatile for web developers to work with.

### DIFF
--- a/explainer.md
+++ b/explainer.md
@@ -163,7 +163,7 @@ async function getFederatedCredential() {
   * [IDP initiated RP login](#idp-initiated-rp-login)
   * [Token creation](#token-creation)
   * [Token refresh](#token-refresh)
-  * [Iframe support](#iframe-support)
+  * [Cross-Origin iframe Support](#cross-origin-iframe-support)
   * [IDP initiated front-channel logout](#idp-initiated-front-channel-logout)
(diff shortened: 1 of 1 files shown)


johannhof:
Oh, and, assuming this change is accepted I'm happy to update the spec as well. Maybe I should have filed an issue for this instead, apologies :)

npm1 (approved):
I actually was going to ask about this (I thought the right way would be Document Policy, but this seems correct if PP actually supports iframes). I agree with this PR :)

dj2 (approved):
Ah, interesting. So the Policy turns on FedCM for a given page by setting it to 'self'. That doesn't inherit down to iframes. The iframe is then marked up as "allow=fedcm" which enables the policy in that iframe.  That makes sense, and this lgtm.

samuelgoto:
Looks great to me, thanks for the clarification! Accepting and merging (and yes, would love to accept a PR for the spec too if you can get to it!!)!!

PR #236, opened 2022-03-24
Use Permissions Policy instead of sameOriginWithAncestors

johannhof:
See rationale here:
https://github.com/fedidcg/FedCM/pull/233#issue-1171293908

***
<a href="https://pr-preview.s3.amazonaws.com/johannhof/FedCM/pull/236.html" title="Last updated on Aug 16, 2022, 8:03 PM UTC (e2e229f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/236/f702fa1...johannhof:e2e229f.html" title="Last updated on Aug 16, 2022, 8:03 PM UTC (e2e229f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1363,13 +1363,13 @@ Note: go over how we are planning to deal with backwards compatibility.
 # Permissions Policy Integration # {#permissions-policy-integration}
 <!-- ============================================================ -->
 
-FedCM defines a [=policy-controlled feature=] identified by the string `"fedcm"`.
+FedCM defines a [=policy-controlled feature=] identified by the string `"identity-credential-get"`.
 Its [=default allowlist=] is `"self"`.
 
 A {{Document}}’s [=Document/permissions policy=] determines whether any content
 in that document is allowed to obtain a credential object using the [[#browser-api|Browser API]].
 Attempting to invoke <code><a idl for="CredentialsContainer" lt="get()">navigator.credentials.get({identity:..., ...})</a></code>
-in documents that are not [=allowed to use=] the `"fedcm"` feature will result
+in documents that are not [=allowed to use=] the `"identity-credential-get"` feature will result
 in [=a promise rejected with=] a "{{NotAllowedError}}" {{DOMException}}.
 
 This restriction can be controlled using the mechanisms described in [[PERMISSIONS-POLICY]].
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1065,6 +1065,11 @@ This [=internal method=] accepts three arguments:
     ::  This argument is a Boolean value which is [TRUE] if and only if the
          caller's [=environment settings object=] is
          [=same-origin with its ancestors=]. It is [FALSE] if caller is cross-origin.
+
+        Note: Invocation of this [=internal method=] indicates that it was allowed by
+        [=permissions policy=], which is evaluated at the [[!CREDENTIAL-MANAGEMENT-1]] level.
+        See [[#permissions-policy-integration]]. As such, |sameOriginWithAncestors| is unused.
+
 </dl>
 
 NOTE: The {{CredentialRequestOptions/mediation}} flag is currently not used.
@@ -1077,11 +1082,6 @@ requests.
     <dfn for="IdentityCredential" method>\[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)</dfn>\
     algorithm is invoked, the user agent MUST execute the following steps:
 
-      1.  If <var ignore>sameOriginWithAncestors</var> is `false`, throw a
-          "{{NotAllowedError}}" {{DOMException}}.
-
-          Note: This restriction aims to address the concern raised
-          in [[Security-Origin-Confusion]].
       1.  Assert: |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"] [=map/exists=].
       1.  Assert: |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"] [=list/size=] is 1.
 
@@ -1358,6 +1358,27 @@ steps:
 <!-- ============================================================ -->
 
 Note: go over how we are planning to deal with backwards compatibility.
+
+<!-- ============================================================ -->
+# Permissions Policy Integration # {#permissions-policy-integration}
+<!-- ============================================================ -->
+
+FedCM defines a [=policy-controlled feature=] identified by the string `"fedcm"`.
+Its [=default allowlist=] is `"self"`.
+
+A {{Document}}’s [=Document/permissions policy=] determines whether any content
+in that document is allowed to obtain a credential object using the [[#browser-api|Browser API]].
+Attempting to invoke <code><a idl for="CredentialsContainer" lt="get()">navigator.credentials.get({identity:..., ...})</a></code>
+in documents that are not [=allowed to use=] the `"fedcm"` feature will result
+in [=a promise rejected with=] a "{{NotAllowedError}}" {{DOMException}}.
+
+This restriction can be controlled using the mechanisms described in [[PERMISSIONS-POLICY]].
+
+Note: Algorithms specified in [[!CREDENTIAL-MANAGEMENT-1]] perform the actual
+permissions policy evaluation. This is because such policy evaluation needs to
+occur when there is access to the [=current settings object=]. The [=internal method=]s
+modified by this specification do not have such access since they are invoked [=in parallel=]
+by {{CredentialsContainer}}'s <a abstract-op>Request a `Credential`</a> abstract operation.
 
 <!-- ============================================================ -->
 # Security # {#security}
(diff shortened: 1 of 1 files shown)


johannhof:
(I'm going to join fedidcg to make the IPR pass)

dj2:
We inherit the `[[DiscoverFromExternalSource]]` from Credential Management Level 1 (https://www.w3.org/TR/credential-management-1/#algorithm-discover-creds). So, this would need to be changed there and then propagated to FedCM.

johannhof:
Ah, right, I missed that this parameter came from the original function. I'll file an issue with credential management about PP vs. sameOriginWithAncestors, but I think for now it's probably acceptable to mention the parameter but ignore it in the algorithm.

Thanks!

johannhof:
@dj2 updated, PTAL :)

nsatragno:
Note that permission policy check itself should be added to the relevant algorithm in the credential management side since [[[Discover From External Source]] is called from "in parallel" steps and cannot access the environment settings object](https://w3c.github.io/webauthn/#sctn-permissions-policy).

johannhof:
Thanks @nsatragno, I think that makes sense. I'll update the PR.

dj2 (approved):

johannhof:
@nsatragno @dj2 

I noticed another potential issue that I wanted to get your thoughts on:

Besides `[[DiscoverFromExternalSource]]`, [Request a Credential](https://w3c.github.io/webappsec-credential-management/#abstract-opdef-request-a-credential) will also invoke "Collect Credentials from the credential store" which calls into  `[[CollectFromCredentialStore]]`. This is already implemented by FederatedCredential and [includes another sameOriginWithAncestors check](https://w3c.github.io/webappsec-credential-management/#collectfromcredentialstore-federatedcredential).

I'm not yet super familiar with Credentials so maybe I'm wrong here, but it looks to me like this algorithm would also be called for FedCM FederatedCredentials and effectively prevent the Permissions Policy check in `[[DiscoverFromExternalSource]]` from happening.

Potential solutions:

- Override `[[CollectFromCredentialStore]]` in FedCM and get rid of the same-origin check
- Use the presence of some indicator (e.g. the providers array) in the Credential Management to skip the check conditionally.
- Move the same-origin check out of the parallel steps and into the main "Request a credential" algorithm, where the PP check already lives.

Any thoughts/preferences?

dj2:
There is also #237 which could move us away from monkey patching FederatedCredential if people agree with that direction. This would mean we'd have to put the definition for the `[[CollectFromCredentialStore]]` in this spec and we could handle it the same as Discover.

johannhof:
Right, that might be good. Either way, it does feel like we could move `[[CollectFromCredentialStore]]` into this spec (overriding FederatedCredential for now) and potentially do #237 as a follow-up.

nsatragno:
(Sorry I somehow missed this)

My recommendation is to move away from monkey patching and instead create a brand-new credential type to avoid this class of issues now and in the future, e.g. if we fix a bug on credman's `FederatedCredential` we might accidentally break FedCM. Making eventual deprecation of credman's `FederatedCredential` easier is a nice bonus.

If it's at all helpful to have the current implementation that overrides `FederatedCredential` in the meantime, re-specifying `[[CollectFromCredentialStore]]` in the FedCM spec sounds good -- it'll be less work to migrate to `FederatedCredential2`.

johannhof:
@dj2 @nsatragno made some updates to reflect our discussion, PTAL. I wanted to avoid overriding `CollectFromCredentialStore` since that would be global (i.e. making the version in credman irrelevant), instead opting to modify the algorithm to carve out a case for federated credentials. This should definitely get cleaned up long-term and as mentioned in #237 I'm starting to think that moving this proposal's `FederatedCredential` implementation into the Credential Management spec could be a way forward.

If you're okay with this new version I'll also submit a PR to credman adding the allowed to use check.

spec/index.bs
```diff
@@ -1026,6 +1026,19 @@ requests.
     ::  Optional |accountId| hint. Maybe used by the UA when displaying the
 </dl>
 
+This specification modifies the {{FederatedCredential}}'s {{FederatedCredential/[[CollectFromCredentialStore]](origin, options, sameOriginWithAncestors)}} method by prepending the following step:
```
npm1, line 1029: Technically this would be the second step, i.e. after the assert, right?

spec/index.bs
```diff
@@ -1026,6 +1026,19 @@ requests.
...
+
+  1.  If |options|["{{CredentialRequestOptions/federated}}"]["{{FederatedCredentialRequestOptions/providers}}"] [=map/exists=], return the result of <a abstract-op lt="Retrieve a list of credentials">retrieving</a>
+      credentials from the [=credential store=] that match the following filter:
+        1.  The credential is a {{FederatedCredential}}
```
npm1, line 1033: nit: add period for consistency with other bullet points.

spec/index.bs
```diff
@@ -1026,6 +1026,19 @@ requests.
...
+        1.  The credential's {{Credential/[[origin]]}} is the [=same origin=] as |origin|.
+        1.  |options|["{{CredentialRequestOptions/federated}}"]["{{FederatedCredentialRequestOptions/providers}}"]
+            [=list/contains=] the credentials's {{FederatedCredential/provider}}.
+
+        Note: This step skips any checks for |sameOriginWithAncestors|, which is instead done by
+        evaluating whether the document is [=allowed to use=] the `"fedcm"` feature at at the
+        [[!CREDENTIAL-MANAGEMENT-1]] level. See [[#permissions-policy-integration]].
+
 This specification overrides the {{FederatedCredential}}'s
```
npm1, line 1042: Should we later not override this method? Maybe add an Issue

spec/index.bs
```diff
@@ -1375,6 +1388,20 @@ Note: go over how this is implemented.
...
+
+FedCM defines a [=policy-controlled feature=] identified by the string `"fedcm"`.
+Its [=default allowlist=] is `"self"`.
+
+A {{Document}}’s [=Document/permissions policy=] determines whether any content
+in that document is allowed to obtain a credential object using the [[#browser-api-federated-credential|Browser API]]. Attempting to invoke <code><a idl for="CredentialsContainer" lt="get()">navigator.credentials.get({federated:..., ...})</a></code> in documents that are not [=allowed to use=] the `"fedcm"` feature will result in [=a promise rejected with=] a "{{NotAllowedError}}" {{DOMException}}.
+
+This restriction can be controlled using the mechanisms described in [[PERMISSIONS-POLICY]].
+
+Note: Algorithms specified in [[!CREDENTIAL-MANAGEMENT-1]] perform the actual permissions policy evaluation. This is because such policy evaluation needs to occur when there is access to the [=current settings object=]. The [=internal method=]s modified by this specification do not have such access since they are invoked [=in parallel=] by {{CredentialsContainer}}'s <a abstract-op>Request a `Credential`</a> abstract operation [[!CREDENTIAL-MANAGEMENT-1]].
```
npm1, line 1403: The last mention to `[[!CREDENTIAL-MANAGEMENT-1]]` here seems off, should we remove it?

npm1 (approved):
Looks good to me. @samuelgoto PTAL?

spec/index.bs
johannhof: Yeah, that seems removable.

spec/index.bs
johannhof: I wouldn't require us to change this as long as CredMan has a no-op in place of the overriden function, this draft is for incubation anyway and a final decision could be made in #237 as a result of which this will likely change anyway.

spec/index.bs
johannhof: Yeah, I guess technically you're right, I figured we can skip the assert for simplicity as existence of the map is somewhat implied by `exists`, but I can try to reword.

johannhof:
I think I can take another stab at this now that we have IdentityCredential in place.

spec/index.bs
```diff
@@ -1359,6 +1359,27 @@ steps:
 
 Note: go over how we are planning to deal with backwards compatibility.
 
+<!-- ============================================================ -->
+# Permissions Policy Integration # {#permissions-policy-integration}
+<!-- ============================================================ -->
+
+FedCM defines a [=policy-controlled feature=] identified by the string `"fedcm"`.
```
npm1, line 1366: Let's use a different name, not `fedcm`. How about `identity-credential` (or the plural, not sure what is better in this case)?

npm1 (commented):
Looks good modulo naming!

spec/index.bs
johannhof: Hmm so for publickey (which is the closest to this, I guess) we have `publickey-credentials-get`, which might make sense as it describes the capability that we're exposing, so `identity-credential-get`? It sounds a bit funky though.

@clelland do you have any thoughts/advice on naming for PP?

npm1: Ping

samuelgoto: `identity-credential-get` for consistency with the publickey SGTM

samuelgoto:
LGTM with the small naming changes to keep it consistent. Once you resolve them, I'll merge.

spec/index.bs
johannhof: Thanks, let's agree on that, then!

johannhof:
Done! :)

samuelgoto:
Merged!

spec/index.bs
yi-gu: Should we do the plural `identity-credentials-get` to be consistent?

johannhof: Yeah, I agree, it's also more consistent with the API call (credentials.get) thanks for catching that! I can update this later, as I was going to make it linkable as well.

johannhof: @samuelgoto @npm1 assuming no concerns with this but lmk otherwise

npm1: No concern from my end

PR #239, opened 2022-03-30
Clarify that manifest fetches should not send a Referer

### LINKED ISSUE #215: Update spec to clarify no 'referer' sent with fedcm.json request.
The referer is not sent with the fedcm.json spec but is not currently listed in the spec.

cbiesinger:
Fixes #215


dj2 (approved):
lgtm

PR #270, opened 2022-06-04
Specify how the user agent should use the approved_clients list

samuelgoto:
***
<a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/270.html" title="Last updated on Jun 7, 2022, 5:03 AM UTC (858a7a6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/270/244c4c6...858a7a6.html" title="Last updated on Jun 7, 2022, 5:03 AM UTC (858a7a6)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1185,13 +1185,15 @@ To <dfn>fetch the accounts list</dfn>:
         TODO: explain why we must not follow redirects in the security section.
     1. Return the |accounts list|.
 
-To <dfn>sign-up</dfn> the user:
+To <dfn>sign-up</dfn> the user with a given |account|:
     1. Let |metadata| be the result of running the [=fetch the client metadata=]
         algorithm.
-    1. If |metadata|["privacy_policy_url"] is defined then display the
-        |metadata|["privacy_policy_url"] link.
-    1. If |metadata|["terms_of_service_url"] is defined then display the
-        |metadata|["terms_of_service_url"] link.
+    1. If |metadata|["{{client_metadata_endpoint_response/privacy_policy_url}}"] is defined and
+        the provider["{{FederatedIdentityProvider/clientId}}"] is not in the list of |account|["{{accounts_endpoint_response_accounts/approved_clients}}"]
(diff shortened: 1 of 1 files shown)


npm1 (approved):
Looks good. Perhaps we can add some text explaining this in the description in section 7.2

samuelgoto:
> Looks good. Perhaps we can add some text explaining this in the description in section 7.2

Good point! Done!

samuelgoto:
Neat, merging!

spec/index.bs
```diff
@@ -1185,13 +1185,15 @@ To <dfn>fetch the accounts list</dfn>:
...
-To <dfn>sign-up</dfn> the user:
+To <dfn>sign-up</dfn> the user with a given |account|:
     1. Let |metadata| be the result of running the [=fetch the client metadata=]
         algorithm.
-    1. If |metadata|["privacy_policy_url"] is defined then display the
-        |metadata|["privacy_policy_url"] link.
-    1. If |metadata|["terms_of_service_url"] is defined then display the
-        |metadata|["terms_of_service_url"] link.
+    1. If |metadata|["{{client_metadata_endpoint_response/privacy_policy_url}}"] is defined and
+        the provider["{{FederatedIdentityProvider/clientId}}"] is not in the list of |account|["{{accounts_endpoint_response_accounts/approved_clients}}"]
```
cbiesinger, line 1192: As written, this assumes that the IDP did send an approved_clients list, but that is optional. Should probably talk about what happens if that list was not sent?

samuelgoto: Doesn't this already address the case where it is not provided? Meaning, that "if" never executes, so it goes on to do nothing? Is there anything that gets done when this optional field is NOT passed that I forgot to capture?

cbiesinger: The if is about privacy_policy_url, not about approved_clients?

samuelgoto: Ah, good point, will send a PR to resolve that!

PR #273, opened 2022-06-07
Add Security Section

npm1:
The Privacy Section will be rewritten in a followup. This PR adds a section for Security considerations

***
<a href="https://pr-preview.s3.amazonaws.com/npm1/FedCM/pull/273.html" title="Last updated on Jun 14, 2022, 7:53 PM UTC (7b0937f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/273/7da41fa...npm1:7b0937f.html" title="Last updated on Jun 14, 2022, 7:53 PM UTC (7b0937f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1623,16 +1623,16 @@ which ensures that the request was initiated by the user agent, based on the Fed
 actor cannot spam FedCM API calls, so this is sufficient protection for the new [=IDP=] endpoints.
 
 <!-- ============================================================ -->
-## UI faking ## {#ui-faking}
+## Browser Surface Impersonation ## {#browser-surface-impersonation}
 <!-- ============================================================ -->
 
-The FedCM API introduces new user agent UI, and the user agent may choose to show the UI entirely
+The FedCM API introduces new (trusted) user agent UI, and the user agent may choose to show the UI entirely
 on top of the page's contents, if anything because the page was the one responsible for this UI.
-This introduces a potential concern of a malicious site to try to replicate the FedCM UI. This is
-not possible because the FedCM UI requires knowledge of the user accounts within the [=IDP=]. If
+This introduces a potential concern of a malicious site to try to replicate the FedCM UI, gain the user's trust that the user is interacting with a trusted browser surface and gather information from the user that they would only give to the browser rather than the site (e.g. usernames/passwords of **another** site). This is
+not possible because the FedCM UI uses the metadata about the user accounts of the [=IDP=], which the website doesn't have access. If
 this is a malicious site, it would not know the user accounts, otherwise the user is already
 compromised. In addition, the FedCM UI does not prompt the user to provide additional information
-such as username or password. Thus, UI attempting to retrieve sensitive information from the user
+such as username or password. Thus, the website UI (e.g. iframes) attempting to retrieve sensitive information from the user
 would be noticeably different from the FedCM UI. Finally, because the FedCM UI can only be queried
 from the top-level frame (or potentially from an iframe with explicit permission from the top-level
 frame), the priviledged UI surface is only shown when the top-level frame wants it so. A sneaky
@@ -1653,7 +1653,7 @@ to the site using FedCM. The following table has information about the network r
 
 <script type="text/typogram">
 ┌──────────────────────────┬─────────┬───────────┬──────────┐
-│ Endpoint                 │ Cookies │ Client_id │ Referrer │
+│ Endpoint                 │ cookies │ client_id │ referrer │
 ├──────────────────────────┼─────────┼───────────┼──────────┤
 │ fedcm.json               │  no     │  no       │  no      │
 ├──────────────────────────┼─────────┼───────────┼──────────┤
@@ -1679,7 +1679,7 @@ for an [=IDP=] to adopt the FedCM API without introducing security issues on the
     id=], or referrer. Thus, anyone could perform this fetch, and the information contained therein
     is considered public.
 
-* The accounts are fetched with cookies from the [=IDP=]. This in theory is a new power that the
+* The accounts fetch can't be used to track users because it is performed with cookies from the [=IDP=] but, importantly, without the [=client id=] or referrer. This in theory is a new power that the
     [=RP=] gains that it would not have otherwise. Preventing too many of these fetches may be important,
     but [=IDP=]s are already expected to protect against DoS attacks. In addition, the user agent
     should only allow one FedCM flow per page at any given moment, immediately rejecting any
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -81,7 +81,7 @@ dl.domintro dt code {
 </style>
 
 <script src="https://fedidcg.github.io/FedCM/static/underscore-min.js"></script>
-<script src="https://fedidcg.github.io/FedCM/static/raphael.min.js"></script>   
+<script src="https://fedidcg.github.io/FedCM/static/raphael.min.js"></script>
 <script src="https://fedidcg.github.io/FedCM/static/webfont.js"></script>
 <script src="https://fedidcg.github.io/FedCM/static/typogram.js"></script>
 
@@ -458,6 +458,11 @@ provides the main security boundary on the web.
     * [[OIDC-Connect-Core#Authentication]]
     * [[SAML-Glossary]]
 
+: <dfn>client id</dfn>
+:: Each [=IDP=] assigns to each [=RP=] a [=client id=] to uniquely identify the [=RP=]. Note that
+    this ID is dependent on both the [=IDP=] and the [=RP=], but the [=client id=] of an [=RP=] only
+    needs to be unique with respect to any other [=client id=] within the same [=IDP=].
+
 : <dfn>directed identifier</dfn>
 :: A [=user identifier=] that that is unique for each [=site=] the user visits. A
     goal of anti-tracking policy is to promote [=user identifiers=] to become
@@ -590,7 +595,6 @@ The [[#idp-api]] and the [[#rp-api]] defines a set of HTTP APIs that cooperating
 that they can use.
 
 <script type="text/typogram">
-
 ┌───────────┐                  ┌───────────┐                      ┌───────────┐
 │           │                  │           │                      │           │
 │  Relying  │                  │   User    │                      │  Identity │
@@ -627,6 +631,71 @@ The user agent intermediates in such a matter that makes it impractical for the
 API to be used for tracking purposes, while preserving the functionality of
 identity federation.
 
+<script type="text/typogram">
+    Relying                         User                           Identity
+    Party                           Agent                          Provider
+
+      |                               |                               |
+      | "credential.login()"          |                               |
+      |------------------------------>|                               |
+      |                               |                               |
+      |                               | "GET /.well-known/fedcm.json" |
+      |                               |------------------------------>|
+      |                               |<------------------------------| "{"
+      |                               |                               |   "provider_urls: [ ... ]"
+      |                               |                               | "}"
+      |                               |                               |
+      |                               | "GET /fedcm.json"             |
+      |                               |------------------------------>|
+      |                               |<------------------------------| "{"
+      |                               |                               |   "id_token_endpoint: ...,"
+      |                               |                               |   "accounts_endpoint: ...,"
+      |                               |                               |   "client_metadata_endpoint: ...,"
+      |                               |                               |   "revocation_endpoint: ...,"
+      |                               |                               |   "branding: ...,"
+      |                               |                               | "}"
+      |                               |                               |
+      |                               | "GET client_metadata.php"     |
+      |                               | "client_id"                   |
+      |                               |------------------------------>|
+      |                               |<------------------------------| "{"
+      |                               |                               |   "terms_of_service_url: ...,"
+      |                               |                               |   "privacy_policy_url: ...,"
+      |                               |                               | "}"
+      |                               |                               |
+      |                               |                               |
+      |                               | "GET accounts.php"            |
+      |                               | "cookies"                     |
+      |                               |------------------------------>|
+      |                               |<------------------------------| "{"
+      |                               |                               |   "accounts: [ ... ]"
+      |                               |                               | "}"
+      |                               |                               |
+      |                               |                               |
+      |                     +---------+                               |
+      |                     |         |                               |
+      |                     |         |                               |
+      |                     |         | "account chooser"             |
+      |                     |         |                               |
+      |                     |         |                               |
+      |                     +-------->|                               |
+      |                               |                               |
+      |                               |                               |
+      |                               |                               |
+      |                               | "POST token.php"              |
+      |                               | "client_id, cookies, account" |
+      |                               |------------------------------>|
+      |                               |<------------------------------| "{"
+      |                               |                               |   "id_token: ...,"
+      |                               |                               | "}"
+      |                "{ id_token }" |                               |
+      |<------------------------------|                               |
+      |                               |                               |
+      |                               |                               |
+     -+-                             -+-                             -+-
+</script>
+
+
 This document defines the APIs in the following order:
 
 1. The [[#idp-api]]
@@ -658,8 +727,8 @@ The manifest discovery endpoint is fetched:
 
 (a) **without** cookies,
 (b) **with** a special [[#Sec-FedCM-CSRF]] header,
-(c) **without** a [[RFC7231#header.referer|Referer]] header, and
-(c) **without** following [[RFC7231#header.location|HTTP redirects]].
+(c) **without** a [[RFC9110#header.referer|Referer]] header, and
+(c) **without** following [[RFC9110#header.location|HTTP redirects]].
 
 For example:
 
@@ -672,7 +741,9 @@ Sec-FedCM-CSRF: ?1
 ```
 </div>
 
-The file is parsed expecting the following properties:
+The file is parsed expecting [=Manifest=] JSON object.
+
+The <dfn>Manifest</dfn> JSON object has the following properties:
 
 <dl dfn-type="argument" dfn-for="Manifest">
     :   <dfn>accounts_endpoint</dfn> (required)
@@ -752,8 +823,8 @@ The accounts list endpoint provides the list of accounts the user has at the [=I
 The accounts list endpoint is fetched
 (a) **with** [=IDP=] cookies,
 (b) **with** a special [[#Sec-FedCM-CSRF]] header,
-(c) **without** a [[RFC7231#header.referer|Referer]] header, and
-(d) **without** following [[RFC7231#header.location|HTTP redirects]].
+(c) **without** a [[RFC9110#header.referer|Referer]] header, and
+(d) **without** following [[RFC9110#header.location|HTTP redirects]].
 
 For example:
 
@@ -786,7 +857,7 @@ Every <dfn>Account JSON</dfn> is expected to have the following properties:
     :   <dfn>given_name</dfn> (optional)
     ::  The user's given name.
     :   <dfn>approved_clients</dfn> (optional)
-    ::  A list of [=RP=]s (in the form of Client IDs) this account is already registered with. Used in the [=sign-up=] to allow the [=IDP=] to control whether to show the [=Privacy Policy=] and the [=Terms of Service=].
+    ::  A list of [=RP=]s (in the form of [=Client ID=]s) this account is already registered with. Used in the [=sign-up=] to allow the [=IDP=] to control whether to show the [=Privacy Policy=] and the [=Terms of Service=].
 </dl>
 
 For example:
@@ -822,10 +893,10 @@ The client metadata endpoint provides metadata about [=RP=]s.
 The client medata endpoint is fetched
 (a) **without** cookies,
 (b) **with** a special [[#Sec-FedCM-CSRF]] header,
-(c) **with** a [[RFC7231#header.referer|Referer]] header indicating the [=RP=]'s origin
+(c) **with** a [[RFC9110#header.referer|Referer]] header indicating the [=RP=]'s origin
     (as if [[referrer-policy#referrer-policy-strict-origin|Referer-Policy: strict-origin]]
     was in use), and
-(d) **without** following [[RFC7231#header.location|HTTP redirects]].
+(d) **without** following [[RFC9110#header.location|HTTP redirects]].
 
 The user agent also passes the **client_id**.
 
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -1518,7 +1518,66 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+
+* The token and revocation fetches expose the most: they are peformed with cookies, client id, and
+    referrer. The former is gated on the user interacting with the user agent UI, and enables the
+    [=IDP=] to communicate to the [=RP=] the information required to perform a federated
+    signin/signup. It is not possible for the [=RP=] or the [=IDP=] to force this fetch to happen,
+    as the user agent cannot be spoofed or otherwise tricked. The later can only be performed once a
+    token has been created, but does not require user interaction. That said, performing this fetch
+    does not provide any new information to the [=RP=] or the [=IDP=]. Its purpose is only to make
+    bookkeeping easier, as at this point both the [=RP=] and the [=IDP=] have information about the
+    user. The revocation obviously requires collaboration from both parties: a malicious [=RP=]
```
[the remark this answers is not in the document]
samuelgoto: > I think that makes sense generally for parts other than network fetches

Something that is still not connecting to me is that the network fetches are mitigations for Privacy concerns, *not* Security, right? That is, which requests are made with cookies and which aren't, and why, that fits into the Privacy section, not the security section, right?

spec/index.bs
```diff
@@ -1585,7 +1590,110 @@ Note: go over how we are planning to deal with backwards compatibility.
...
 
-Note: go over security.
+This section provides a few of the security considerations for the FedCM API. Note that there is a
+separate section for [[#privacy]] considerations.
+
+<!-- ============================================================ -->
+## Network requests ## {#network-requests}
+<!-- ============================================================ -->
+
+The FedCM API introduces the ability for a site to ask the browser to execute a few different
```
samuelgoto, line 1600: This is a really solid section, which I think we should keep, but I think this is misplaced: we didn't break the network requests in the specific way you are documenting here for **security** reasons, but rather we did it for **privacy** reasons, meaning that this is well tied as a mitigation to one of the **privacy** concerns. 

So, I'd suggest moving this (in this PR, or in another PR, if you want to re-do the privacy section separately).

Does that make sense?

spec/index.bs
```diff
@@ -1585,7 +1590,110 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+│ token_endpoint           │  yes    │  yes      │  yes     │
+├──────────────────────────┼─────────┼───────────┼──────────┤
+│ client_metadata_endpoint │  no     │  yes      │  yes     │
+├──────────────────────────┼─────────┼───────────┼──────────┤
+│ revocation_endpoint      │  yes    │  yes      │  yes     │
+└──────────────────────────┴─────────┴───────────┴──────────┘
+</script>
+
+For fetches that are sent with cookies, we actually consider them equivalent to first-party fetches,
+and hence include
```
samuelgoto, line 1622:
```suggestion
and hence include first-party cookies as if the resource was loaded as a first-party, e.g. regardless of the SameSite value (which is used when a resource loaded as a third-party, not first-party).
```

spec/index.bs
```diff
@@ -1585,7 +1590,110 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+[SameSite](https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#name-the-samesite-attribute-2)=None
+cookies. This makes it easy for an [=IDP=] to adopt the FedCM API without introducing security
+issues on the API, since the [=RP=] cannot inspect the results from the fetches in any way.
+
+* The manifest fetch is safe because it is performed without cookies, [=client id=], or referrer.
```
samuelgoto, line 1627:
```suggestion
* The manifest fetch can't be used to track users because it is performed without cookies, [=client id=], or referrer.
```

spec/index.bs
```diff
@@ -1585,7 +1590,110 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+    Thus, anyone could perform this fetch, and the information contained therein is considered
+    public.
+
+* The accounts are fetched with cookies from the [=IDP=]. This in theory is a new power that the
+    [=RP=] gains that it would not have otherwise. Preventing too many of these fetches may be important,
+    but [=IDP=]s are already expected to protect against DoS attacks. In addition, the user agent
+    should only allow one FedCM flow per page at any given moment, immediately rejecting any
+    attempts to start another one. Since a FedCM flow can only be terminated when the user interacts
+    or after a long timer, the number of fetches performed is not a concern. The [=IDP=] does learn
+    a lot about the user from this fetch, but this is discussed in detail in the [[#privacy]]
```
samuelgoto, line 1637:
```suggestion
    a lot about the website (e.g. in the form of a [Referer](link to referer spec) header or the [=client_id=] value) the user is in from this fetch, and this is discussed in detail in the [[#privacy]]
```

spec/index.bs
```diff
@@ -1585,7 +1590,110 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+    section.
+
+* The client metadata fetch is performed without cookies from the [=IDP=] but with a [=client id=]
```
samuelgoto, line 1640:
```suggestion
* The client metadata fetch can't be used to track users too because it is performed without cookies from the [=IDP=], albeit with a [=client id=]
```

spec/index.bs
```diff
@@ -1585,7 +1590,110 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+    and a referrer. This allows the [=IDP=] to communicate the relevant [=Privacy Policy=] and
+    [=Terms of Service=] to the user agent, in case they need to be displayed. Again, the [=RP=]
```
samuelgoto, line 1642:
```suggestion
    [=Terms of Service=] to the user agent, in case they need to be displayed. Again, besides possible timing attacks described here, the [=RP=]
```

spec/index.bs
```diff
@@ -1585,7 +1590,110 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+    gains nothing from this fetch, and the [=RP=] could already perform this fetch if it wanted to
+    since it involves no cookies from the [=IDP=].
+
+* The token and revocation fetches expose the most: they are peformed with cookies, [=client id=],
```
samuelgoto, line 1646:
```suggestion
* By design, the token and revocation fetches expose the user at the website to the [=IDP=]: they are peformed with cookies, [=client id=],
```

spec/index.bs
```diff
@@ -1585,7 +1590,110 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+    and referrer. The former is gated on the user interacting with the user agent UI, and enables
+    the [=IDP=] to communicate to the [=RP=] the information required to perform a federated
+    signin/signup. It is not possible for the [=RP=] or the [=IDP=] to force this fetch to happen,
```
samuelgoto, line 1649: > It is not possible for the [=RP=] or the [=IDP=] to force this fetch to happen,

Can you expand on this a bit more? It is hard to comprehend what you are trying to convey here.

spec/index.bs
```diff
@@ -1585,7 +1590,110 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+    bookkeeping easier, as at this point both the [=RP=] and the [=IDP=] have information about the
+    user. The revocation obviously requires collaboration from both parties: a malicious [=RP=]
+    cannot sign the user out of their accounts on other [=RP=]s.
+
+<!-- ============================================================ -->
+## Content Security Policy ## {#content-security-policy}
+<!-- ============================================================ -->
+
+The first fetch triggered by the FedCM API is the manifest list, which is public. However, this one
+points to the actual internal manifest, so it is important to perform a [[!CSP]] check on this to
```
samuelgoto, line 1662: > check on this to prevent some sneaky script from introducing FedCM calls into the [RP](https://pr-preview.s3.amazonaws.com/npm1/FedCM/pull/273.html#rp)'s site

I don't think I understand what is the threat vector here. Can you expand this a bit more on what specific problem we are trying to solve here? What's the attack here?

spec/index.bs
```diff
@@ -1585,7 +1590,110 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+The non-same-origin fetches include, for example, the brand icon. The user agent does not perform a
+[[!CSP]] check on these because they are directly specified from the manifest. In addition, the
+rendering of this image is performed by the user agent, and as such this image cannot affect the
+[=RP=] site nor can they be inspected by the [=RP=] in any way.
+
+<!-- ============================================================ -->
+## Sec-FedCM-CSRF ## {#sec-fedcm-csrf}
+<!-- ============================================================ -->
+
+The FedCM API introduces a bunch of endpoints on the [=IDP=], so these need to be protected from XSS
```
samuelgoto, line 1676:
```suggestion
The FedCM API introduces a many non-static endpoints on the [=IDP=], so these need to be protected from XSS
```

spec/index.bs
```diff
@@ -1585,7 +1590,110 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+attacks. In order to do so, the FedCM API introduces the [[#Sec-FedCM-CSRF]] header, a [=forbidden
+header name=]. An [=IDP=] is encouraged to check for this header in the requests it receives, which
```
samuelgoto, line 1678:
```suggestion
header name=] which cannot be set by random websites (so that the [=IDP=] can be confident that the request was originated by the FedCM browser rather than sent by a websites trying to run an XSS). An [=IDP=] is encouraged to check for this header in the requests it receives, which
```

samuelgoto, line 1678:
```suggestion
header name=]. An [=IDP=] must check for this header in the id_token and revocation endpoints, which
```

samuelgoto: This isn't something that they should check for non-credentialed fetches, right (e.g. the manifest)?

spec/index.bs
```diff
@@ -1585,7 +1590,110 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+ensures that the request was initiated by the user agent, based on the FedCM API. A malicious actor
+cannot spam FedCM API calls, so this is sufficient protection for the new [=IDP=] endpoints.
+
+<!-- ============================================================ -->
+## UI faking ## {#ui-faking}
+<!-- ============================================================ -->
+
+The FedCM API introduces new user agent UI, and the user agent may choose to show the UI entirely
+on top of the page's contents, if anything because the page was the one responsible for this UI.
+This introduces a potential concern of a malicious site to try to replicate the FedCM UI. This is
```
samuelgoto, line 1688: Could you write this in terms of a threat vector? what specifically are we trying to prevent here? What would allow a website to replicate the FedCM that would be harmful?

spec/index.bs
npm1: Will reword to `It is not possible for the [=RP=] or the [=IDP=] to force the token fetch to happen without user consent`

spec/index.bs

spec/index.bs

spec/index.bs

spec/index.bs
npm1: Aren't all those questions are answered in this paragraph? Essentially 1) it cannot be replicated 2) because no sensitive information is requested from the user, any harmful such fake would have to be different.

spec/index.bs
```diff
@@ -1585,12 +1590,122 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+This section provides a few of the security considerations for the FedCM API. Note that there is a
+separate section for [[#privacy]] considerations.
+
+<!-- ============================================================ -->
+## Content Security Policy ## {#content-security-policy}
+<!-- ============================================================ -->
+
+The first fetch triggered by the FedCM API is the manifest list, which is public. However, this one
+points to the actual internal manifest, so it is important to perform a [[!CSP]] check on this. This
+prevents a malicious script from introducing a FedCM prompt that is not desired by the [=RP=] or the
```
samuelgoto, line 1602: I'm not sure if I understand the threat model here, but wouldn't it be possible for an attacker to point to an arbitrary IDP that isn't in the CSP list?

Regardless, could you expand more on what specific threat we are trying to resolve here? You say "this prevents a malicious script from introducing a FedCM prompt", but I'm not sure what you mean by this? Is the malicious script running in the RP? In the IDP? Both? Is the attack against the IDP? The RP? What does this have to do with the manifest fetch?

And then, what precisely does the CSP help with?

spec/index.bs
```diff
@@ -1585,12 +1590,122 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+The FedCM API introduces several non-static endpoints on the [=IDP=], so these need to be protected
+from XSS attacks. In order to do so, the FedCM API introduces the [[#Sec-FedCM-CSRF]] header, a
+[=forbidden header name=]. The header cannot be set by random websites, so the [=IDP=] can be
+confident that the request was originated by the FedCM browser rather than sent by a websites trying
+to run an XSS attack. An [=IDP=] must to check for this header in the credentialed requests it receives,
+which ensures that the request was initiated by the user agent, based on the FedCM API. A malicious
+actor cannot spam FedCM API calls, so this is sufficient protection for the new [=IDP=] endpoints.
+
+<!-- ============================================================ -->
+## UI faking ## {#ui-faking}
```
samuelgoto, line 1626:
```suggestion
## Browser Surface Impersonation ## {#browser-surface-impersonation}
```

spec/index.bs
```diff
@@ -1585,12 +1590,122 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+<!-- ============================================================ -->
+
+The FedCM API introduces new user agent UI, and the user agent may choose to show the UI entirely
```
samuelgoto, line 1629:
```suggestion
The FedCM API introduces new (trusted) user agent UI, and the user agent may choose to show the UI entirely
```

spec/index.bs
```diff
@@ -1585,12 +1590,122 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+on top of the page's contents, if anything because the page was the one responsible for this UI.
+This introduces a potential concern of a malicious site to try to replicate the FedCM UI. This is
```
samuelgoto, line 1631:
```suggestion
This introduces a potential concern of a malicious site to try to replicate the FedCM UI, gain the user's trust that the user is interacting with a trusted browser surface and gather information from the user that they would only give to the browser rather than the site (e.g. usernames/passwords of **another** site). This is
```

spec/index.bs
```diff
@@ -1585,12 +1590,122 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+not possible because the FedCM UI requires knowledge of the user accounts within the [=IDP=]. If
```
samuelgoto, line 1632:
```suggestion
not possible because the FedCM UI uses the metadata about the user accounts of the [=IDP=], which the website doesn't have access. If
```

spec/index.bs
```diff
@@ -1585,12 +1590,122 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+this is a malicious site, it would not know the user accounts, otherwise the user is already
+compromised. In addition, the FedCM UI does not prompt the user to provide additional information
+such as username or password. Thus, UI attempting to retrieve sensitive information from the user
```
samuelgoto, line 1635:
```suggestion
such as username or password. Thus, the website UI (e.g. iframes) attempting to retrieve sensitive information from the user
```

spec/index.bs
samuelgoto: LGTM

spec/index.bs
```diff
@@ -1585,12 +1590,122 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+<!-- ============================================================ -->
+
+The FedCM API introduces the ability for a site to ask the browser to execute a few different
+network requests, as shown in the [[#high-level-design]] section. It is important for the browser to
+execute these in such a way that it does not accidentally leak sensitive information about the user
+to the site using FedCM. The following table has information about the network requests performed:
+
+<script type="text/typogram">
+┌──────────────────────────┬─────────┬───────────┬──────────┐
+│ Endpoint                 │ Cookies │ Client_id │ Referrer │
```
samuelgoto, line 1656:
```suggestion
│ Endpoint                 │ cookies │ client_id │ referrer │
```

spec/index.bs
```diff
@@ -1585,12 +1590,122 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+[SameSite](https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#name-the-samesite-attribute-2)
+value (which is used when a resource loaded as a third-party, not first-party). This makes it easy
+for an [=IDP=] to adopt the FedCM API without introducing security issues on the API, since the
+[=RP=] cannot inspect the results from the fetches in any way.
+
+* The manifest fetch can't be used to track users because it is performed without cookies, [=client
+    id=], or referrer. Thus, anyone could perform this fetch, and the information contained therein
+    is considered public.
+
+* The accounts are fetched with cookies from the [=IDP=]. This in theory is a new power that the
```
samuelgoto, line 1682:
```suggestion
* The accounts fetch can't be used to track users because it is performed with cookies from the [=IDP=] but, importantly, without the [=client id=] or referrer. This in theory is a new power that the
```

samuelgoto (commented):
Looks generally good to me. Final round of review, will merge once resolved.

Biggest concern is with the description of the CSP feature, which is unclear to me what purpose it serve (I'm sure there is one, I just can't figure that out by reading this text, so please expand it a bit more to explain what specific problem does it solve).

spec/index.bs
npm1: Sorry for not being clear here. I tried to rephrase this to make it clearer.

npm1:
Ping

spec/index.bs
```diff
@@ -1414,8 +1419,8 @@ The <dfn>fetch the internal manifest</dfn> algorithm accepts a {{FederatedIdenti
     1. Return the result of fetching the |manifest url| and parsing it into a [=Manifest=].
         TODO: specify how the fetching and the parsing works.
         The fetch MUST be done with the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]] header but
-	without the [=Identity Provider=]'s cookies. This request MUST NOT follow
-        [[RFC7231#header.location|HTTP redirects]] and instead abort with an
+	      without the [=Identity Provider=]'s cookies. This request MUST NOT follow
+        [[RFC9110#header.location|HTTP redirects]] and instead abort with an
```
samuelgoto, line 1423: In a follow up PR, could you explain why we disable following redirects in the security (or privacy?) section? If easy in this PR, add a TODO so we don't forget?

spec/index.bs
```diff
@@ -1585,12 +1590,129 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+<!-- ============================================================ -->
+## Content Security Policy ## {#content-security-policy}
+<!-- ============================================================ -->
+
+The first fetches triggered by the FedCM API are the manifest list, which is public, and the
+internal manifest. Imagine a malicious script attempting to execute the FedCM API calls to a
+malicious [=IDP=], one which is not trusted by the [=RP=]. If the call is successful, this would
+introduce browser UI on the [=RP=] with sign in options into a malicious [=IDP=]. This malicious
+[=IDP=] could then attempt to trick the user. The protection against this attack is the [[!CSP]]
+check, which would fail because the manifest would not be an origin allowed by the [=RP=], hence
```
samuelgoto, line 1605:
```suggestion
check, which would fail because the origin of the manifest of the malicious [=IDP=] would not be an origin included in the allowlist specified by the [[!CSP]] of the [=RP=], hence
```

spec/index.bs
```diff
@@ -1585,12 +1590,129 @@ Note: go over how we are planning to deal with backwards compatibility.
...
-Note: go over security.
+This section provides a few of the security considerations for the FedCM API. Note that there is a
+separate section for [[#privacy]] considerations.
+
+<!-- ============================================================ -->
+## Content Security Policy ## {#content-security-policy}
+<!-- ============================================================ -->
+
+The first fetches triggered by the FedCM API are the manifest list, which is public, and the
+internal manifest. Imagine a malicious script attempting to execute the FedCM API calls to a
```
samuelgoto, line 1601:
```suggestion
internal manifest. Imagine a malicious script included by (and running as) the [=RP=] attempting to execute the FedCM API calls to a
```

spec/index.bs
```diff
@@ -1585,12 +1590,129 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+
+The FedCM API introduces new (trusted) user agent UI, and the user agent may choose to show the UI
+entirely on top of the page's contents, if anything because the page was the one responsible for
+this UI. This introduces a potential concern of a malicious site to try to replicate the FedCM UI,
+gain the user's trust that the user is interacting with a trusted browser surface, and gather
+information from the user that they would only give to the browser rather than the site (e.g.
+usernames/passwords of **another** site). This is not possible because the FedCM UI uses the
+metadata about the user accounts of the [=IDP=], which the website doesn't have access. If this is a
+malicious site, it would not know the user accounts, otherwise the user is already compromised. In
+addition, the FedCM UI does not prompt the user to provide additional information such as username
```
samuelgoto, line 1640:
```suggestion
addition, the FedCM UI deliberately constructed to not prompt the user to provide additional information, such as username
```

spec/index.bs
```diff
@@ -1585,12 +1590,129 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+or password. Thus, the website UI (e.g. iframes) attempting to retrieve sensitive information from
```
samuelgoto, line 1641:
```suggestion
or password. Thus, an attacker trying to impersonate the browser using exclusively UI that is accessible to the content area (e.g. iframes) to attempt to retrieve sensitive information from
```

spec/index.bs
```diff
@@ -1585,12 +1590,129 @@ Note: go over how we are planning to deal with backwards compatibility.
...
+├──────────────────────────┼─────────┼───────────┼──────────┤
+│ token_endpoint           │  yes    │  yes      │  yes     │
+├──────────────────────────┼─────────┼───────────┼──────────┤
+│ client_metadata_endpoint │  no     │  yes      │  yes     │
+├──────────────────────────┼─────────┼───────────┼──────────┤
+│ revocation_endpoint      │  yes    │  yes      │  yes     │
+└──────────────────────────┴─────────┴───────────┴──────────┘
+</script>
+
+For fetches that are sent with cookies, we actually consider them equivalent to first-party fetches,
```
samuelgoto, line 1676: Add a TODO here to cover the reason why we don't follow the redirects?

samuelgoto (approved):
LGTM

Will merge once last round of comments are addressed.

Thanks @npm1 !

spec/index.bs
npm1: Added TODO.

samuelgoto:
Thanks! Merging!

PR #275, opened 2022-06-08
Define a Manifest List object and break down the algorithms into smaller parts.

samuelgoto:
Following up a TODO from the last PR.

***
<a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/275.html" title="Last updated on Jun 9, 2022, 11:49 PM UTC (e87f218)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/275/7da41fa...e87f218.html" title="Last updated on Jun 9, 2022, 11:49 PM UTC (e87f218)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -714,10 +714,10 @@ by exposing a series of HTTP endpoints:
 ## Manifest List ## {#idp-api-manifest-list}
 <!-- ============================================================ -->
 
-We use a two-tier manifest list in order to prevent the [=IDP=] to easily determine the [=RP=]
+We use a two-tier [=manifest list=] in order to prevent the [=IDP=] to easily determine the [=RP=]
 that a user is visiting by encoding the information in the manifest path. We solve this issue by
-requiring a manifest list to be on the root of the [=IDP=]. The [=manifest=] itself can be anywhere, but
-it will not be used if the user agent does not find it in the manifest list. This allows the [=IDP=]
+requiring a [=manifest list=] to be on the root of the [=IDP=]. The [=manifest=] itself can be anywhere, but
+it will not be used if the user agent does not find it in the [=manifest list=]. This allows the [=IDP=]
 to keep their actual [=manifest=] on an arbitary path while allowing the user agent to prevent manifest
 manipulation to fingerprint.
 
@@ -727,7 +727,7 @@ The <dfn>Manifest List</dfn> has the following properties:
 
 <dl dfn-type="argument" dfn-for="Manifest List">
     :   <dfn>provider_urls</dfn> (required)
-    ::  An array of URL that points the supported [=manifests=].
+    ::  An [=list=] of URLs that points to supported [=manifests=].
 </dl>
 
 <!-- ============================================================ -->
@@ -1397,8 +1397,7 @@ The <dfn>fetch the manifest</dfn> algorithm accepts a {{FederatedIdentityProvide
 The <dfn>check the manifest list</dfn> accepts a {{FederatedIdentityProvider}} |provider| and returns whether the manifest is included in the manifest list:
     1. Let |response| be the result of the [=fetch the manifest list=] for the |provider|.
     1. Let |manifest list| be the result of running the [=parse the manifest list=] algorithm with the |response|.
-    1. Return true if |manifest list|["provider_urls"][0] [=string/is=] equal to |provider|'s
-        {{FederatedIdentityProvider/url}}, otherwise return false.
+    1. Return true if if any of the entries in the |manifest list| is equal to |provider|'s {{FederatedIdentityProvider/url}}.
 
 The <dfn>fetch the manifest list</dfn> algorithm accepts a {{FederatedIdentityProvider}} |provider| and returns a <a spec=fetch for=/>response</a>:
     1. Let |url| be |provider|'s {{FederatedIdentityProvider/url}}.
@@ -1408,7 +1407,7 @@ The <dfn>fetch the manifest list</dfn> algorithm accepts a {{FederatedIdentityPr
     1. Set |rootUrl|'s [=url/path=] to ".well-known/fedcm.json".
     1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is |rootUrl| and
         [=request/redirect mode=] set to "error". TODO: what other fields need to be set?
-    1. Return the result o[=Fetch=]ing |request| with <a spec=fetch>processResponseConsumeBody</a>.
+    1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the following steps given a <a spec=fetch for=/>response</a> |response|/
 
 The <dfn>parse the manifest list</dfn> algorithm accepts a <a spec=fetch for=/>response</a> and returns a [=Manifest List=] or false if it fails to parse it:
     1. If |response| is a [=network error=] or its [=response/status=] is not an [=ok status=],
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1368,55 +1388,69 @@ To <dfn>create tokens</dfn>:
...
+    1. If |response| is a [=network error=] or its [=response/status=] is not an [=ok status=],
+        return false.
+    1. If |response|'s [=Content-Type Metadata=] is not a [=JSON MIME Type=], return false.
+    1. Let |json| be the result of [=parse JSON bytes to an Infra value=] passing |response|'s
+        [=response/body=].
+    1. If |json| is a parsing exception, or if |json| is not an [=ordered map=], return false.
+    1. If |json|["accounts_endpoint"] does not exist, or if it is not a [=string=], return false.
+    1. If |json|["client_metadata_endpoint"] does not exist, or if it is not a [=string=], return false.
+    1. If |json|["id_token_endpoint"] does not exist, or if it is not a [=string=], return false.
+    1. Let |result| be a new [=Manifest=] created from |json|.
```
samuelgoto, line 1451: I wasn't sure if this conversion was cool to do, but it seemed like too verbose to copy the fields from the |json| to the |result| manually. WDYT?

spec/index.bs
```diff
@@ -710,6 +710,26 @@ by exposing a series of HTTP endpoints:
...
+to keep their actual [=manifest=] on an arbitary path while allowing the user agent to prevent manifest
+manipulation to fingerprint.
+
+Note: See https://github.com/fedidcg/FedCM/issues/230 for more details on this attack.
+
+The <dfn>Manifest List</dfn> has the following properties:
+
+<dl dfn-type="argument" dfn-for="Manifest List">
+    :   <dfn>provider_urls</dfn> (required)
+    ::  An array of URL that points the supported [=manifests=].
```
cbiesinger, line 730: array -> list?

spec/index.bs
```diff
@@ -1368,55 +1388,69 @@ To <dfn>create tokens</dfn>:
...
-        1. If |json| is a parsing exception, or if |json| is not an [=ordered map=], return false.
-        1. If |json|["provider_urls"] does not exist, or if it is not a [=list=], return false.
-        1. If the [=list/size=] of |json|["provider_urls"] is greater than 1, return false. TODO:
-            determine the right value, probably greater than 1.
-        1. If |json|["provider_urls"][0] is not a [=string=], return false.
-        1. Return true if |json|["provider_urls"][0] [=string/is=] equal to |provider|'s
-            {{FederatedIdentityProvider/url}}, otherwise return false.
-
-The <dfn>fetch the internal manifest</dfn> algorithm accepts a {{FederatedIdentityProvider}} |provider| and returns a [=Manifest=]:
+    1. Return the result o[=Fetch=]ing |request| with <a spec=fetch>processResponseConsumeBody</a>.
```
cbiesinger, line 1411: o -> of?

spec/index.bs
```diff
@@ -1368,55 +1388,69 @@ To <dfn>create tokens</dfn>:
...
-that a user is visiting by encoding the information in the manifest path. We solve this issue by
-requiring a manifest list to be on the root of the [=IDP=]. The manifest itself can be anywhere, but
-it will not be used if the user agent does not find it in the manifest list. This allows the [=IDP=]
-to keep their actual manifest on an arbitary path while allowing the user agent to prevent manifest
-manipulation to fingerprint. See https://github.com/fedidcg/FedCM/issues/230 for more details on
-this attack.
+The <dfn>check the manifest list</dfn> accepts a {{FederatedIdentityProvider}} |provider| and returns whether the manifest is included in the manifest list:
+    1. Let |response| be the result of the [=fetch the manifest list=] for the |provider|.
+    1. Let |manifest list| be the result of running the [=parse the manifest list=] algorithm with the |response|.
+    1. Return true if |manifest list|["provider_urls"][0] [=string/is=] equal to |provider|'s
```
cbiesinger, line 1400: Perhaps write this more generically as "if any of the entries in the manifest list is equal to provider's URL"? I suggest this for two reasons: at this point in the spec, the reader has not yet seen that the list is limited to one entry and thus the [0] will be confusing, and if we decide to use a higher limit then that will avoid having to update this section.

spec/index.bs
```diff
@@ -1368,55 +1388,69 @@ To <dfn>create tokens</dfn>:
...
     1. Run a [[!CSP]] check with a [[CSP#directive-connect-src|connect-src]]
         directive on the URL passed as |provider|'s {{FederatedIdentityProvider/url}}
         and return if the check fails.
     1. Let the |manifest url| be the file "fedcm.json" relative to the directory
         prefix passed as |provider|'s {{FederatedIdentityProvider/url}}.
-    1. Return the result of fetching the |manifest url| and parsing it into a [=Manifest=].
-        TODO: specify how the fetching and the parsing works.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is |rootUrl| and
+        [=request/redirect mode=] set to "error". TODO: what other fields need to be set?
+    1. Let |response| be the result o[=Fetch=]ing |request| with <a spec=fetch>processResponseConsumeBody</a>.
```
cbiesinger, line 1434: o -> of?

spec/index.bs
```diff
@@ -710,6 +710,26 @@ by exposing a series of HTTP endpoints:
 1. An [[#idp-api-id-token-endpoint]] endpoint
 1. A [[#idp-api-revocation-endpoint]] endpoint
 
+<!-- ============================================================ -->
+## Manifest List ## {#idp-api-manifest-list}
+<!-- ============================================================ -->
+
+We use a two-tier manifest list in order to prevent the [=IDP=] to easily determine the [=RP=]
```
npm1, line 717:
```suggestion
We use a two-tier manifest list in order to prevent the [=IDP=] from easily determining the [=RP=]
```

spec/index.bs
```diff
@@ -710,6 +710,26 @@ by exposing a series of HTTP endpoints:
...
+that a user is visiting by encoding the information in the manifest path. We solve this issue by
+requiring a manifest list to be on the root of the [=IDP=]. The [=manifest=] itself can be anywhere, but
```
npm1, line 719: link manifest list? Here and the next line

spec/index.bs
```diff
@@ -710,6 +710,26 @@ by exposing a series of HTTP endpoints:
...
+it will not be used if the user agent does not find it in the manifest list. This allows the [=IDP=]
+to keep their actual [=manifest=] on an arbitary path while allowing the user agent to prevent manifest
+manipulation to fingerprint.
+
+Note: See https://github.com/fedidcg/FedCM/issues/230 for more details on this attack.
+
+The <dfn>Manifest List</dfn> has the following properties:
```
npm1, line 726: This definition is unclear. What is the type of this object? In fact, we don't even need a new definition. The JSON parsing algorithm returns an object that we can directly use. What is the purpose here?

spec/index.bs
npm1, line 1411: The changes in this step are incorrect. Fetch returns a fetch controller. The response must be inspected from `processResponseConsumeBody`.

spec/index.bs
```diff
@@ -1368,55 +1388,69 @@ To <dfn>create tokens</dfn>:
...
 
-NOTE: We use a two-tier manifest list in order to prevent the [=IDP=] to easily determine the [=RP=]
-that a user is visiting by encoding the information in the manifest path. We solve this issue by
-requiring a manifest list to be on the root of the [=IDP=]. The manifest itself can be anywhere, but
-it will not be used if the user agent does not find it in the manifest list. This allows the [=IDP=]
-to keep their actual manifest on an arbitary path while allowing the user agent to prevent manifest
-manipulation to fingerprint. See https://github.com/fedidcg/FedCM/issues/230 for more details on
-this attack.
+The <dfn>check the manifest list</dfn> accepts a {{FederatedIdentityProvider}} |provider| and returns whether the manifest is included in the manifest list:
+    1. Let |response| be the result of the [=fetch the manifest list=] for the |provider|.
```
npm1, line 1398: Fetch is truly async, so I think having this as step 1 of 3 is not correct.

spec/index.bs
npm1, line 1434: Also same problem as above: incorrect way to obtain the `response`.

spec/index.bs
```diff
@@ -1368,55 +1388,69 @@ To <dfn>create tokens</dfn>:
...
+    1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is |rootUrl| and
+        [=request/redirect mode=] set to "error". TODO: what other fields need to be set?
+    1. Let |response| be the result o[=Fetch=]ing |request| with <a spec=fetch>processResponseConsumeBody</a>.
         The fetch MUST be done with the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]] header but
 	without the [=Identity Provider=]'s cookies. This request MUST NOT follow
         [[RFC7231#header.location|HTTP redirects]] and instead abort with an
         error if there are any.
+    1. Return the result of the [=parse the manifest=] algorithm with the |response|.
+
+The <dfn>parse the manifest</dfn> algorithm accepts a <a spec=fetch for=/>response</a> and returns a [=Manifest=] or false if it fails to parse it:
```
npm1, line 1441: return null instead of false? Makes more sense given this returns an object

spec/index.bs
npm1: Why not just return |json|? Manifest and Manifest List are really just ordered maps. No need to redefine objects here (all the needed checks were also performed in the parsing steps above)

samuelgoto: Wouldn't it be constructive for the return object to be typed (e.g. A Manifest object, which has well defined key-value semantics) as opposed to JSON?

spec/index.bs
samuelgoto: ah, yes, thanks, done.

spec/index.bs
samuelgoto: SGTM. Done.

spec/index.bs
samuelgoto: Done.

spec/index.bs
samuelgoto: Done.

spec/index.bs
samuelgoto: Sure. Done.

spec/index.bs
samuelgoto: Done.

spec/index.bs
samuelgoto: We do have a <dfn> for the Manifest, and I'm partially just replicating that because it seemed constructive to return a structured well defined object, rather than JSON. But open to other ways to go about this.

WDYT?

spec/index.bs
samuelgoto: How should we go about this? Is there some version of async/await for specs that I should be using?

npm1: Fetch has a parameter called `processResponseConsumeBody`, which is an algorithm. So steps to be taken once the `response` object is ready would be called in the algorithm passed as parameter to fetch. Per your other question, in specs we use tasks for async work. There's also 'in parallel' for stuff where you execute a bunch of things at the same time (i.e. no particular order).

spec/index.bs
npm1: IIUC `parse JSON bytes to an Infra value` returns an ordered map, which is a well defined infra value. That said, I'm not totally opposed to having the new defn, just think it may be unnecessary. Let's hear from Johann (tagged him in the above comment).

npm1:

PR #277, opened 2022-06-15
Specify what happens when the approved_clients list is undefined

samuelgoto:


spec/index.bs
```diff
@@ -1317,11 +1317,12 @@ To <dfn>fetch the accounts list</dfn>:
 To <dfn>sign-up</dfn> the user with a given |account|:
     1. Let |metadata| be the result of running the [=fetch the client metadata=]
         algorithm.
+    1. Let |approved_clients| be |account|["{{accounts_endpoint_response_accounts/approved_clients}}"] if that's defined or an empty list otherwise.
```
cbiesinger, line 1320: Should we specify that the user agent should keep track locally of whether the user has previously logged in? Either way lgtm

cbiesinger (approved):

spec/index.bs
samuelgoto: Have we implemented that? Do we actually keep track and use that in conjunction with the approved_clients array?

cbiesinger: Yes. If we do not get an approved_clients field we use locally stored data instead. https://source.chromium.org/chromium/chromium/src/+/main:content/browser/webid/federated_auth_request_impl.cc;l=971

samuelgoto: Yeah, that makes sense. I added some text to note that (probably a bit confusing, but still I think it works).

cbiesinger (approved):

Ahlamawada (approved):

spec/index.bs
asr-enid: As per my understanding how approved_clients works this will likely lead to incorrect user interactions / state mismatch with the IDP.

1. User signs-up with an RP through a FedCM Interaction -> (will be added to the approved client by the IDP)
2. User revokes that status out-of-band (could be on another device, completely different session) - revoking means user revokes his approval to provide claims to the RP
3. User visits RPs, IDP provides empty list, browser uses cached status -> User is not shown sign-up dialog (which is a problem for the IDP)

Mixing local and IDP side state will yield incorrect flows (specifically in cases where users revoke approvals)

yi-gu: Right I think we could make it clear in the spec that:
1. if IDP does not provide `approved_clients`, the local state in UA is the source of truth.
2. if IDP provides an empty `approved_clients` list OR the list is not empty but the said client is not on the list, UA would treat the user as a new user and show the disclosure text accordingly.

asr-enid: That makes sense to me. Not sure how the use-case for 1 looks like, but thats a separate question

npm1:

PR #278, opened 2022-06-15
Avoid hard-coding the fedcm.json manifest and allow the manifest to b…

samuelgoto:
…e passed directly

***
<a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/278.html" title="Last updated on Jun 16, 2022, 1:07 PM UTC (09cccff)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/278/1a1c446...09cccff.html" title="Last updated on Jun 16, 2022, 1:07 PM UTC (09cccff)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -720,7 +720,7 @@ by exposing a series of HTTP endpoints:
 <!-- ============================================================ -->
 
 The manifest discovery endpoint is an endpoint
-and serves as a discovery device to other endpoints provided by the
+which serves as a discovery device to other endpoints provided by the
 [=IDP=].
 
 The manifest discovery endpoint is fetched:
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -719,7 +719,7 @@ by exposing a series of HTTP endpoints:
 ## Manifest ## {#idp-api-manifest}
 <!-- ============================================================ -->
 
-The manifest discovery endpoint is an endpoint located at the [=IDP=]'s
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -720,8 +720,8 @@ by exposing a series of HTTP endpoints:
 <!-- ============================================================ -->
 
 The manifest discovery endpoint is an endpoint located at the [=IDP=]'s
```
npm1, line 722:
```suggestion
The manifest discovery endpoint is an endpoint
```

spec/index.bs
```diff
@@ -720,8 +720,8 @@ by exposing a series of HTTP endpoints:
...
-`fedcm.json` file and serves as a discovery device to other endpoints provided
-by the [=IDP=].
+and serves as a discovery device to other endpoints provided by the
```
npm1, line 723:
```suggestion
which serves as a discovery device to other endpoints provided by the
```

npm1 (approved):
LGTM with a sentence reword

matthewp:
It's great that this avoids the hardcoding, but I think using `manifest.json` might not be the best choice. Manifest is such an overloaded term, I see manifest.json and immediately [think of this](https://developer.mozilla.org/en-US/docs/Web/Manifest). Can you not use `fedcm.json` in the examples or `fedcm-manifest.json` or something else less generic?

PR #280, opened 2022-06-15
Create a new IdentityCredential interface to avoid breaking changes to the existing FederatedCredential

samuelgoto:
Subclassing the existing FederatedCredential forces us to make backwards compatible changes to its current deployment, which is proving to be harder than we anticipated. It was worth starting from there, but it has become clear to us that extending isn't worth the cost of making backwards incompatible changes (or a deprecation).

In this PR, we introduce a new interface type, IdentityCredential, which allows us to decouple the FedCM API from the current deployment of FederatedCredential.

***
<a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/280.html" title="Last updated on Jun 16, 2022, 10:49 PM UTC (b7c1e96)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/280/b6782b2...b7c1e96.html" title="Last updated on Jun 16, 2022, 10:49 PM UTC (b7c1e96)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -287,7 +287,7 @@ entries provided by the [=Identity Provider=].
 # Examples # {#examples}
 <!-- ============================================================ -->
 
-This specification extends the {{FederatedCredential}} type and internal
+This specification defines a new {{IdentityCredential}} type and internal
 algorithms to allow the exchange of identity between [=IDP=]s and [=RP=]s.
 When it succeeds, it returns to the [=RP=] a signed [=id token=] which the
 [=RP=] can use to authenticate the user.
@@ -311,7 +311,7 @@ could be implemented.
 
   <script>
   async function login() {
-    const credential = getFederatedCredential();
+    const credential = getCredential();
 
     // This will prompt when !credential["registered"], but
     // it won’t when credential["registered"].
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1371,7 +1390,7 @@ To <dfn>create tokens</dfn>:
         * {{FederatedTokens/idToken}} as |tokens|["id_token"]
     1. <a for=Promise>Resolve</a> |promise| with |tokens| and return |promise|.
 
-The <dfn>fetch the manifest</dfn> algorithm accepts a {{FederatedIdentityProvider}} |provider| and returns a [=Manifest=] object:
+The <dfn>fetch the manifest</dfn> algorithm accepts a {{IdentityProvider}} |provider| and returns a [=Manifest=] object:
```
npm1, line 1393:
```suggestion
The <dfn>fetch the manifest</dfn> algorithm accepts an {{IdentityProvider}} |provider| and returns a [=Manifest=] object:
```

npm1: There are several like this below, easy to find though

spec/index.bs
```diff
@@ -1173,21 +1180,33 @@ requests.
...
-<code><dfn for="FederatedCredential" method>
+<!-- ============================================================ -->
+## The IdentityCredential Interface ## {#browser-api-identity-credential-interface}
+<!-- ============================================================ -->
+
+This specification introduces a new type of {{Credential}}, called an {{IdentityCredential}}:
+
+<pre class="idl">
+  [Exposed=Window, SecureContext]
+  interface IdentityCredential : Credential {
```
npm1, line 1191: The methods should be declared here in this idl. It does not make sense to have them all spread out in the sections. The actual definitions can be though

spec/index.bs
```diff
@@ -1568,12 +1587,12 @@ When this method is invoked, the user agent MUST execute the following algorithm
   1. Let |promise| be a new {{Promise}}.
```
npm1, line 1587: Once we have defined the methods in a central place in IDL, it is much better to change "When this method is invoked" to actually name the method being defined

spec/index.bs
samuelgoto: Ah, yes, good catch, done everywhere.

spec/index.bs
samuelgoto: Do you mind if I address that in a separate PR and keep this PR exclusively to a rename? Seems worth separating the concerns here, since renaming is a key thing in and on itself?

spec/index.bs
samuelgoto: SGTM, will resolve that in a separate PR if you don't mind, so that I keep this exclusively about a rename.

spec/index.bs
npm1: Sure!

npm1 (approved):

PR #292, opened 2022-06-29
Remove logout and revoke from explainer

npm1:


explainer.md
```diff
@@ -233,61 +208,6 @@ The `login` method will call into the IDP's
 [Client Metadata Endpoint](#client_metadata_endpoint) and
 [Token Endpoint](#token_endpoint).
 
-
-##### Mediation Values
-
-The `get` request for the `FederatedCredential` uses the `mediation` parameter
```
npm1, line 239: Removed these since this makes it sound like the API currently supports these but we do not. But I could keep, WDYT

cbiesinger (approved):

explainer.md
samuelgoto: LGTM

samuelgoto:
Thanks!

PR #304, opened 2022-06-30
Improve the StateMachine

npm1:
This PR defines the State Machine a bit more precisely by noting it is based on a map for the RP, IDP, and account id. Some algorithms are improved to make more sense, and some duplicate definitions are removed. Also fixes https://github.com/fedidcg/FedCM/issues/298, https://github.com/fedidcg/FedCM/issues/297, https://github.com/fedidcg/FedCM/issues/296, and part of https://github.com/fedidcg/FedCM/issues/295.

***
<a href="https://pr-preview.s3.amazonaws.com/npm1/FedCM/pull/304.html" title="Last updated on Jul 13, 2022, 7:50 PM UTC (3c02878)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/304/cd9737d...npm1:3c02878.html" title="Last updated on Jul 13, 2022, 7:50 PM UTC (3c02878)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -47,7 +47,8 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     type: dfn
         text: origin; for: html-origin-def; url: origin.html#concept-origin
         text: Content-Type Metadata; url: urls-and-fetching.html#content-type
-
+    type: method
+        text: setTimeout; url: timers-and-user-prompts.html#dom-settimeout
 
 spec: webidl; urlPrefix: https://webidl.spec.whatwg.org/
     for: Promise
@@ -428,7 +429,7 @@ provides the main security boundary on the web.
     * [[OIDC-Connect-Core]]
 
 : <dfn>site</dfn>
-:: A set of [=origins=] that are all [=site/same site=] with each other. Note that
+:: A set of [=origins=] that are all [=same site=] with each other. Note that
     there are problems ([[PSL-PROBLEMS]]) with using [=registrable domains=] as
     a logical boundary.
 
@@ -445,12 +446,10 @@ provides the main security boundary on the web.
     site may or may not know that multiple user identifiers refer to the same user.
 
 :  <dfn>Privacy Policy</dfn>
-:: The policies described at
-    {{client_metadata_endpoint_response/privacy_policy_url}}.
+:: The policies described at {{client_metadata/privacy_policy_url}}.
 
 :  <dfn>Terms of Service</dfn>
-:: The policies described at
-    {{client_metadata_endpoint_response/terms_of_service_url}}.
+:: The policies described at {{client_metadata/terms_of_service_url}}.
 
 : <dfn>registered</dfn>
 :: The account is `registered` if the user agent thinks that the user has registered an account in
@@ -641,8 +640,7 @@ It may have the following properties:
     :   <dfn>background_color</dfn> (optional)
     ::  Background [=color=] for [=IDP=]-branded widgets such as buttons.
     :   <dfn>color</dfn> (optional)
-    ::  [=color=] for text on [=IDP=] branded widgets using
-        [=background_color=].
+    ::  [=color=] for text on [=IDP=] branded widgets.
     :   <dfn>icons</dfn> (optional)
     ::  A list of [=Icon JSON=] objects.
 </dl>
@@ -790,7 +788,7 @@ Sec-FedCM-CSRF: ?1
 
 The file is parsed expecting the following properties:
 
-<dl dfn-type="argument" dfn-for="client_metadata_endpoint_response">
+<dl dfn-type="argument" dfn-for="client_metadata">
     :   <dfn>privacy_policy_url</dfn> (optional)
     ::  A link to the [=RP=]'s [=privacy policy=].
     :   <dfn>terms_of_service_url</dfn> (optional)
@@ -984,8 +982,8 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
   };
 </pre>
 
-<dl dfn-type="argument" dfn-for="IdentityCredential">
-    :   <dfn>token</dfn>
+<dl>
+    :   <b>{{IdentityCredential/token}}</b>
     ::  The {{IdentityCredential/token}}'s attribute getter returns the value it is set to.
         It represents the minted [=token=] provided by the [=IDP=].
 </dl>
@@ -1027,12 +1025,12 @@ dictionary IdentityProvider {
 };
 </xmp>
 
-<dl dfn-type="argument" dfn-for="IdentityProvider">
-    :   <dfn>configURL</dfn>
+<dl>
+    :   <b>{{IdentityProvider/configURL}}</b>
     ::  The URL of the configuration file for the identity provider.
-    :   <dfn>clientId</dfn>
+    :   <b>{{IdentityProvider/clientId}}</b>
     ::  The [=client id=] provided to the [=RP=] out of band by the [=IDP=]
-    :   <dfn>nonce</dfn>
+    :   <b>{{IdentityProvider/nonce}}</b>
     ::  A random number of the choice of the [=RP=]. It is generally used to associate a client
         session with a [=token=] and to mitigate replay attacks. Therefore, this value should have
         sufficient entropy such that it would be hard to guess.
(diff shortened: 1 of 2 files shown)


spec/index.bs
```diff
@@ -951,25 +951,24 @@ Lastly, it overrides the default implementation of [[#browser-api-rp-sign-in]].
...
+<dl dfn-type="argument" dfn-for="AccountState">
+    :   <dfn>registration state</dfn>
+    ::  Keeps track of whether the user agent is aware that the user has registered an account in
+         the [=RP=] or not.
          Can be [=registered=] or [=unregistered=] (by default).
-    :   <dfn>Session State</dfn>
-    ::  Keeps track of whether the user has an open or closed session.
+    :   <dfn>logged in state</dfn>
+    ::  Keeps track of whether the user appears to be logged into the [=RP=] with the account.
          Can be [=logged-in=] or [=logged-out=] (by default).
```
samuelgoto, line 967: do we actually know when the user gets logged out? i don't think so, since we removed the logout() method?

spec/index.bs
```diff
@@ -1082,24 +1080,37 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
         |accounts list| algorithm.
     1. If |account| is null, return.
-    1. If the {{StateMachine/Account State}} is [=unregistered=] then run the
-        [=sign-up=] algorithm for |account|, and return.
-    1. Run the [=sign-in=] algorithm for |account|.
+    1. Let |idpURL| be |provider|'s {{IdentityProvider/configURL}}.
+    1. Let |idpOrigin| be the [=origin=] corresponding to |idpURL|. TODO: specify how to go from
+        string to [=origin=].
+    1. Let |rpOrigin| be [=this=]'s [=Document/origin=].
+    1. Let |accountId| be |account|'s {{accounts_endpoint_response_accounts/id}}.
```
samuelgoto, line 1098: this seems a bit off ... don't you need to run the sign-up/sign-in algorithm somewhere to know which of of these is the accounts id? that is, accounts_endpoint_response_accounts is a list (of accounts), not a single account.

samuelgoto (commented):
This is a massive improvement, really excited to check this in. Some clarification questions before I go much further in this review.

spec/index.bs
npm1: The only reason we need this is to ensure we are not calling `logoutRPs` multiple times using the same credentials. But yea the naming seems confusing, any idea?

spec/index.bs
npm1: The account is already chosen from `Let |account| be the result of running the [=select an account=] from the |accounts list| algorithm.`? And yes the naming is bad but if you look at the Account JSON dfn it uses dfn-for=accounts_endpoint_response_accounts. I'll change that.

spec/index.bs
npm1: I have renamed this to `allows logout` and added an explanation on how it is used.

npm1:
@cbiesinger could you take a look?

spec/index.bs
```diff
@@ -1081,24 +1074,37 @@ When this method is invoked, the user agent MUST execute the following algorithm
...
         [=fetch the accounts list=] algorithm.
     1. Let |account| be the result of running the [=select an account=] from the
         |accounts list| algorithm.
     1. If |account| is null, return.
-    1. If the {{StateMachine/Account State}} is [=unregistered=] then run the
-        [=sign-up=] algorithm for |account|, and return.
-    1. Run the [=sign-in=] algorithm for |account|.
+    1. Let |idpURL| be |provider|'s {{IdentityProvider/configURL}}.
+    1. Let |idpOrigin| be the [=origin=] corresponding to |idpURL|. TODO: specify how to go from
+        string to [=origin=].
```
cbiesinger, line 1090: I'm not sure if we need to specify that here beyond linking to the origin definition.

cbiesinger (approved):
Sorry for the delay!

npm1:
I did a few more changes, would appreciate it if you can take another look!

spec/index.bs
```diff
@@ -1088,27 +1082,59 @@ requests.
...
+        1. If the user consents, run the [=sign-in=] algorithm with |accountState|.
+    1. Otherwise:
+        1. Let |account| be the result of running the [=select an account=] from the
+            |accountsList|.
+        1. If |account| is null, return null.
+        1. Let |accountState| be the result of running the [=compute account state=] algorithm
+            given |provider| and |account|.
+        1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then run the
+            [=sign-up=] algorithm with |account|, |accountState|, |manifest|, and |provider|.
+        1. Otherwise, run the [=sign-in=] algorithm with |accountState|.
```
cbiesinger, line 1111: I still think it's confusing that for sign-up, the dialog is inside the signup algorithm, but for sign-in the dialog is up here...

I guess the reason is that signup always needs user consent whereas signin only needs it for the single-account case... maybe that's ok...

spec/index.bs
```diff
@@ -1088,27 +1082,59 @@ requests.
...
+    1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then return null.
+    1. Let |token| be the result of running the [=create tokens=] algorithm with |accountState|,
+        |account|'s {{account_json/id}}, and |provider|.
+    1. Let |credential| be a new {{IdentityCredential}}.
+    1. Set |credential|'s {{IdentityCredential/token}} to |token|.
+    1. return |credential|.
```
cbiesinger, line 1117: Uppercase R?

cbiesinger (approved):

spec/index.bs
npm1: Ok sure, done

spec/index.bs

spec/index.bs
npm1: Yea, that is confusing. Changed name of sign-up so that this difference it is hopefully clearer. The sign-in one does not require consent in the multi-account case (just choosing the account implies consent given), hence the difference...

npm1:
Added some more changes right on time for when Sam comes back from vacation :P

npm1:
@cbiesinger could you take a look at the changes since your last review? Sorry, didn't know Sam was going OOO

spec/index.bs
```diff
@@ -1087,13 +1084,21 @@ requests.
            Note: At some point we would like to support choosing accounts from
            multiple [=Identity Provider=]s.
       1. Let |provider| be |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"][0].
-      1. Return the [=potentially create IdentityCredential=] algorithm with |provider|.
+      1. Let |credential| be the result of running the [=potentially create IdentityCredential=]
+          algorithm with |provider|.
+      1. If |credential| is null, run {{setTimeout}} passing a [=task=] which throws a
+          {{NetworkError}}, after a timeout of 3 seconds.
+
+          Note: the purpose of having a timer here is to avoid leaking the reason causing this
```
cbiesinger, line 1092: I don't think that's right. This specific timeout is so the user actually gets to see the "Verifying" dialog, right? (As such, I don't think this needs to be in the spec, or could be specified as optional)

spec/index.bs
```diff
@@ -1176,14 +1183,12 @@ To <dfn>request consent to sign-up</dfn> the user with a given [=Account JSON=]
...
-        |provider|["{{IdentityProvider/configURL}}"]
-    1. Let the |policies| be the result of fetching the
-        |client_metadata_endpoint| with the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]]
-        header but without the [=Identity Provider=]'s cookies.  This request
-        MUST NOT follow [[RFC9110#header.location|HTTP redirects]] and instead
-        abort with an error if there are any. See also [[#idp-api-client-id-metadata-endpoint]].
+    1. Let the |clientMetadataEndpoint| url be the relative url
+        |manifest|["{{Manifest/client_metadata_endpoint}}"] of |provider|'s
+        {{IdentityProvider/configURL}}.
+    1. Let the |policies| be the result of fetching the |clientMetadataEndpoint| with the
```
cbiesinger, line 1189: Sort of unrelated to this change but why do we call this policies?

cbiesinger (approved):

spec/index.bs
npm1: No, this is the timeout for the response to the web developer in the case where the user declines or the token is not received for some other reason. It needs to be specified as it is important for privacy reasons.

spec/index.bs
npm1: Changing to |clientMetadata|

spec/index.bs
cbiesinger: But that timeout is not 3 seconds?

npm1: Oops my bad, I think it should say 3 minutes. I think that's the timeout?

npm1: Hmm never mind it says 120 seconds in the code right now. I'm not sure where I got '3 seconds from' :(

cbiesinger (approved):

spec/index.bs
npm1: Ok, moved the `setTimeout` to before the other steps, in particular before `potentially create IdentityCredential`.

PR #330, opened 2022-08-19
Refer to the right "Mozilla Persona"

fmarier:
Mozilla Personas were theme add-ons for Firefox (see https://blog.mozilla.org/addons/personas-are-now-firefox-themes/) whereas Mozilla Persona was the identity system that started its life as BrowserID.


samuelgoto:
Ah of course! Thank you!!

PR #338, opened 2022-08-25
Add mandatory IDP referrer check

npm1:
Fixes https://github.com/fedidcg/FedCM/issues/325

***
<a href="https://pr-preview.s3.amazonaws.com/npm1/FedCM/pull/338.html" title="Last updated on Aug 26, 2022, 5:27 PM UTC (01bc080)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/338/1928c23...npm1:01bc080.html" title="Last updated on Aug 26, 2022, 5:27 PM UTC (01bc080)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -862,6 +862,16 @@ The response is parsed as a JSON file expecting the following properties:
     ::  The resulting [=token=].
 </dl>
 
+The content of the [=token=] is opaque to the user agent and can contain
+anything that the [=Identity Provider=] would like to pass to the
+[=Relying Party=] to facilitate the login.
+
+NOTE: For [=Identity Providers=], it is worth considering how
+[portable](https://github.com/fedidcg/FedCM/issues/314) accounts are.
+Portability is left entirely up to [=Identity Providers=], who can choose
+between a variety of different mechanisms to accomplish it
+(e.g. [OIDC's Account Porting](https://openid.net/specs/openid-connect-account-porting-1_0.html)).
+
 For example:
 
 <div class=example>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -79,6 +79,12 @@ dl.domintro dt a {
 dl.domintro dt code {
     font-size: inherit;
 }
+.idp-normative-text {
+    background-color: rgba(165, 42, 42, 0.3);
+    margin: 16px 0px;
+    padding: 8px;
+    border-left: 8px solid brown;
+}
 </style>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -855,6 +855,10 @@ account_id=123&client_id=client1234&nonce=Ct60bD&disclosure_text_shown=true
 ```
 </div>
 
+An [=IDP=] needs to check the referrer to ensure that a malicious [=RP=] does not receive an ID
+token corresponding to another [=RP=]. In other words, the referrer must match the [=client id=]. As
```
samuelgoto, line 859:
```suggestion
token corresponding to another [=RP=]. In other words, the referrer must be represented by the [=client id=]. As
```

Referrers are origins, client_ids are arbitrary strings (which aren't origins necessarily), so string matching wouldn't work.

spec/index.bs
```diff
@@ -855,6 +855,10 @@ account_id=123&client_id=client1234&nonce=Ct60bD&disclosure_text_shown=true
 ```
 </div>
 
+An [=IDP=] needs to check the referrer to ensure that a malicious [=RP=] does not receive an ID
```
samuelgoto, line 858: Is this a normative conformace requirement to IDPs? If so, is there something like NOTE: that we could use to collect those?

samuelgoto (approved):
LGTM 

Will merge when comments are addressed

spec/index.bs
npm1: Anything within a note is non-normative text. So I chose to not put it inside a note.

npm1:
ptal

spec/index.bs
npm1: Added style per offline discussion.

spec/index.bs
```diff
@@ -855,6 +861,13 @@ account_id=123&client_id=client1234&nonce=Ct60bD&disclosure_text_shown=true
 ```
 </div>
 
+<div class=idp-normative-text>
+An [=IDP=] needs to check the referrer to ensure that a malicious [=RP=] does not receive an ID
```
kdenhartog, line 865: Nit: using RFC 2119 language here to identify this more clearly as a normative statement would be helpful.

E.g. An IDP MUST check the referrer header matches the expected client id for the RP. If the referrer does not match the client id the IDP MUST reject the request.

kdenhartog (commented):
Editorial nit added

PR #345, opened 2022-09-01
The idp sign in status api

samuelgoto:
Writes down more formally one of the proposals to address the timing attack problem:

https://github.com/fedidcg/FedCM/issues/230#issuecomment-1233290946

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Sep 27, 2022, 8:38 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Ffedidcg%2FFedCM%2Fd54e941cdb4fea4567c4fc1d262a776b13028287%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>500 Internal Server Error</title>
</head><body>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error or
misconfiguration and was unable to complete
your request.</p>
<p>Please contact the server administrator at 
 [no address given] to inform them of the time this error occurred,
 and the actions you performed just before this error.</p>
<p>More information about this error may be available
in the server error log.</p>
<hr>
<address>Apache/2.4.10 (Debian) Server at api.csswg.org Port 443</address>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20fedidcg/FedCM%23345.)._
</details>


spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+
+A strict subset of the [=push model=] that seems promising is to narrow the pushing to exclusively
+the user's sign-in status in the IdP (as opposed to the entirety of the user's account data,
+e.g. name, email).
+
+Since the sign-in status is specific to a browser instance/profile, this bit can reflect reality
+almost perfectly (with the exception of server-side invalidations, e.g. users changing passwords or
+deleting accounts on other devices).
+
+In this variation, the [=user agent=] exposes an API that allows an [=IDP=] to record the user's
```
yi-gu, line 1749: s/record/push?

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+IdP-Sign-in-Status: action=login
+// ... later ...
+IdP-Sign-in-Status: action=logout
+```
+</div>
+
+IDPs are also offered an extension to the [=Manifest=] JSON object to include:
+
+<dl dfn-type="argument" dfn-for="Manifest">
+    :   <dfn>signin_url</dfn> (required)
```
yi-gu, line 1780: should we make it optional? i.e. if an IDP is not comfortable with browser assisting signing in to IDP (UX concern, security concern etc.), we show a failure UI without this functionality.

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+
+We would then use the follow [=maybe fetch the accounts list=] (described below) in the [=potentially create IdentityCredential=] algorithm, as opposed to the [=fetch the accounts list=].
+
+<div algorithm>
+To <dfn>maybe fetch the accounts list</dfn> given a [=Manifest=] |manifest| and an {{IdentityProvider}}
+    |provider|:
+    1. Let |status| be the [=Sign-in Status=] of the |provider|.
+    1. If |status| is {{Sign-in Status/unknown}}
+        1. Let |accounts| be the result of the [=fetch the accounts list=] algorithm
+        1. Set the [=Sign-in Status=] of the |provider| to {{Sign-in Status/signed-in}} if |accounts|'s size > 0, {{Sign-in Status/signed-out}} otherwise
```
yi-gu, line 1803: Does "otherwise" include "no account returned due to reasons like network issues? Should we make it explicit?

spec/index.bs
samuelgoto: "record" seems more appropriate than "push"? "register" maybe?

spec/index.bs
samuelgoto: Ah, good point, done.

spec/index.bs
samuelgoto: Yes it does. I figure it would be useful to just leave it open for now to avoid low level details. WDYT?

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+
+        NOTE: This handles the case where the [=IDP=] hasn't had the chance to call the API before the accounts list is needed. This can incur into a [=timing attack=], but it is limited to 1 per [=IDP=] per [=user agent=], so not very practical.
+
+    1. If |status| is {{Sign-in Status/signed-out}}
+        1. Terminate the request
+	1. Return an empty list
```
yi-gu: What does "return an empty list" mean?

samuelgoto: Returns an "empty accounts list", [].

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+
+    1. If |status| is {{Sign-in Status/signed-in}}
+        1. Let |accounts| be the result of the [=fetch the accounts list=] algorithm
+        1. if |accounts|'s size == 0
+            1. Set the [=Sign-in Status=] of the |provider| to {{Sign-in Status/signed-out}}
+            1. Prompt the user to confirm they want to sign-in to their [=IDP=].
+            1. Navigate the user to the [=IDP=]'s {{Manifest/signin_url}} to allow them to sign-in.
+            1. Set |accounts| to the result of the [=fetch the accounts list=] algorithm
+
+            NOTE: This happens when the user's local client credentials are invalidated on the server (e.g. changing passwords or deleting accounts on a different device).
```
yi-gu, line 1821: s/happens/may happen?
There may be other reasons the account size is 0, e.g. slow network, right?

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+
+        1. Return |accounts|.
+</div>
+
+
```
yi-gu, line 1826: Should we add when the Sign-in Status could be reset to `unknown`? e.g. user clearing cookies etc.

spec/index.bs
yi-gu: My concern was it didn't mention "where" to record so "push" implies the UA. Guess "record" would work if we add "in the UA" at the end. WDYT?

spec/index.bs
yi-gu: SG.

spec/index.bs
yi-gu: Sorry I mean after we terminate the request, the flow is over so where do we return the empty list to?

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+<div algorithm>
+To <dfn>maybe fetch the accounts list</dfn> given a [=Manifest=] |manifest| and an {{IdentityProvider}}
+    |provider|:
+    1. Let |status| be the [=Sign-in Status=] of the |provider|.
+    1. If |status| is {{Sign-in Status/unknown}}
+        1. Let |accounts| be the result of the [=fetch the accounts list=] algorithm
+        1. Set the [=Sign-in Status=] of the |provider| to {{Sign-in Status/signed-in}} if |accounts|'s size > 0, {{Sign-in Status/signed-out}} otherwise
+
+        NOTE: This handles the case where the [=IDP=] hasn't had the chance to call the API before the accounts list is needed. This can incur into a [=timing attack=], but it is limited to 1 per [=IDP=] per [=user agent=], so not very practical.
```
bvandersloot-mozilla, line 1805: One counterargument that has come to me reading this is that it is also only one instance of friction for the IDP, especially if we set a user as logged in on successful non-empty account list fetch. I would be happier on the privacy end there.

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
     a user is logged into, regardless of whether the user ever uses the FedCM API or not. This means
     that all users pay the cost, and only some get the reward.
 
+<!-- ============================================================ -->
+#### The IdP Sign-in Status API #### {#the-idp-sign-in-status-api}
+<!-- ============================================================ -->
+
+Issue: this is a more formal description of what is proposed [here](https://github.com/fedidcg/FedCM/issues/230#issuecomment-1233290946).
```
npm1, line 1739: This is not an `Issue` per se. Maybe this should be a `Note` instead.

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+
+A strict subset of the [=push model=] that seems promising is to narrow the pushing to exclusively
```
npm1, line 1741:
```suggestion
A strict subset of the [=push model=] that seems promising is to only push
```

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+the user's sign-in status in the IdP (as opposed to the entirety of the user's account data,
+e.g. name, email).
+
+Since the sign-in status is specific to a browser instance/profile, this bit can reflect reality
+almost perfectly (with the exception of server-side invalidations, e.g. users changing passwords or
```
npm1, line 1746:
```suggestion
almost perfectly (with the exception of server-side invalidations, e.g. users changing password, and
```

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+deleting accounts on other devices).
+
+In this variation, the [=user agent=] exposes an API that allows an [=IDP=] to record the user's
+sign-in status: whether the user is signed-in or not.
+
+This is a compelling variation because (a) it is comparably simpler to implement by [=IDPs=] and (b)
+if the [=user agent=] knew whether the user is signed-in or not to the [=IDP=], it can guarantee that
+a user prompt will always be displayed, which makes the timing attack impossible to be performed
+invisibly (i.e. without any explicit and observable indication to the user).
```
npm1, line 1755: The last statement about impossible to be performed is not really accurate (we still allow it when status is undefined)

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+</div>
+
+IDPs are also offered an extension to the [=Manifest=] JSON object to include:
+
+<dl dfn-type="argument" dfn-for="Manifest">
+    :   <dfn>signin_url</dfn> (optional)
+    ::  A URL that allows a user to sign-in to the [=IDP=].
+</dl>
+
+Internally, each [=user agent=] keeps track of a global <dfn>Sign-in Status</dfn> map per [=IDP=] that can be:
```
npm1, line 1784: Unclear what this map's key type is. This needs to be more precise, as [=IDP=] is abstract.

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+<dl dfn-type="argument" dfn-for="Sign-in Status">
+    :   <dfn>unknown</dfn> (default)
+    ::  By default, the [=user agent=] assumes that the user's signed in status is undefined.
+    :   <dfn>signed-in</dfn>
+    ::  The user has explicitly signed-in to the IDP.
+    :   <dfn>signed-out</dfn>
+    ::  The user has explicitly signed-out to the IDP.
+</dl>
+
+We would then use the following [=maybe fetch the accounts list=] in the [=potentially create IdentityCredential=] algorithm, as opposed to the [=fetch the accounts list=].
```
npm1, line 1795: It may be possible to abort even earlier (before the manifest fetch even) if we know the sign in status, no?

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+
+    1. If |status| is {{Sign-in Status/signed-out}}
+        1. Terminate the request
+        1. Return an empty list
+
+        NOTE: By terminating the request here before running [=fetch the accounts list=] algorithm we prevent the timing attack to be performed without any user prompt.
+
+    1. If |status| is {{Sign-in Status/signed-in}}
+        1. Let |accounts| be the result of the [=fetch the accounts list=] algorithm
+        1. if |accounts|'s size == 0
```
npm1, line 1815:
```suggestion
        1. If |accounts|'s size is 0:
```

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+            1. Set the [=Sign-in Status=] of the |provider| to {{Sign-in Status/signed-out}}
+            1. Prompt the user to confirm they want to sign-in to their [=IDP=].
+            1. Navigate the user to the [=IDP=]'s {{Manifest/signin_url}} to allow them to sign-in.
```
npm1, line 1818: This is too handwavy. What does navigate mean? Create a new browsing context? Does this unload the existing page?

spec/index.bs
```diff
@@ -1732,6 +1732,98 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+            1. Set |accounts| to the result of the [=fetch the accounts list=] algorithm
```
npm1, line 1819: This is supposed to happen after the user signed in, I guess? It probably should be in its own task. Also, is there no way for the navigation to signal failure?

npm1 (changes requested):

spec/index.bs
samuelgoto: I changed from "record" to "signal" and added the "to the user agent" that you suggested.

spec/index.bs
samuelgoto: Ah, this returns an empty list to the algorithm that is going to use it, the "potentially create IdentityCredential" which handles gracefully when an empty list is returned. Removed the "terminate the request".

spec/index.bs
samuelgoto: Done.

spec/index.bs
samuelgoto: Ah, good point. I'm going to leave this as a NOTE for now, because I think we haven't quite thought about this enough.

spec/index.bs
samuelgoto: Done.

spec/index.bs
samuelgoto: Done.

spec/index.bs
samuelgoto: Yeah, agreed.

We use the "unknown" state exclusively for backwards compatibility, but we've been wondering if we would be able to "deprecate it" after some point in time, once IdPs have had a chance to "keep all of the user's sign-in status up to date with the browser".

WDYT?

bvandersloot-mozilla: It depends on your deprecation timeline. It may make more sense to just have the friction of one more click per-user per-IDP from where we sit. But we are kind of splitting hairs on short-term minor leak vs short-term minor inconvenience here.

 A middle ground may be to make it optional with a note that it will be standardized to  to be removed by a particular sunset date and when the two-click is added to the spec make a successful account fetch after the first click result in a login bit being set.

spec/index.bs
samuelgoto: Yeah, good point. I broke this into a different algorithm and tried to give a high level description. Not super precise quite yet, but a bit more. WDYT?

spec/index.bs
samuelgoto: I broke the step before into its own algorithm and used tasks to be clear about what order this should happen.

WDYT?

spec/index.bs
samuelgoto: Oh, yeah, good point, added another condition here.

I considered breaking down this algorithm into smaller parts so that it integrates better with the "potentially create IdentityCredential" algorithm, but it started getting messy and deviating from the algorithm that we've been using. I think it is helpful to keep the algorithm close to what we've been using for clarity and deal with integrating with the existing algorithms later.

WDYT?

spec/index.bs
samuelgoto: Just thinking out loud here, so apologies if this isn't well articulated.

> But we are kind of splitting hairs on short-term minor leak vs short-term minor inconvenience here.

I think I have a different intuition than you do on the level of "minor inconvenience", because if the default value is the `unknown` state, the two-tap interstitial would show, by default, to all of the IDP's users at the moment that they start using the API (so, while it is a small inconvenience for a single user, if you multiple by O(B) of users, that becomes a problem, I think).

> It depends on your deprecation timeline.
> A middle ground may be to make it optional with a note that it will be standardized to to be removed by a particular sunset date

I thought a bit about this, but I'm starting to suspect that we would never be able to fully deprecate / remove the `unknown` state. Consider the following example in the long term:

1. Let's say an IDP has O(M) of users who are already logged in to their services 
2. An IDP decides to make the FedCM integration and starts calling the `IdentityProvider.login()` API.
3. There is a non-zero amount of the IDP's users who are going to be in a state that they **are** logged in (because their sessions are valid) but the IDP has never called the API.

How does an IDP know when "100% of my users that **are** logged in have **also** gone through the explicit `IdentityProvider.login()`"?

One idea that occurs to me is that the IDP could wait until  "the longest expiration time of the last user's session I issued **after** I started calling `IdentityProvider.login()` for new sessions".

We heard from some IdPs that they in practice "never" expire the user's sessions (e.g. google.com cookies are valid for 2 years I think).

So, I think, even long term, we are going to need to be able to represent the state `unknown`.

I think the question then is what to do in that state.

I think one option that is not very controversial is to allow IdPs to:

1. opt-into showing the two-click interstitial on `unknown` or
2. opt-into treating `unknown` as `signed-out` and not show any UI while their users transition from `unknown` to `signed-out`

I'm wondering if we could put the "fetch" under heuristics (e.g. site engagement scores).

> I would be happier on the privacy end there.

You are raising a good point here. 

I don't know what the right balance is here yet, but I have an intuition that I'm leaning towards the "make IdP's deployment easier" end (given the privacy benefit: it is a single-time-per-origin fetch that **also** requires a time fingerprint correlation - that's quite constrained).

More random ramblings than organized thoughts, but WDYT?

spec/index.bs
```diff
@@ -1732,6 +1732,138 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+interface IdentityProvider {
+  static undefined login();
+  static undefined logout();
+};
+</xmp>
+
+<div algorithm=login>
+When the {{IdentityProvider/login()}} static method is invoked the user agent MUST execute the following
+steps:
+    1. Set the [=Sign-in Status=] to {{Sign-in Status/signed-in}}.
```
npm1, line 1772: The `Sign-in Status` needs to be keyed by some origin presumably? Note that this should probably be some global map which the user agent stores for ~a long time.

spec/index.bs
```diff
@@ -1732,6 +1732,138 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+</div>
+
+<div algorithm=logout>
+When the {{IdentityProvider/logout()}} static method is invoked the user agent MUST execute the following
+steps:
+    1. Set the [=Sign-in Status=] to {{Sign-in Status/signed-out}}.
```
npm1, line 1778: ditto need to change the status of some global map

spec/index.bs
```diff
@@ -1732,6 +1732,138 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+    :   <dfn>signed-out</dfn>
+    ::  The user has explicitly signed-out to the IDP.
+</dl>
+
+The [=user agent=] uses the following [=maybe fetch the accounts list=] in the [=potentially create IdentityCredential=] algorithm, as opposed to the [=fetch the accounts list=] algorithm. It would also return early on the [=potentially create IdentityCredential=] algorithm if the user was {{Sign-in Status/signed-out}}.
+
+<div algorithm>
+To <dfn>maybe fetch the accounts list</dfn> given a [=Manifest=] |manifest| and an {{IdentityProvider}}
+    |provider|:
+    1. Let |status| be the [=Sign-in Status=] of the |provider|.
```
npm1, line 1828: `provider` is an `IdentityProvider` object that just got created in this page (probably need to compute the origin?)

spec/index.bs
```diff
@@ -1732,6 +1732,138 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+        1. Return |accounts|.
+</div>
+
+<div algorithm>
+To <dfn>Sign-in to the IDP</dfn> given a [=Manifest=] |manifest| and an {{IdentityProvider}}
+    |provider|:
+    1. Assert that the [=Sign-in Status=] of the |provider| is {{Sign-in Status/signed-out}}
+    1. In parallel, wait for one these tasks to return:
+        1. Open a dialog that directs the user user to the [=IDP=]'s {{Manifest/signin_url}}.
+        1. Wait until the [=Sign-in Status=] of the |provider| becomes {{Sign-in Status/signed-in}}
```
npm1, line 1859: The waits here are simultaneous, so maybe wait until one of the following ... then continue? Also please add a note to point out that this wait means that the website is going to be useless during this time

spec/index.bs
npm1: Yi's implementation checks the sign in status before fetching the manifest.

samuelgoto: Yep, that's covered in the text right now in this sentence: 

```
It would also return early on the [potentially create IdentityCredential](http://localhost:8000/#potentially-create-identitycredential) algorithm if the user was [signed-out](http://localhost:8000/#dom-sign-in-status-signed-out).
```

spec/index.bs
samuelgoto: Done, broke that on a separate algorithm.

spec/index.bs
samuelgoto: That's correct. That's written down here:

```
Internally, each [user agent](https://infra.spec.whatwg.org/#user-agent) keeps track of a global Sign-in Status map per [IDP](https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/345.html#idp). The [keys](https://infra.spec.whatwg.org/#map-getting-the-keys) in the [Sign-in Status](https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/345.html#sign-in-status) map is the [origin](https://html.spec.whatwg.org/multipage/origin.html#concept-origin) of the [IDP](https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/345.html#idp). The [values](https://infra.spec.whatwg.org/#map-getting-the-values) in the [Sign-in Status](https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/345.html#sign-in-status) map are status objects which can be one of the following values:
```

And used consistently with the state machine maps.

spec/index.bs
samuelgoto: it does change the status of the global map, right?

spec/index.bs
samuelgoto: > The waits here are simultaneous, so maybe wait until one of the following ... then continue?

Done.

> so please add a note to point out that this wait means that the website is going to be useless during this time

I'm assuming you are not suggesting that we block the JS/UI thread, right? That is, unlike prompt() and alert(), we wouldn't be "stopping the JS execution", right?

And if not, I'm also assuming that we don't need to be opinionated about the modality of the dialog in the spec, right? That is, each browser vendor may choose independently how to present this to their users (e.g. some may choose not to make this a modal dialog and "allow the user to still use the website"?)?

spec/index.bs
samuelgoto: I think it is going to take us a lot of deployment trial and error here before we understand what the trade-offs are, so I added a note that says "Albeit small, removing this attack surface is an active area of investigation." to be transparent that this is something that we expect to be working on and learning from in the coming months.

With that, resolving this now and will follow up on this challenge.

samuelgoto:
friendly ping?

spec/index.bs
```diff
@@ -1032,32 +1032,32 @@ This specification starts by introducing an extension to the
 </pre>
 
 The {{IdentityCredentialRequestOptions}} contains a list of
-{{IdentityProvider}}s that the [=Relying Party=] supports and has
+{{IdentityProviderConfig}}s that the [=Relying Party=] supports and has
```
npm1, line 1035: Can we decouple this rename from the PR? It seems unrelated. Please send a separate PR for this.

spec/index.bs
npm1: This needs to use a key then?

spec/index.bs
```diff
@@ -1732,6 +1732,178 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+
+The [=user agent=] uses the following [=maybe fetch the accounts list=] in the [=potentially create IdentityCredential=] algorithm, as opposed to the [=fetch the accounts list=] algorithm. It would also return early on the [=potentially create IdentityCredential=] algorithm if the user was {{Sign-in Status/signed-out}}.
+
+<div algorithm>
+To <dfn>maybe fetch the accounts list</dfn> given a [=Manifest=] |manifest| and an {{IdentityProviderConfig}}
+    |provider|:
+    1. Let |status| be the [=Sign-in Status=] of the |provider|.
+    1. If |status| is {{Sign-in Status/unknown}}
+        1. Let |accounts| be the result of the [=fetch the accounts list=] algorithm
+        1. Set the [=Sign-in Status=] of the |provider| to {{Sign-in Status/signed-in}} if |accounts|'s size > 0, {{Sign-in Status/signed-out}} otherwise
```
npm1, line 1871:
```suggestion
        1. Set the [=Sign-in Status=] of the |provider| to {{Sign-in Status/signed-in}} if |accounts| is non-empty, {{Sign-in Status/signed-out}} otherwise
```

spec/index.bs
```diff
@@ -1732,6 +1732,178 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+
+    1. If |status| is {{Sign-in Status/signed-out}}
+        1. Return an empty list
+
+        NOTE: By terminating the request here before running [=fetch the accounts list=] algorithm we prevent the timing attack to be performed without any user prompt.
+
+    1. If |status| is {{Sign-in Status/signed-in}}
+        1. Let |accounts| be the result of the [=fetch the accounts list=] algorithm
+        1. If |accounts|'s size is 0:
+            1. Set the [=Sign-in Status=] of the |provider| to {{Sign-in Status/signed-out}}
```
npm1, line 1883: ditto

spec/index.bs
samuelgoto: I defined IdentityProvider.login() and IdentityProvider.logout(), which requires me to change this definition (bikeshed complains otherwise). I think that the JS API is an important part of this PR,  so I'd rather keep it.

spec/index.bs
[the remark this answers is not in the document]
samuelgoto: Ah, good point. Done, here and elsewhere.

spec/index.bs
```diff
@@ -1732,6 +1732,178 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+    :   <dfn>signed-out</dfn>
+    ::  The user has explicitly signed-out to the IDP.
+</dl>
+
+The [=user agent=] uses the following [=maybe fetch the accounts list=] in the [=potentially create IdentityCredential=] algorithm, as opposed to the [=fetch the accounts list=] algorithm. It would also return early on the [=potentially create IdentityCredential=] algorithm if the user was {{Sign-in Status/signed-out}}.
+
+<div algorithm>
+To <dfn>maybe fetch the accounts list</dfn> given a [=Manifest=] |manifest| and an {{IdentityProviderConfig}}
+    |provider|:
+    1. Let |status| be the [=Sign-in Status=] of the |provider|.
```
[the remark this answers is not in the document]
samuelgoto: Done.

spec/index.bs
```diff
@@ -1732,6 +1732,178 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+    |provider|:
+    1. Assert that the [=Sign-in Status=] of the |provider| is {{Sign-in Status/signed-out}}
+    1. In parallel, wait until one of the following tasks returns to continue:
+        1. Open a dialog that directs the user to the [=IDP=]'s {{Manifest/signin_url}}.
+        1. Wait until the [=Sign-in Status=] of the |provider| becomes {{Sign-in Status/signed-in}}
+            1. Close the dialog
+            1. Return the result of the [=fetch the accounts list=] algorithm
+        1. Wait until the user explicitly cancels the dialog
+            1. Close the dialog
+            1. Return null
```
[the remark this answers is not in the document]
samuelgoto: Done.

spec/index.bs
```diff
@@ -1732,6 +1732,178 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+        1. Return an empty list
+
+        NOTE: By terminating the request here before running [=fetch the accounts list=] algorithm we prevent the timing attack to be performed without any user prompt.
+
+    1. If |status| is {{Sign-in Status/signed-in}}
+        1. Let |accounts| be the result of the [=fetch the accounts list=] algorithm
+        1. If |accounts|'s size is 0:
+            1. Set the [=Sign-in Status=] of the |provider| to {{Sign-in Status/signed-out}}
+            1. Ask the user to confirm they want to sign-in to their [=IDP=]
+            1. If they decline, return null
```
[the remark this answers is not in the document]
samuelgoto: Done.

spec/index.bs
samuelgoto: done.

spec/index.bs
npm1: Still not using key

spec/index.bs
npm1: Ping? Why are you resolving the comments without acting on them

npm1 (commented):
Looks good modulo comments. Also, the change of name of the IdentityProvider dict affects our implementation. We should think if it is necessary, and if it is file a bug to implement this change asap

spec/index.bs
samuelgoto: Ah, blargh done now.

spec/index.bs
samuelgoto: done here too.

samuelgoto:
> Looks good modulo comments. Also, the change of name of the IdentityProvider dict affects our implementation. We should think if it is necessary, and if it is file a bug to implement this change asap

Seems like we agreed that we should merge this and keep the separation that this is a work in progress with the headers.

PTAL?

spec/index.bs
```diff
@@ -1783,6 +1783,185 @@ While the [=push model=] seems like an improvement for privacy, the current API
...
+When site data (e.g. cookies) are cleared manually by the user, the [=user agent=] also sets the [=Sign-in Status=] to {{Sign-in Status/unknown}}.
+
+IDPs are also offered an extension to the [=Manifest=] JSON object to include:
+
+<dl dfn-type="argument" dfn-for="Manifest">
+    :   <dfn>signin_url</dfn> (optional)
+    ::  A URL that allows a user to sign-in to the [=IDP=].
+</dl>
+
+Internally, each [=user agent=] keeps track of a global <dfn>Sign-in Status</dfn> map per [=IDP=], initially an empty [=map=]. The [=map/keys=] in the [=Sign-in Status=] map is the [=origin=] of the [=IDP=]. The [=map/values=] in the [=Sign-in Status=] map are <dfn>status</dfn> objects which can be one of the following values:
```
npm1, line 1905: It might make sense to move this higher up? It is a bit weird that it is used so much before being defined.

npm1 (approved):

spec/index.bs
samuelgoto: ah, good point. done!

PR #351, opened 2022-09-23
Document merging get() calls for multiple IDP usage

cbiesinger:
***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/351.html" title="Last updated on Oct 14, 2022, 7:48 PM UTC (0a2d80a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/351/e842e7e...cbiesinger:0a2d80a.html" title="Last updated on Oct 14, 2022, 7:48 PM UTC (0a2d80a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1077,7 +1077,7 @@ NOTE: The {{CredentialRequestOptions/mediation}} flag is currently not used.
 The {{CredentialRequestOptions/signal}} is used as an abort signal for the
 requests.
 
-<div algorithm>
+<div algorithm="DiscoverFromExternalSource">
     When the {{IdentityCredential}}'s
     <dfn for="IdentityCredential" method>\[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)</dfn>\
     algorithm is invoked, the user agent MUST execute the following steps:
@@ -1092,11 +1092,51 @@ requests.
           Note: the purpose of having a timer here is to avoid leaking the reason causing this
           method to return null. If there was no such timer, the developer could easily infer
           whether the user has an account with the [=IDP=] or not, or whether the user closed the UI without granting permission to share the [=IDP=] account information with the [=RP=].
+      1. Let |all_options| be a list of pairs of {{CredentialRequestOptions}}
+          and {{Promise}} objects.
+          This list is reused for all {{Credential/[[DiscoverFromExternalSource]]}}
+          calls within the same global object.
+      1. Create a new promise and append (|options|, new promise) to |all_options|.
+      1. If a load event listener (next step) has already been registered or
+        a task has already been posted (step after next), return the promise.
+      1. If the document is not <a spec=HTML>completely loaded</a>, register
+          an [=event handler=] on the {{Window}} object for the
+          {{Window/load}} event. When the listener is called, proceed to the
+          next step.
+
+          Note: Typically, RPs interact with IDPs using an SDK provided by the
+          IDP. It is desirable to be able to show multiple IDPs together in the
+          account chooser even with IDP SDKs that do not cooperate with each
+          other. This and the next step make this possible.
+      1. [=Queue a global task=] where the global object is the current window
+          and the task source is the <dfn export>federated credential management
+          task source</dfn> and steps is "go to the next step".
+      1. Let |options| be the result of running the [=combine CredentialRequestOptions=]
+          algorithm with |all_options|.
       1. Let |provider| be |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"][0].
       1. Let |credential| be the result of running the [=potentially create IdentityCredential=]
           algorithm with |provider|.
-      1. If |credential| is null, wait for the task that throws a {{NetworkError}}, otherwise return
-          |credential|.
+      1. If |credential| is null, wait for the task that throws a {{NetworkError}}, otherwise
+          find the pair in |all_options| where the {{IdentityProviderConfig/configURL}}
+          matches the selected |credential|. Resolve the promise in that pair with
+          this |credential| and reject all other promises in the list with
+          {{NetworkError}}.
+      1. If, at any time, any of the |options|.{{CredentialRequestOptions/signal}}s
+          in |all_options| is aborted, abort further processing, close any UI
+          that may have been shown as part of the request, and reject all
+          promises with an {{AbortError}}.
+      1. Clear |all_options|.
+</div>
+
+<div algorithm>
+To <dfn>combine CredentialRequestOptions</dfn>:
+    1. Let |all_options| be the input options
+    1. Let |options| be a new {{CredentialRequestOptions}} object
+    1. The {{CredentialRequestOptions}}.{{CredentialRequestOptions/identity}}.
+        {{IdentityCredentialRequestOptions/providers}} arrays of all objects
+        in the list are concatenated and set as |options|`.identity.providers`.
+    1. If there is a duplicate provider configURL in the array, all
+        promises are rejected.
 </div>
 
 <div algorithm>


spec/index.bs
```diff
@@ -1424,6 +1424,28 @@ steps:
...
+<!-- ======================================================= -->
+## Merging multiple get() calls ## {#browser-api-merging-calls}
+<!-- ======================================================= -->
+
+Typically, RPs interact with IDPs using an SDK provided by the IDP. It is
+desirable to be able to show multiple IDPs together in the account chooser
+even with IDP SDKs that do not cooperate with each other. To make this
+possible, the UA has to merge get() calls as follows:
+
+    1. If `navigator.credentials.get()` is called before or during onload the
```
npm1, line 1436: Need to clarify this is window onload

spec/index.bs
```diff
@@ -1424,6 +1424,28 @@ steps:
...
+        UA saves the provided options and delays further processing until
+        immediately after onload.
```
npm1, line 1438: Unclear what immediately after means. Microtask/task/something else?

spec/index.bs
```diff
@@ -1424,6 +1424,28 @@ steps:
...
+    1. If `navigator.credentials.get()` is called after onload, the UA saves
+        the provided options, enqueues a microtask and delays further
+        processing until that microtask.
+    1. When resuming processing, the saved options from all calls to get are
+        combined as follows:
+      1. The provider arrays are concatenated
```
npm1, line 1444: What happens if there are repeated configURLs?

npm1 (commented):
Could we integrate the changes into the existing processing logic in the spec?

spec/index.bs

spec/index.bs

spec/index.bs

spec/index.bs
```diff
@@ -1424,6 +1424,28 @@ steps:
     1. <a for=Promise>Resolve</a> |promise| with [undefined] and return |promise|.
 </div>
 
+<!-- ======================================================= -->
```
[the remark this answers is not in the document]
cbiesinger: This should be done, please take another look

spec/index.bs
```diff
@@ -1112,6 +1112,24 @@ requests.
           Note: the purpose of having a timer here is to avoid leaking the reason causing this
           method to return null. If there was no such timer, the developer could easily infer
           whether the user has an account with the [=IDP=] or not, or whether the user closed the UI without providing consent to share the [=IDP=] account information with the [=RP=].
+      1. If the load event on the window object has not fired yet, or is
```
npm1, line 1115: Can we link to `load` event

spec/index.bs
```diff
@@ -1112,6 +1112,24 @@ requests.
...
+          currently being fired, save the provided options and delay further
```
npm1, line 1116: You should use the variable name. "|options|". Also save it where?

spec/index.bs
```diff
@@ -1112,6 +1112,24 @@ requests.
...
+          processing until a microtask after that load event.
+
+          Note: Typically, RPs interact with IDPs using an SDK provided by the
+          IDP. It is desirable to be able to show multiple IDPs together in the
+          account chooser even with IDP SDKs that do not cooperate with each
+          other. This and the next step make this possible.
+      1. Otherwise, enqueue a microtask and delay further processing until that
```
npm1, line 1123: Links missing here too

spec/index.bs
```diff
@@ -1112,6 +1112,24 @@ requests.
...
+          microtask.
+      1. When resuming processing, the saved options from all calls to get are
```
npm1, line 1125: The scope of the saving here is unclear since it's not clear where this is being saved...

npm1, line 1125: Should probably be a new function

spec/index.bs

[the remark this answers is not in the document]

spec/index.bs

spec/index.bs

spec/index.bs

cbiesinger (commented):
Please take another look!

spec/index.bs
```diff
@@ -1092,11 +1092,51 @@ requests.
...
+      1. If the document is not <a spec=HTML>completely loaded</a>, register
+          an [=event handler=] on the {{Window}} object for the
+          {{Window/load}} event. When the listener is called, proceed to the
+          next step.
+
+          Note: Typically, RPs interact with IDPs using an SDK provided by the
+          IDP. It is desirable to be able to show multiple IDPs together in the
+          account chooser even with IDP SDKs that do not cooperate with each
+          other. This and the next step make this possible.
+      1. [=Queue a global task=] where the global object is the current window
```
samuelgoto, line 1111: A couple of things are unclear to me in this PR:

(a) what happens to all of the promises that are in the |all_options| array that aren't picked? Do they get resolved? Failed? I think that, as specified, they never resolve? Is that deliberate?

(b) what happens to the AbortSignals that were passed? If one of the AbortSignals that was passed in the |all_options| gets aborted, does that get ignored?

spec/index.bs
```diff
@@ -1112,6 +1112,24 @@ requests.
...
+          microtask.
+      1. When resuming processing, the saved options from all calls to get are
+          combined as follows:
+        1. The provider arrays are concatenated
+        1. If there is a duplicate provider configURL in the array, all
+            promises are rejected.
+        1. All AbortSignals are saved and if any is aborted, all further
+            processing is aborted and the account chooser, if any, is closed.
+            (TODO: What is the desirable behavior here?)
       1. Let |provider| be |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"][0].
```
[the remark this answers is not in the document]
cbiesinger: Yes... I wanted to decouple this specific behavior of merging get calls from general multi-IDP support. Do you think I should change this PR to cover both?

spec/index.bs
cbiesinger: For a): See the paragraph starting at line 1119

For b): See the paragraph starting at line 1124.

cbiesinger (commented):
Sam, please see my responses below.

spec/index.bs
```diff
@@ -1092,11 +1092,51 @@ requests.
           Note: the purpose of having a timer here is to avoid leaking the reason causing this
           method to return null. If there was no such timer, the developer could easily infer
           whether the user has an account with the [=IDP=] or not, or whether the user closed the UI without granting permission to share the [=IDP=] account information with the [=RP=].
+      1. Let |all_options| be a list of pairs of {{CredentialRequestOptions}}
+          and {{Promise}} objects.
+          This list is reused for all {{Credential/[[DiscoverFromExternalSource]]}}
```
npm1, line 1097: Generally how this is handled is that we add associated members to the relevant object. So here it would probably be "Each Window has an associated allOptions which is initially bla"

spec/index.bs
```diff
@@ -1092,11 +1092,51 @@ requests.
...
+          calls within the same global object.
+      1. Create a new promise and append (|options|, new promise) to |all_options|.
```
npm1, line 1099: Might be clearer to name the new promise some var.

spec/index.bs
```diff
@@ -1092,11 +1092,51 @@ requests.
...
+      1. If a load event listener (next step) has already been registered or
+        a task has already been posted (step after next), return the promise.
+      1. If the document is not <a spec=HTML>completely loaded</a>, register
+          an [=event handler=] on the {{Window}} object for the
+          {{Window/load}} event. When the listener is called, proceed to the
```
npm1, line 1104: It is just better to add new algorithm as the listener, and that algorithm is actually the 'next steps'

spec/index.bs
```diff
@@ -1092,11 +1092,51 @@ requests.
           Note: the purpose of having a timer here is to avoid leaking the reason causing this
           method to return null. If there was no such timer, the developer could easily infer
           whether the user has an account with the [=IDP=] or not, or whether the user closed the UI without granting permission to share the [=IDP=] account information with the [=RP=].
+      1. Let |all_options| be a list of pairs of {{CredentialRequestOptions}}
+          and {{Promise}} objects.
+          This list is reused for all {{Credential/[[DiscoverFromExternalSource]]}}
+          calls within the same global object.
+      1. Create a new promise and append (|options|, new promise) to |all_options|.
+      1. If a load event listener (next step) has already been registered or
```
npm1, line 1100: Could use a boolean to track this more formally.

spec/index.bs
```diff
@@ -1092,11 +1092,51 @@ requests.
...
+          Note: Typically, RPs interact with IDPs using an SDK provided by the
+          IDP. It is desirable to be able to show multiple IDPs together in the
+          account chooser even with IDP SDKs that do not cooperate with each
+          other. This and the next step make this possible.
+      1. [=Queue a global task=] where the global object is the current window
+          and the task source is the <dfn export>federated credential management
+          task source</dfn> and steps is "go to the next step".
+      1. Let |options| be the result of running the [=combine CredentialRequestOptions=]
+          algorithm with |all_options|.
       1. Let |provider| be |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"][0].
```
npm1, line 1116: This would need to not limit to the first one, and the "potentially create IdentityCredential" would need to allow an array of providers instead of a single one.

npm1 (commented):
Hmm thinking about this more, it might actually be better to have a section for multiIDP that monkey patches the main sections because we are not fully confident on the design here yet and because Mozilla may want to implement the single IDP case like us for now, so having this in the main algorithm might cause some confusion.

cbiesinger:
I made https://github.com/fedidcg/FedCM/pull/368 to support multi IDP in general, as a prerequisite for this PR.

samuelgoto:
Two high level comments:

- can we add a section with use cases / requirements (e.g. a table with "with vs without gesture" vs "single vs multiple")
- can we add examples for each of these cases?
- should we introduce a "mediation" flag to make these use cases more explicit?

> I made https://github.com/fedidcg/FedCM/pull/368 to support multi IDP in general, as a prerequisite for this PR.

I don't know if we should submit this PR right now.

npm1:

spec/index.bs
0711261847: Payment

PR #356, opened 2022-09-27
Add fetch calls

npm1:
Relevant issues:
* https://github.com/fedidcg/FedCM/issues/261
* https://github.com/fedidcg/FedCM/issues/320

This PR adds the fetch calls to the other FedCM fetches. It is pretty noncontroversial but is missing the majority of request parameters. A future PR will address that.

***
<a href="https://pr-preview.s3.amazonaws.com/npm1/FedCM/pull/356.html" title="Last updated on Sep 30, 2022, 7:39 PM UTC (b930834)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/356/1572b7a...npm1:b930834.html" title="Last updated on Sep 30, 2022, 7:39 PM UTC (b930834)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1186,10 +1186,21 @@ To <dfn>fetch the accounts list</dfn> given a [=Manifest=] |manifest| and an {{I
         |manifest|[{{Manifest/accounts_endpoint}}] (the relative URL) and |provider|'s
         {{IdentityProviderConfig/configURL}} (the base URL).
     1. If |accountsUrl| is failure, return an empty list.
-    1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is
-        |accountsUrl| and [=request/redirect mode=] set to "error".
-    1. TODO: set other fields of the request, including the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]]
-        header or equivalent, as well as using [=Identity Provider=]'s cookies.
+    1. If |accountsUrl| is not [=same origin=] with |provider|'s
+        {{IdentityProviderConfig/configURL}}, return an empty list.
+    1. If |accountsUrl| is not a [=potentially trustworthy URL=], return an empty list.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |accountsUrl|
+        1. [=request/redirect mode=] set to "error"
+        1. [=request/client=] set to null
+        1. [=request/window=] set to "no-window"
+        1. [=request/service-workers mode=] set to "none"
+        1. [=request/destination=] set to "identity"
+        1. [=request/origin=] set to a unique [=opaque origin=]
+        1. [=request/header list=] set to a [=list=] containing a single [=header=] with
+            [=header/name=] set to "Accept" and [=header/value=] set to "JSON"
+        1. [=request/referrer policy=] set to "no-referrer"
+        1. [=request/credentials mode=] set to "include"
 
         Issue: The credentialed fetch in this algorithm can lead to a timing attack that leaks the user's identities before the user consents. This is an active area of investigation that is being explored [here](https://github.com/fedidcg/FedCM/issues/230#issuecomment-1089040953).
     1. Let |accountsList| be an empty list.
@@ -1255,10 +1266,21 @@ To <dfn noexport>fetch the client metadata</dfn> given a [=Manifest=] |manifest|
         |manifest|["{{Manifest/client_metadata_endpoint}}"] (the relative URL) and |provider|'s
         {{IdentityProviderConfig/configURL}} (the base URL).
     1. If |clientMetadataUrl| is failure, return null.
-    1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is
-        |clientMetadataUrl| and [=request/redirect mode=] set to "error".
-    1. TODO: set other fields of the request, including the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]]
-        header or equivalent.
+    1. If |clientMetadataUrl| is not a [=potentially trustworthy URL=], return null.
+    1. If |clientMetadataUrl| is not [=same origin=] with |provider|'s
+        {{IdentityProviderConfig/configURL}}, return null.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |clientMetadataUrl|
+        1. [=request/redirect mode=] set to "error"
+        1. [=request/client=] set to null
+        1. [=request/window=] set to "no-window"
+        1. [=request/service-workers mode=] set to "none"
+        1. [=request/destination=] set to "identity"
+        1. [=request/origin=] set to a unique [=opaque origin=]
+        1. [=request/header list=] set to a [=list=] containing a single [=header=] with
+            [=header/name=] set to "Accept" and [=header/value=] set to "JSON"
+        1. [=request/referrer=] set to [=RP=]'s URL (TODO).
+        1. [=request/credentials mode=] set to "omit"
     1. Let |clientMetadata| be a new [=ordered map=].
     1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the following
         steps given a <a spec=fetch for=/>response</a> |response|:
@@ -1300,16 +1322,28 @@ To <dfn>create tokens</dfn> given an [=AccountState=] |accountState|, a {{USVStr
     1. Let |tokenUrl| be the result of running [=url parser=] given
         |manifest|[{{Manifest/id_token_endpoint}}] (the relative URL) and |provider|'s
         {{IdentityProviderConfig/configURL}} (the base URL).
-    1. If |tokenUrl| is failure, return an empty list.
+    1. If |tokenUrl| is failure, return null.
+    1. If |tokenUrl| is not [=same origin=] with |provider|'s {{IdentityProviderConfig/configURL}},
+        return null.
+    1. If |tokenUrl| is not a [=potentially trustworthy URL=], return null.
     1. Let |requestBody| be a new object constructed as follows (TODO: what is the correct type):
         1. Set |requestBody|["account_id"] to |accountId|.
         1. Set |requestBody|["client_id"] to |provider|'s {{IdentityProviderConfig/clientId}}.
         1. Set |requestBody|["nonce"] to |provider|'s {{IdentityProviderConfig/nonce}}.
-    1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is
-        |tokenUrl|, [=request/redirect mode=] set to "error", [=request/mode=] set to `POST`, and
-        [=request/body=] set to |requestBody|.
-    1. TODO: set other fields of the request, including the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]]
-        header or equivalent, as well as using [=Identity Provider=]'s cookies.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |tokenUrl|
+        1. [=request/mode=] set to "POST"
+        1. [=request/body=] set to |requestBody|
+        1. [=request/redirect mode=] set to "error"
+        1. [=request/client=] set to null
+        1. [=request/window=] set to "no-window"
+        1. [=request/service-workers mode=] set to "none"
+        1. [=request/destination=] set to "identity"
+        1. [=request/origin=] set to a unique [=opaque origin=]
+        1. [=request/header list=] set to a [=list=] containing a single [=header=] with
+            [=header/name=] set to "Accept" and [=header/value=] set to "JSON"
+        1. [=request/referrer=] set to [=RP=]'s URL (TODO).
+        1. [=request/credentials mode=] set to "include"
     1. Let |token| be null.
     1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the following
         steps given a <a spec=fetch for=/>response</a> |response|:
@@ -1351,8 +1385,19 @@ whether the manifest is included in the manifest list:
     1. Set |rootUrl|'s [=url/scheme=] to |configUrl|'s [=url/scheme=].
     1. Set |rootUrl|'s [=url/host=] to |configUrl|'s [=url/host=]'s [=host/registrable domain=].
     1. Set |rootUrl|'s [=url/path=] to the <a>list</a> «".well-known", "web-identity"».
-    1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is |rootUrl| and
-        [=request/redirect mode=] set to "error". TODO: what other fields need to be set?
+    1. If |rootUrl| is not a [=potentially trustworthy URL=], return false.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |rootUrl|
+        1. [=request/redirect mode=] set to "error"
+        1. [=request/client=] set to null
+        1. [=request/window=] set to "no-window"
+        1. [=request/service-workers mode=] set to "none"
+        1. [=request/destination=] set to "identity"
+        1. [=request/origin=] set to a unique [=opaque origin=]
+        1. [=request/header list=] set to a [=list=] containing a single [=header=] with
+            [=header/name=] set to "Accept" and [=header/value=] set to "JSON"
+        1. [=request/referrer policy=] set to "no-referrer"
+        1. [=request/credentials mode=] set to "omit"
     1. Let |check| be false.
     1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the following
         steps given a <a spec=fetch for=/>response</a> |response|:
@@ -1379,14 +1424,23 @@ whether the manifest is included in the manifest list:
 The <dfn>fetch the internal manifest</dfn> algorithm accepts an {{IdentityProviderConfig}} |provider| and returns a [=Manifest=] or null if there is some failure fetching or parsing the manifest:
     1. Run a [[!CSP]] check with a [[CSP#directive-connect-src|connect-src]]
         directive on the URL passed as |provider|'s {{IdentityProviderConfig/configURL}}
-        and return if the check fails. TODO: Add parameters to integrate this check into fetch.
+        and return if the check fails.
     1. Let |url| be |provider|'s {{IdentityProviderConfig/configURL}}.
+    1. If |url| is not a [=potentially trustworthy URL=], return null.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |url|
+        1. [=request/redirect mode=] set to "error"
+        1. [=request/client=] set to null
+        1. [=request/window=] set to "no-window"
+        1. [=request/service-workers mode=] set to "none"
+        1. [=request/destination=] set to "identity"
+        1. [=request/origin=] set to a unique [=opaque origin=]
+        1. [=request/header list=] set to a [=list=] containing a single [=header=] with
+            [=header/name=] set to "Accept" and [=header/value=] set to "JSON"
+        1. [=request/referrer policy=] set to "no-referrer"
+        1. [=request/credentials mode=] set to "omit"
     1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is |url| and
         [=request/redirect mode=] set to "error".
-    1. TODO: set other fields of the request, including the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]]
-        header or equivalent.
-
-        Issue: it is unclear if the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]] header is [needed](https://github.com/fedidcg/FedCM/issues/267) on the manifest fetch.
     1. Let |manifest| be null.
     1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the following
         steps given a <a spec=fetch for=/>response</a> |response|:
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1348,14 +1348,13 @@ To <dfn>create tokens</dfn> given an [=AccountState=] |accountState|, a {{USVStr
     1. If |idTokenUrl| is failure, return null.
     1. If |idTokenUrl| is not [=same origin=] with |configUrl|, return null.
     1. If |idTokenUrl| is not a [=potentially trustworthy URL=], return null.
-    1. Let |requestBody| be a new object constructed as follows (TODO: what is the correct type):
-        1. Set |requestBody|["account_id"] to |accountId|.
-        1. Set |requestBody|["client_id"] to |provider|'s {{IdentityProviderConfig/clientId}}.
-        1. Set |requestBody|["nonce"] to |provider|'s {{IdentityProviderConfig/nonce}}.
+    1. Let |requestBody| be a [=string=] resulting in concatenating "client_id=",
+        |provider|'s {{IdentityProviderConfig/clientId}}, "&nonce=",
+        |provider|'s {{IdentityProviderConfig/nonce}}, "&account_id=", and |accountId|.
     1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
         1. [=request/url=] set to |idTokenUrl|
         1. [=request/mode=] set to "POST"
-        1. [=request/body=] set to |requestBody|
+        1. [=request/body=] set to the [=UTF-8 encode=] of |requestBody|
         1. [=request/redirect mode=] set to "error"
         1. [=request/client=] set to null
         1. [=request/window=] set to "no-window"
@@ -1459,7 +1458,6 @@ The <dfn>fetch the internal manifest</dfn> algorithm accepts an {{IdentityProvid
         steps given a <a spec=fetch for=/>response</a> |response|:
         1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
         1. If |json| is null, return.
-        1. If |json| is a parsing exception, or if |json| is not an [=ordered map=], return.
         1. If |json|["accounts_endpoint"] does not exist, or if it is not a [=string=], return.
         1. If |json|["client_metadata_endpoint"] does not exist, or if it is not a [=string=],
             return.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1168,8 +1168,9 @@ To <dfn>potentially create IdentityCredential</dfn>, given an {{IdentityProvider
 <div algorithm>
 To <dfn>compute account state</dfn> given an {{IdentityProviderConfig}} |provider| and an
     [=Account JSON=] |account|, run the following steps:
-    1. Let |idpURL| be |provider|'s {{IdentityProviderConfig/configURL}}.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1181,19 +1182,47 @@ To <dfn>compute account state</dfn> given an {{IdentityProvider}} |provider| and
...
+            1. If |account| is not an [=ordered map=], return.
+            1. If |account|["id"] does not exist or is not a [=string=], return.
+            1. If |account|["name"] does not exist or is not a [=string=], return.
+            1. If |account|["email"] does not exist or is not a [=string=], return.
+            1. Let |acc| be a new [=Account JSON=] with {{account_json/id}} set to |account|["id"],
+                {{account_json/name}} set to |account|["name"], and {{account_json/email}} set to
+                |account|["email"].
+            1. If |account|["given_name"] exists and is a [=string=], set |acc|'s
+                {{account_json/given_name}} to |account|["given_name"].
+            1. If |account|["approved_clients"] exists and is a list where each item is a
```
samuelgoto, line 1218: maybe approved_clients need to default to empty list by default (otherwise you are going to have to test for its presence later?)? i guess either way could work.

spec/index.bs
```diff
@@ -1222,13 +1251,30 @@ To <dfn>request consent to sign-up</dfn> the user with a given [=Account JSON=]
...
-        {{IdentityProvider/configURL}}.
-    1. Let the |clientMetadata| be the result of fetching the |clientMetadataEndpoint| with the
-        [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]] header but without the [=Identity Provider=]'s cookies.  This request MUST NOT follow [[RFC9110#header.location|HTTP redirects]] and instead return
-        null if there is any error. See also [[#idp-api-client-id-metadata-endpoint]].
-    1. Return |clientMetadata|.
+    1. Let |clientMetadataUrl| be the result of running [=url parser=] given
+        |manifest|["{{Manifest/client_metadata_endpoint}}"] (the relative URL) and |provider|'s
+        {{IdentityProvider/configURL}} (the base URL).
+    1. If |clientMetadataUrl| is failure, return null.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is
```
samuelgoto, line 1258: It is important that we capture here the fact in line 1129 that you are removing that NO cookies are passed to this request.

spec/index.bs
```diff
@@ -1248,16 +1294,36 @@ To <dfn>sign-in</dfn> the user with a given an [=AccountState=] |accountState|:
...
+        {{IdentityProvider/configURL}} (the base URL).
+    1. If |tokenUrl| is failure, return an empty list.
+    1. Let |requestBody| be a new object constructed as follows (TODO: what is the correct type):
+        1. Set |requestBody|["account_id"] to |accountId|.
+        1. Set |requestBody|["client_id"] to |provider|'s {{IdentityProvider/clientId}}.
+        1. Set |requestBody|["nonce"] to |provider|'s {{IdentityProvider/nonce}}.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is
+        |tokenUrl|, [=request/redirect mode=] set to "error", [=request/mode=] set to `POST`, and
+        [=request/body=] set to |requestBody|.
+    1. TODO: set other fields of the request, including the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]]
```
samuelgoto, line 1311: Again, here goes the user's IdP's cookies.

samuelgoto (commented):
Looks generally good to me, but we need to resolve TODOs before merging. Will take another pass when you resolve them.

spec/index.bs
```diff
@@ -1181,19 +1182,47 @@ To <dfn>compute account state</dfn> given an {{IdentityProviderConfig}} |provide
...
-        [=Identity Provider=]'s cookies.  This request MUST NOT follow
-        [[RFC9110#header.location|HTTP redirects]] and instead return null if there is any error.
-        See also [[#idp-api-accounts-endpoint]].
+    1. Let |accountsUrl| be the result of running [=url parser=] given
+        |manifest|[{{Manifest/accounts_endpoint}}] (the relative URL) and |provider|'s
+        {{IdentityProviderConfig/configURL}} (the base URL).
+    1. If |accountsUrl| is failure, return an empty list.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is
+        |accountsUrl| and [=request/redirect mode=] set to "error".
+    1. TODO: set other fields of the request, including the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]]
```
samuelgoto, line 1191: For all of these requests, can you make sure that we set the flags such that:

- we don't follow redirects and
- we don't get intercepted by service works and
- we fail when it is not HTTPs?

spec/index.bs
```diff
@@ -1181,19 +1182,47 @@ To <dfn>compute account state</dfn> given an {{IdentityProviderConfig}} |provide
...
 To <dfn>fetch the accounts list</dfn> given a [=Manifest=] |manifest| and an {{IdentityProviderConfig}}
     |provider|:
-    1. Let the |accountsEndpoint| url be the relative url
-        |manifest|["{{Manifest/accounts_endpoint}}"] of |provider|'s {{IdentityProviderConfig/configURL}}.
-    1. Let |accountsList| be the result of fetching the |accountsEndpoint| with the
-        [=Identity Provider=]'s cookies.  This request MUST NOT follow
-        [[RFC9110#header.location|HTTP redirects]] and instead return null if there is any error.
-        See also [[#idp-api-accounts-endpoint]].
+    1. Let |accountsUrl| be the result of running [=url parser=] given
+        |manifest|[{{Manifest/accounts_endpoint}}] (the relative URL) and |provider|'s
```
samuelgoto, line 1186: Can we make sure we check that all of the endpoints in the |manifest| are same-origin with the |manifest| itself?

npm1:
PTAL, addressed all of your comments. The pending thing is the referrer being handwavy (in the case we do want to send it), while I figure out how to set it properly.

spec/index.bs
```diff
@@ -1181,19 +1182,58 @@ To <dfn>compute account state</dfn> given an {{IdentityProviderConfig}} |provide
...
+        {{IdentityProviderConfig/configURL}} (the base URL).
```
samuelgoto, line 1187: Don't you need to check that configURL and accountsURL is same-origin? Here and elsewhere?

spec/index.bs
```diff
@@ -1181,19 +1182,58 @@ To <dfn>compute account state</dfn> given an {{IdentityProviderConfig}} |provide
...
 To <dfn>fetch the accounts list</dfn> given a [=Manifest=] |manifest| and an {{IdentityProviderConfig}}
     |provider|:
-    1. Let the |accountsEndpoint| url be the relative url
-        |manifest|["{{Manifest/accounts_endpoint}}"] of |provider|'s {{IdentityProviderConfig/configURL}}.
-    1. Let |accountsList| be the result of fetching the |accountsEndpoint| with the
-        [=Identity Provider=]'s cookies.  This request MUST NOT follow
-        [[RFC9110#header.location|HTTP redirects]] and instead return null if there is any error.
-        See also [[#idp-api-accounts-endpoint]].
+    1. Let |accountsUrl| be the result of running [=url parser=] given
+        |manifest|[{{Manifest/accounts_endpoint}}] (the relative URL) and |provider|'s
```
samuelgoto, line 1186: This isn't guaranteed to be a "relative URL" right? What happens if an absolute URL is passed?

spec/index.bs
```diff
@@ -1181,19 +1182,58 @@ To <dfn>compute account state</dfn> given an {{IdentityProviderConfig}} |provide
...
+    1. If |accountsUrl| is not [=same origin=] with |provider|'s
+        {{IdentityProviderConfig/configURL}}, return an empty list.
+    1. If |accountsUrl| is not a [=potentially trustworthy URL=], return an empty list.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |accountsUrl|
+        1. [=request/redirect mode=] set to "error"
+        1. [=request/client=] set to null
+        1. [=request/window=] set to "no-window"
+        1. [=request/service-workers mode=] set to "none"
+        1. [=request/destination=] set to "identity"
```
samuelgoto, line 1198:
```suggestion
        1. [=request/destination=] set to "web-identity"
```

"identity" is probably too broad/ambiguous (e.g. maybe it means "cookies"?)?

spec/index.bs
```diff
@@ -1181,19 +1182,58 @@ To <dfn>compute account state</dfn> given an {{IdentityProviderConfig}} |provide
...
+        1. [=request/origin=] set to a unique [=opaque origin=]
```
samuelgoto, line 1199: what does this do? is this for the cache partitioning?

spec/index.bs
samuelgoto: ah, nevermind, i see line #1189 now.

samuelgoto, line 1187: IdentityProviderConfig/configURL is also a "string". don't you need to parse it?

spec/index.bs
```diff
@@ -1222,13 +1262,41 @@ To <dfn>request consent to sign-up</dfn> the user with a given [=Account JSON=]
...
+    1. If |clientMetadataUrl| is not a [=potentially trustworthy URL=], return null.
+    1. If |clientMetadataUrl| is not [=same origin=] with |provider|'s
+        {{IdentityProviderConfig/configURL}}, return null.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |clientMetadataUrl|
+        1. [=request/redirect mode=] set to "error"
+        1. [=request/client=] set to null
+        1. [=request/window=] set to "no-window"
+        1. [=request/service-workers mode=] set to "none"
+        1. [=request/destination=] set to "identity"
```
samuelgoto, line 1278:
```suggestion
        1. [=request/destination=] set to "web-identity"
```

Same here, I think something narrower would be useful.

spec/index.bs
```diff
@@ -1248,16 +1316,48 @@ To <dfn>sign-in</dfn> the user with a given an [=AccountState=] |accountState|:
...
-        [[#idp-api-id-token-endpoint]].
-    1. Return |tokens|["token"].
+    1. Let |tokenUrl| be the result of running [=url parser=] given
+        |manifest|[{{Manifest/id_token_endpoint}}] (the relative URL) and |provider|'s
+        {{IdentityProviderConfig/configURL}} (the base URL).
+    1. If |tokenUrl| is failure, return null.
+    1. If |tokenUrl| is not [=same origin=] with |provider|'s {{IdentityProviderConfig/configURL}},
+        return null.
+    1. If |tokenUrl| is not a [=potentially trustworthy URL=], return null.
+    1. Let |requestBody| be a new object constructed as follows (TODO: what is the correct type):
```
samuelgoto, line 1329: Would that be `application/x-www-form-urlencoded`?

spec/index.bs
```diff
@@ -1248,16 +1316,48 @@ To <dfn>sign-in</dfn> the user with a given an [=AccountState=] |accountState|:
...
     1. Assert |accountState|'s {{AccountState/registration state}} is [=registered=].
     1. Assert |accountState|'s {{AccountState/allows logout}} is true.
-    1. Let |request| be a new object.
-        1. Set |request|["account_id"] to |accountId|.
-        1. Set |request|["client_id"] to |provider|'s {{IdentityProviderConfig/clientId}}.
-        1. Set |request|["nonce"] to |provider|'s {{IdentityProviderConfig/nonce}}.
-    1. Let |tokens| be the result of making a POST request described in
-        [[#idp-api-id-token-endpoint]].
-    1. Return |tokens|["token"].
+    1. Let |tokenUrl| be the result of running [=url parser=] given
```
samuelgoto, line 1322:
```suggestion
    1. Let |idTokenUrl| be the result of running [=url parser=] given
```

For consistency?

spec/index.bs
```diff
@@ -1248,16 +1316,48 @@ To <dfn>sign-in</dfn> the user with a given an [=AccountState=] |accountState|:
...
+        1. Set |requestBody|["nonce"] to |provider|'s {{IdentityProviderConfig/nonce}}.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |tokenUrl|
+        1. [=request/mode=] set to "POST"
+        1. [=request/body=] set to |requestBody|
+        1. [=request/redirect mode=] set to "error"
+        1. [=request/client=] set to null
+        1. [=request/window=] set to "no-window"
+        1. [=request/service-workers mode=] set to "none"
+        1. [=request/destination=] set to "identity"
```
samuelgoto, line 1341:
```suggestion
        1. [=request/destination=] set to "web-identity"
```

spec/index.bs
```diff
@@ -1248,16 +1316,48 @@ To <dfn>sign-in</dfn> the user with a given an [=AccountState=] |accountState|:
...
+        1. [=request/origin=] set to a unique [=opaque origin=]
+        1. [=request/header list=] set to a [=list=] containing a single [=header=] with
+            [=header/name=] set to "Accept" and [=header/value=] set to "JSON"
+        1. [=request/referrer=] set to [=RP=]'s URL (TODO).
+        1. [=request/credentials mode=] set to "include"
+    1. Let |token| be null.
+    1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the following
+        steps given a <a spec=fetch for=/>response</a> |response|:
+        1. If |response| is a [=network error=] or its [=response/status=] is not an [=ok status=],
```
samuelgoto, line 1350: This block shows up multiple times .. I'm wondering if we should create an algorithm and reuse it in the other algorithms. WDYT?

spec/index.bs
```diff
@@ -1280,13 +1380,24 @@ manipulation to fingerprint. See [[#manifest-fingerprinting]].
...
-    1. Set |rootUrl|'s [=url/host=] to |url|'s [=url/host=]'s [=host/registrable domain=].
+    1. Set |rootUrl|'s [=url/scheme=] to |configUrl|'s [=url/scheme=].
+    1. Set |rootUrl|'s [=url/host=] to |configUrl|'s [=url/host=]'s [=host/registrable domain=].
     1. Set |rootUrl|'s [=url/path=] to the <a>list</a> «".well-known", "web-identity"».
-    1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is |rootUrl| and
-        [=request/redirect mode=] set to "error". TODO: what other fields need to be set?
+    1. If |rootUrl| is not a [=potentially trustworthy URL=], return false.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |rootUrl|
+        1. [=request/redirect mode=] set to "error"
```
samuelgoto, line 1391: I think that following redirects here is OK ... e.g. it does currently redirect from "https://google.com/.well-known/web-identity" to "https://www.google.com/.well-known/web-identity", I think.

spec/index.bs
```diff
@@ -1309,18 +1420,48 @@ whether the manifest is included in the manifest list:
...
-        [=Identity Provider=]'s cookies. This request MUST NOT follow
-        [[RFC9110#header.location|HTTP redirects]] and instead throw an
-        error if there are any.
-
-        Issue: it is unclear if the [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]] header is [needed](https://github.com/fedidcg/FedCM/issues/267) on the manifest fetch.
+    1. Let |url| be |provider|'s {{IdentityProviderConfig/configURL}}.
+    1. If |url| is not a [=potentially trustworthy URL=], return null.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |url|
+        1. [=request/redirect mode=] set to "error"
```
samuelgoto, line 1432: I'm also wondering if following redirects here is also OK.

samuelgoto: Also, the client_metadata_endpoint too, right?

spec/index.bs
```diff
@@ -1309,18 +1420,48 @@ whether the manifest is included in the manifest list:
...
+        1. [=request/client=] set to null
+        1. [=request/window=] set to "no-window"
+        1. [=request/service-workers mode=] set to "none"
+        1. [=request/destination=] set to "identity"
```
samuelgoto, line 1436:
```suggestion
        1. [=request/destination=] set to "web-identity"
```

spec/index.bs
npm1: Done

spec/index.bs
npm1: Done (it was handwavy everywhere)

spec/index.bs
npm1: Cookies are a web thing too? Unclear how you can have a non-web destination in fetch, but sure I don't have strong opinions about the name.

spec/index.bs
npm1: It sets the origin of the request. The origin is 'user agent' but there is not such thing.

spec/index.bs
npm1: It is optional.

spec/index.bs
npm1: Not sure how to set the body for application/x-www-form-urlencoded, need to ask.

spec/index.bs

spec/index.bs

spec/index.bs
npm1: I think impl only allows the root one to redirect

npm1:
Ready for more review, PTAL.

spec/index.bs
samuelgoto: sgtm

spec/index.bs
samuelgoto: Yeah, only the root one. That works for me.

spec/index.bs
```diff
@@ -1181,19 +1182,58 @@ To <dfn>compute account state</dfn> given an {{IdentityProviderConfig}} |provide
...
-    1. Let |accountsList| be the result of fetching the |accountsEndpoint| with the
-        [=Identity Provider=]'s cookies.  This request MUST NOT follow
-        [[RFC9110#header.location|HTTP redirects]] and instead return null if there is any error.
-        See also [[#idp-api-accounts-endpoint]].
+    1. Let |accountsUrl| be the result of running [=url parser=] given
+        |manifest|[{{Manifest/accounts_endpoint}}] (the relative URL) and |provider|'s
+        {{IdentityProviderConfig/configURL}} (the base URL).
+    1. If |accountsUrl| is failure, return an empty list.
+    1. If |accountsUrl| is not [=same origin=] with |provider|'s
+        {{IdentityProviderConfig/configURL}}, return an empty list.
```
samuelgoto: I meant here in this line, configURL is a string, so you can't test if it is [=same origin=] with |accountsUrl| unless you [=url parser=]?

Here and elsewhere where you are using configURL in a way that is not a string.

spec/index.bs
samuelgoto: Is this supposed to be referring to this [1]? [=opaque origin=] serializes to "null" rather than "user agent" right [2]?

Do you know what this is for?

[1] https://fetch.spec.whatwg.org/#concept-request-origin
[2] https://html.spec.whatwg.org/multipage/origin.html#concept-origin-opaque

spec/index.bs
```diff
@@ -1222,13 +1262,41 @@ To <dfn>request consent to sign-up</dfn> the user with a given [=Account JSON=]
...
-        {{IdentityProviderConfig/configURL}}.
-    1. Let the |clientMetadata| be the result of fetching the |clientMetadataEndpoint| with the
-        [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]] header but without the [=Identity Provider=]'s cookies.  This request MUST NOT follow [[RFC9110#header.location|HTTP redirects]] and instead return
-        null if there is any error. See also [[#idp-api-client-id-metadata-endpoint]].
-    1. Return |clientMetadata|.
+    1. Let |clientMetadataUrl| be the result of running [=url parser=] given
+        |manifest|["{{Manifest/client_metadata_endpoint}}"] (the relative URL) and |provider|'s
+        {{IdentityProviderConfig/configURL}} (the base URL).
+    1. If |clientMetadataUrl| is failure, return null.
+    1. If |clientMetadataUrl| is not a [=potentially trustworthy URL=], return null.
```
samuelgoto, line 1291: ah, neat, this captures "localhost": https://w3c.github.io/webappsec-secure-contexts/#potentially-trustworthy-origin

spec/index.bs
```diff
@@ -1248,16 +1316,48 @@ To <dfn>sign-in</dfn> the user with a given an [=AccountState=] |accountState|:
...
+        return null.
+    1. If |tokenUrl| is not a [=potentially trustworthy URL=], return null.
+    1. Let |requestBody| be a new object constructed as follows (TODO: what is the correct type):
+        1. Set |requestBody|["account_id"] to |accountId|.
+        1. Set |requestBody|["client_id"] to |provider|'s {{IdentityProviderConfig/clientId}}.
+        1. Set |requestBody|["nonce"] to |provider|'s {{IdentityProviderConfig/nonce}}.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |tokenUrl|
+        1. [=request/mode=] set to "POST"
+        1. [=request/body=] set to |requestBody|
```
samuelgoto: This seems a bit off to me, because it seems that a [=request/body=] has to be one of the following: a stream, a source or a length [1], and here |requestBody| is neither of those.

https://fetch.spec.whatwg.org/#concept-body
https://fetch.spec.whatwg.org/#concept-request-body

spec/index.bs
samuelgoto: Still unresolved?

spec/index.bs
samuelgoto: still unresolved?

spec/index.bs
```diff
@@ -1280,13 +1380,24 @@ manipulation to fingerprint. See [[#manifest-fingerprinting]].
...
-    1. Let |request| be a new <a spec=fetch for=/>request</a> whose [=request/url=] is |rootUrl| and
-        [=request/redirect mode=] set to "error". TODO: what other fields need to be set?
+    1. If |rootUrl| is not a [=potentially trustworthy URL=], return false.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |rootUrl|
+        1. [=request/redirect mode=] set to "error"
+        1. [=request/client=] set to null
+        1. [=request/window=] set to "no-window"
+        1. [=request/service-workers mode=] set to "none"
+        1. [=request/destination=] set to "identity"
```
samuelgoto:
```suggestion
        1. [=request/destination=] set to "web-identity"
```

spec/index.bs
```diff
@@ -1248,16 +1316,48 @@ To <dfn>sign-in</dfn> the user with a given an [=AccountState=] |accountState|:
...
+        1. [=request/mode=] set to "POST"
+        1. [=request/body=] set to |requestBody|
+        1. [=request/redirect mode=] set to "error"
+        1. [=request/client=] set to null
+        1. [=request/window=] set to "no-window"
+        1. [=request/service-workers mode=] set to "none"
+        1. [=request/destination=] set to "identity"
+        1. [=request/origin=] set to a unique [=opaque origin=]
+        1. [=request/header list=] set to a [=list=] containing a single [=header=] with
+            [=header/name=] set to "Accept" and [=header/value=] set to "JSON"
```
samuelgoto:
```suggestion
            [=header/name=] set to "Accept" and [=header/value=] set to [=JSON MIME Type=]
```

Shouldn't this be a mime type?

samuelgoto (commented):
Looks generally good to me, but still feels like the parameters aren't quite right.

On a separate note, any chance we can get @domfarolino to take a pass at this too?

spec/index.bs
npm1: I'm confused. What's unresolved? Are you looking at the latest version?

npm1:
> Looks generally good to me, but still feels like the parameters aren't quite right.
> 
> On a separate note, any chance we can get @domfarolino to take a pass at this too?

Can you look again but looking at the latest version??

npm1:
Added `Sec-` header changes

spec/index.bs
samuelgoto: If we don't know what this is for, why are we setting it? Is that a required field? Can we just leave it unset for now?

spec/index.bs
```diff
@@ -1181,19 +1183,72 @@ To <dfn>compute account state</dfn> given an {{IdentityProviderConfig}} |provide
...
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |accountsUrl|
+        1. [=request/redirect mode=] set to "error"
+        1. [=request/client=] set to null
+        1. [=request/window=] set to "no-window"
+        1. [=request/service-workers mode=] set to "none"
+        1. [=request/destination=] set to "web-identity"
+        1. [=request/origin=] set to a unique [=opaque origin=]
+        1. [=request/header list=] set to a [=list=] containing a single [=header=] with
+            [=header/name=] set to "Accept" and [=header/value=] set to "JSON"
```
samuelgoto, line 1200: Again, shouldn't this be a mime type? Are you sure this isn't supposed to be text/json?

spec/index.bs
```diff
@@ -1222,13 +1277,41 @@ To <dfn>request consent to sign-up</dfn> the user with a given [=Account JSON=]
...
-    1. Let the |clientMetadata| be the result of fetching the |clientMetadataEndpoint| with the
-        [[#Sec-FedCM-CSRF|Sec-FedCM-CSRF]] header but without the [=Identity Provider=]'s cookies.  This request MUST NOT follow [[RFC9110#header.location|HTTP redirects]] and instead return
-        null if there is any error. See also [[#idp-api-client-id-metadata-endpoint]].
-    1. Return |clientMetadata|.
+    1. Let |clientMetadataUrl| be the result of running [=url parser=] given
+        |manifest|["{{Manifest/client_metadata_endpoint}}"].
+    1. If |clientMetadataUrl| is failure, let |clientMetadataUrl| be the result of running
+        [=url parser=] given |manifest|["{{Manifest/client_metadata_endpoint}}"] (the relative URL) and |configUrl| (the base URL).
+    
+    Note: This means the we allow passing the client metadata endpoint as either an absolute or
```
samuelgoto, line 1278: This block is also going to be repeated 5 times (for each of the endpoints), so I think it is worth creating an algorithm to build these URLs form the baseURL (and check for things like same-site-ness).

npm1: It is 3 times, but yea worth doing the helper. Done

spec/index.bs
npm1: I do not think so, because Fetch does not handle that properly. `If request’s origin is "client", then set request’s origin to request’s client’s origin.`
https://fetch.spec.whatwg.org/#ref-for-concept-request-origin%E2%91%A6:~:text=If%20request%E2%80%99s%20origin%20is%20%22client%22%2C%20then%20set%20request%E2%80%99s%20origin%20to%20request%E2%80%99s%20client%E2%80%99s%20origin.

spec/index.bs
npm1: Again, don't think so? MIME type is an object computed from headers[1] but header names and values are byte sequences [2]. I think it should be `application/json` based on the implementation.
[1] https://mimesniff.spec.whatwg.org/#mime-type
[2] https://fetch.spec.whatwg.org/#concept-header

npm1 (commented):
Thanks for the review, PTAL.

samuelgoto:
LGTM

I'll merge, and if there is anything pending we can patch separately.

PR #361, opened 2022-10-07
spec polish

yi-gu:
- fixes some missing changes from other PRs
- uses permission instead of consent
- minor typo fixes

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Oct 8, 2022, 12:15 AM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [CSS Spec Preprocessor](https://api.csswg.org/bikeshed/) - CSS Spec Preprocessor is the web service used to build Bikeshed specs.

:link: [Related URL](https://api.csswg.org/bikeshed/?url=https%3A%2F%2Fraw.githubusercontent.com%2Fyi-gu%2FFedCM%2Fe74fa53dfc10943e1a411e83b5ff39d381e3a1f7%2Fspec%2Findex.bs&force=1&md-warning=not%20ready)

```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>500 Internal Server Error</title>
</head><body>
<h1>Internal Server Error</h1>
<p>The server encountered an internal error or
misconfiguration and was unable to complete
your request.</p>
<p>Please contact the server administrator at 
 [no address given] to inform them of the time this error occurred,
 and the actions you performed just before this error.</p>
<p>More information about this error may be available
in the server error log.</p>
<hr>
<address>Apache/2.4.10 (Debian) Server at api.csswg.org Port 443</address>
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20fedidcg/FedCM%23361.)._
</details>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -224,7 +224,7 @@ sign-out system the [=Identity Provider=] will log the user out of all
 [=Relying Parties=] the user has signed into along with logging the user
 out of the [=Identity Provider=] itself. Upon returning to any associated
 [=Relying Party=], the user will need to sign in with the [=Identity Provider=]
-first and go through the [[#use-cases-sign-in]] flow on [=Relying Party=].
+first and go through the [[#use-cases-sign-in]] flow on the [=Relying Party=].
 
 <!-- ============================================================ -->
 ### Revocation ### {#use-cases-revocation}
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -223,14 +223,14 @@ sign-out system provided by the [=Identity Provider=]. After using the
 sign-out system the [=Identity Provider=] will log the user out of all
 [=Relying Parties=] the user has signed into along with logging the user
 out of the [=Identity Provider=] itself. Upon returning to any associated
-[=Relying Party=], or the [=Identity Provider=], the user will have to
-go through the [[#use-cases-sign-in]] flow.
+[=Relying Party=], the user will need to sign in with the [=Identity Provider=]
+first and go through the [[#use-cases-sign-in]] flow on [=Relying Party=].
 
 <!-- ============================================================ -->
 ### Revocation ### {#use-cases-revocation}
 <!-- ============================================================ -->
 
-After a user has created an account at a [=Relying Party=] there are two
+After a user has created an account on a [=Relying Party=] there are two
 ways a user can cancel their account with the [=Relying Party=]:
 
 <!-- ============================================================ -->
@@ -959,7 +959,7 @@ the following properties:
 
 <dl dfn-type="argument" dfn-for="AccountState">
     :   <dfn>registration state</dfn>
-    ::  Keeps track of whether the user agent is aware that the user has registered an account at the
+    ::  Keeps track of whether the user agent is aware that the user has registered an account on the
         [=RP=] with the [=IDP=] or not.
          Can be [=registered=] or [=unregistered=] (by default).
     :   <dfn>allows logout</dfn>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -117,7 +117,7 @@ it depends on third-party cookies, from sign-in to sign-out and revocation.
 
 In order to provide the federated identity primitives without the use of
 third-party cookies the API places the [=user agent=] as a mediator between
-[=RPs=] and [=IDPs=]. This mediation requires user consent before permitting
+[=RPs=] and [=IDPs=]. This mediation requires user permission before allowing
 the [=RPs=] and [=IDPs=] to know about their connection to the user.
 
 <script type="text/typogram">
@@ -168,9 +168,9 @@ A Sign-up occurs when the user is registering a new account at the
 For instance, a user navigates to a [=Relying Party=] in their browser
 and creates an account. The [=Relying Party=] displays supported
 [=Identity Providers=] to the user who selects their favorite. The user
-is prompted "Do you want to create an account with the [=Relying Party=]?".
-Upon user agreement an [=account=] is created with the [=Relying Party=] and
-the user has a [=session=] initialized.
+is prompted "Do you want to create an account with the [=Identity Provider=]?".
+Upon user agreement an [=account=] is created with the [=Identity Provider=] and
+the user has a [=session=] initialized on [=Relying Party=].
 
 <!-- ============================================================ -->
 ### Sign-in ### {#use-cases-sign-in}
@@ -179,21 +179,6 @@ the user has a [=session=] initialized.
 After a user navigates to a [=Relying Party=] in a browser and decides to
 create an account by going through their [[#use-cases-sign-up]] flow,
 there are two ways a user logs into their [=account=] once their [=session=] expires:
-
-<!-- ============================================================ -->
-#### Auto Sign-in #### {#use-cases-auto-sign-in}
-<!-- ============================================================ -->
-
-Auto Sign-in occurs when the [=Identity Provider=] has already gathered
-permissions from the user to share their identity with the [=Relying Party=]
-and automatically signs the user in.
-
-For example, the user has previously executed the [[#use-cases-sign-up]] flow
-and then changes from their phone to their laptop. On the new device the
-user goes to the [=Relying Party=] and selects to sign-in using their
-[=Identity Provider=]. The [=Identity Provider=] knows, and proves, the user
-has signed up to the [=Relying Party=] and the [=Relying Party=] creates a
-new [=session=] for the users [=account=].
 
 <!-- ============================================================ -->
 #### Explicit Sign-in #### {#use-cases-explicit-sign-in}
@@ -211,9 +196,24 @@ For example, after the user has done the [[#use-cases-sign-out]] flow of the
     * the user already has an account with the [=Relying Party=].
     * the user has logged out of the [=Relying Party=].
 
-The user is then prompted, "Do you want to sign-in with the [=Relying Party=]?"
+The user is then prompted, "Do you want to sign-in with the [=Identity Provider=]?"
 and upon user agreement the [=Relying Party=] creates a new [=session=] with
 the users existing [=account=].
+
+<!-- ============================================================ -->
+#### Auto Sign-in #### {#use-cases-auto-sign-in}
+<!-- ============================================================ -->
+
+Auto Sign-in occurs when the [=Identity Provider=] has already gathered
+permissions from the user to share their identity with the [=Relying Party=]
+and automatically signs the user in.
+
+For example, the user has previously executed the [[#use-cases-sign-up]] flow
+and then changes from their phone to their laptop. On the new device the
+user goes to the [=Relying Party=] and selects to sign-in using their
+[=Identity Provider=]. The [=Identity Provider=] knows, and proves, the user
+has signed up to the [=Relying Party=] and the [=Relying Party=] creates a
+new [=session=] for the users [=account=].
 
 <!-- ============================================================ -->
 ### Sign-out ### {#use-cases-sign-out}
@@ -249,7 +249,7 @@ go through the [[#use-cases-explicit-sign-in]] flow.
 ### Revocation ### {#use-cases-revocation}
 <!-- ============================================================ -->
 
-After a user has created an account with a [=Relying Party=] there are two
+After a user has created an account on a [=Relying Party=] there are two
 ways a user can cancel their account with the [=Relying Party=]:
 
 <!-- ============================================================ -->
@@ -257,7 +257,7 @@ ways a user can cancel their account with the [=Relying Party=]:
 <!-- ============================================================ -->
 
 The user can delete their account through the [=Relying Party=] by using
-the provided cancel account system. The [=Relying Party=] informs the
+the provided account revocation system. The [=Relying Party=] informs the
 [=Identity Provider=] that the user has deleted (revoked) their account.
 When the user returns to the [=Relying Party=] they will need to complete
 the [[#use-cases-sign-up]] flow in order to access the site.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -533,20 +532,20 @@ identity federation.
       |                               |                                 |   "branding: ...,"
       |                               |                                 | "}"
       |                               |                                 |
-      |                               | "GET client_metadata.php"       |
```
samuelgoto, line 536: Are you just switching the orders here because that's how we actually implement it?

Another separate point: can you make the GETs for the manifest and the well-known file to happen in parallel? e.g. make two "requests" and then let the "responses" come back afterwards?

spec/index.bs
```diff
@@ -979,8 +978,8 @@ the following properties:
 
 <dl dfn-type="argument" dfn-for="AccountState">
     :   <dfn>registration state</dfn>
-    ::  Keeps track of whether the user agent is aware that the user has registered an account in
-         the [=RP=] or not.
+    ::  Keeps track of whether the user agent is aware that the user has registered an account on the
```
samuelgoto, line 981:
```suggestion
    ::  Keeps track of whether the user agent is aware that the user has registered an account at the
```

spec/index.bs
```diff
@@ -1704,13 +1703,13 @@ network requests performed:
 ┌──────────────────────────┬─────────┬───────────┬──────────┐
 │ Endpoint                 │ cookies │ client_id │ referrer │
 ├──────────────────────────┼─────────┼───────────┼──────────┤
-│ manifest                 │  no     │  no       │  no      │
+│ manifests                 │  no     │  no       │  no      │
```
samuelgoto, line 1706: I think we should try to call these things differently. I think calling them both "manifests" is causing a big confusion.

samuelgoto (commented):
Thanks for taking a pass with a fresh pair of eyes! Very helpful! Comments inline!

spec/index.bs
```diff
@@ -168,9 +168,9 @@ A Sign-up occurs when the user is registering a new account at the
 For instance, a user navigates to a [=Relying Party=] in their browser
 and creates an account. The [=Relying Party=] displays supported
 [=Identity Providers=] to the user who selects their favorite. The user
-is prompted "Do you want to create an account with the [=Relying Party=]?".
-Upon user agreement an [=account=] is created with the [=Relying Party=] and
-the user has a [=session=] initialized.
+is prompted "Do you want to create an account with the [=Identity Provider=]?".
```
npm1, line 171: I don't think this change is accurate. The user is creating an account in the RP. Maybe 'create an account in RP using credentials from IDP' if you think it's confusing?

spec/index.bs
```diff
@@ -211,10 +196,25 @@ For example, after the user has done the [[#use-cases-sign-out]] flow of the
     * the user already has an account with the [=Relying Party=].
     * the user has logged out of the [=Relying Party=].
 
-The user is then prompted, "Do you want to sign-in with the [=Relying Party=]?"
+The user is then prompted, "Do you want to sign-in with the [=Identity Provider=]?"
```
npm1, line 199: ditto

spec/index.bs
```diff
@@ -1909,7 +1908,7 @@ While the [=push model=] seems like an improvement for privacy, the current API
 ##### The IdP Sign-in Status API ##### {#the-idp-sign-in-status-api}
 <!-- ============================================================ -->
 
-Issue: this is a more formal description of what is proposed [here](https://github.com/fedidcg/FedCM/issues/230#issuecomment-1233290946).
+Issue: there is a more formal description of what is proposed [here](https://github.com/fedidcg/FedCM/issues/230#issuecomment-1233290946).
```
npm1, line 1911: This changes the meaning of the sentence. What is more formal, the comment or the proposal described in this section?

spec/index.bs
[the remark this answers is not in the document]
yi-gu: Yes. Will update. Although it's not "with IDP's account" per se (it's a new account created on the RP site).

spec/index.bs
```diff
@@ -211,10 +196,25 @@ For example, after the user has done the [[#use-cases-sign-out]] flow of the
     * the user already has an account with the [=Relying Party=].
     * the user has logged out of the [=Relying Party=].
 
-The user is then prompted, "Do you want to sign-in with the [=Relying Party=]?"
+The user is then prompted, "Do you want to sign-in with the [=Identity Provider=]?"
 and upon user agreement the [=Relying Party=] creates a new [=session=] with
 the users existing [=account=].
 
+<!-- ============================================================ -->
```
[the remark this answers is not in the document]
yi-gu: That makes sense. The expected behavior is still in discussion and we can always add it back when it's stabilized. Will remove.

spec/index.bs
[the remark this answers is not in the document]
samuelgoto: Maybe we should use the language that we use in the implementation? "Sign-in to [=Relying Party=] with [=Identity Provider=]"?

spec/index.bs
```diff
@@ -249,15 +249,15 @@ go through the [[#use-cases-explicit-sign-in]] flow.
 ### Revocation ### {#use-cases-revocation}
 <!-- ============================================================ -->
 
-After a user has created an account with a [=Relying Party=] there are two
+After a user has created an account on a [=Relying Party=] there are two
```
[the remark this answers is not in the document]
yi-gu: Done.

spec/index.bs
yi-gu: Yes just switched the order. It's also how it's spec'd below.

re fetching in parallel: I think so. That's how they are spec'd and implemented today. Note that there are "two" responses for those "two" requests. Combining two responses into one may give the server better chance to conduct the manifest attack, right?

spec/index.bs
yi-gu: Right. It's tracked in a separate issue for renaming the two configuration files.

spec/index.bs
yi-gu: Sam suggested to mention both parties so updated.

spec/index.bs
yi-gu: reverted.

spec/index.bs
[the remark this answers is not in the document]
yi-gu: The title in the implementation does not differentiate sign-up and sign-in. In addition, from spec's perspective the current question "Do you want to" is more readable to me. WDYT?

spec/index.bs
```diff
@@ -168,23 +168,19 @@ A Sign-up occurs when the user is registering a new account at the
...
 the user has a [=session=] initialized on [=Relying Party=].
 
 <!-- ============================================================ -->
 ### Sign-in ### {#use-cases-sign-in}
 <!-- ============================================================ -->
 
-After a user navigates to a [=Relying Party=] in a browser and decides to
-create an account by going through their [[#use-cases-sign-up]] flow,
-there are two ways a user logs into their [=account=] once their [=session=] expires:
+After a user navigates to a [=Relying Party=] in a browser and create an
```
npm1, line 179:
```suggestion
After a user navigates to a [=Relying Party=] in a browser and creates an
```

spec/index.bs
```diff
@@ -243,13 +224,13 @@ sign-out system the [=Identity Provider=] will log the user out of all
 [=Relying Parties=] the user has signed into along with logging the user
 out of the [=Identity Provider=] itself. Upon returning to any associated
```
npm1, line 225: This last sentence is not quite correct because the user needs to first login to the IDP again before FedCM is shown when they visit the RP. Maybe make it a bit clearer?

spec/index.bs
```diff
@@ -243,13 +224,13 @@ sign-out system the [=Identity Provider=] will log the user out of all
...
 [=Relying Party=], or the [=Identity Provider=], the user will have to
-go through the [[#use-cases-explicit-sign-in]] flow.
+go through the [[#use-cases-sign-in]] flow.
 
 <!-- ============================================================ -->
 ### Revocation ### {#use-cases-revocation}
 <!-- ============================================================ -->
 
-After a user has created an account on a [=Relying Party=] there are two
+After a user has created an account at a [=Relying Party=] there are two
```
npm1, line 233: I do not think 'at' is correct. Maybe change to 'in' if you are unhappy with 'on'? Example: https://support.microsoft.com/en-us/windows/create-a-local-user-or-administrator-account-in-windows-20de74e0-ac7f-3502-a866-32915af2a34d

spec/index.bs
```diff
@@ -978,7 +959,7 @@ the following properties:
 
 <dl dfn-type="argument" dfn-for="AccountState">
     :   <dfn>registration state</dfn>
-    ::  Keeps track of whether the user agent is aware that the user has registered an account on the
+    ::  Keeps track of whether the user agent is aware that the user has registered an account at the
```
npm1, line 962: ditto

spec/index.bs
yi-gu: "on" seems right: https://support.google.com/youtube/answer/161805?hl=en&co=GENIE.Platform%3DDesktop

npm1: I'm fine with either on or in, but not at.

spec/index.bs
yi-gu: Updated. PTAL.

spec/index.bs
```diff
@@ -168,38 +168,19 @@ A Sign-up occurs when the user is registering a new account at the
 For instance, a user navigates to a [=Relying Party=] in their browser
 and creates an account. The [=Relying Party=] displays supported
 [=Identity Providers=] to the user who selects their favorite. The user
-is prompted "Do you want to create an account with the [=Relying Party=]?".
-Upon user agreement an [=account=] is created with the [=Relying Party=] and
-the user has a [=session=] initialized.
+is prompted "Do you want to sign up on the [=Relying Party=] with the [=Identity Provider=]?".
```
samuelgoto, line 171:
```suggestion
is prompted "Do you want to sign-up to the [=Relying Party=] with [=Identity Provider=]?".
```

spec/index.bs
```diff
@@ -211,7 +192,7 @@ For example, after the user has done the [[#use-cases-sign-out]] flow of the
     * the user already has an account with the [=Relying Party=].
     * the user has logged out of the [=Relying Party=].
 
-The user is then prompted, "Do you want to sign-in with the [=Relying Party=]?"
+The user is then prompted, "Do you want to sign in on the [=Relying Party=] with the [=Identity Provider=]?"
```
samuelgoto, line 195:
```suggestion
The user is then prompted, "Do you want to sign-in to [=Relying Party=] with [=Identity Provider=]?"
```

spec/index.bs
samuelgoto: > Combining two responses into one may give the server better chance to conduct the manifest attack, right?

In what way?

yi-gu: tracker could inject RP information in configUrl. Since there's only one response from the server for the two requests, it is relatively easier for the tracker to dynamically respond to the well-known file request with the RP specific url in `provider_urls`?

spec/index.bs
yi-gu: I have been treating "sign-up" (with hyphen) as a noun instead of a verb. Does it make sense?

spec/index.bs
```diff
@@ -223,14 +223,14 @@ sign-out system provided by the [=Identity Provider=]. After using the
 sign-out system the [=Identity Provider=] will log the user out of all
 [=Relying Parties=] the user has signed into along with logging the user
 out of the [=Identity Provider=] itself. Upon returning to any associated
-[=Relying Party=], or the [=Identity Provider=], the user will have to
-go through the [[#use-cases-sign-in]] flow.
+[=Relying Party=], the user will need to sign in with the [=Identity Provider=]
+first and go through the [[#use-cases-sign-in]] flow on [=Relying Party=].
```
npm1, line 227:
```suggestion
first and go through the [[#use-cases-sign-in]] flow on the [=Relying Party=].
```

npm1 (approved):

spec/index.bs
samuelgoto, line 171:
```suggestion
is prompted "Do you want to sign up to the [=Relying Party=] with the [=Identity Provider=]?".
```

spec/index.bs
```diff
@@ -168,38 +168,19 @@ A Sign-up occurs when the user is registering a new account at the
 For instance, a user navigates to a [=Relying Party=] in their browser
 and creates an account. The [=Relying Party=] displays supported
 [=Identity Providers=] to the user who selects their favorite. The user
-is prompted "Do you want to create an account with the [=Relying Party=]?".
-Upon user agreement an [=account=] is created with the [=Relying Party=] and
-the user has a [=session=] initialized.
+is prompted "Do you want to sign up on the [=Relying Party=] with the [=Identity Provider=]?".
+Upon user agreement an [=account=] is created with the [=Identity Provider=] and
+the user has a [=session=] initialized on [=Relying Party=].
```
samuelgoto, line 173:
```suggestion
the user has a [=session=] initialized at the [=Relying Party=].
```

spec/index.bs
samuelgoto, line 195:
```suggestion
The user is then prompted, "Do you want to sign in to the [=Relying Party=] with the [=Identity Provider=]?"
```

spec/index.bs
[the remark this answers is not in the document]
samuelgoto: still "on"?

spec/index.bs
samuelgoto: ah, fair.

spec/index.bs
samuelgoto: still "on"?

spec/index.bs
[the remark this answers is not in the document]
yi-gu: Oh I saw you suggested "on" in another thread so I thought you have seen my mention about this example: https://support.google.com/youtube/answer/161805?hl=en&co=GENIE.Platform%3DDesktop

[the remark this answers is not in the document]
samuelgoto: ah, that makes sense.

samuelgoto (approved):

PR #368, opened 2022-10-19
Allow multiple IDPs in the provider array

cbiesinger:
***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/368.html" title="Last updated on Oct 19, 2022, 7:41 PM UTC (36c02ce)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/368/cf1e847...cbiesinger:36c02ce.html" title="Last updated on Oct 19, 2022, 7:41 PM UTC (36c02ce)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1083,32 +1083,34 @@ requests.
     algorithm is invoked, the user agent MUST execute the following steps:
 
       1.  Assert: |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"] [=map/exists=].
-      1.  Assert: |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"] [=list/size=] is 1.
-
-           Issue: Support choosing accounts from multiple [=Identity Provider=]s, as described [here](https://github.com/fedidcg/FedCM/issues/319).
       1. Run {{setTimeout}} passing a [=task=] which throws a {{NetworkError}}, after a timeout of
           60 seconds.
 
           Note: the purpose of having a timer here is to avoid leaking the reason causing this
           method to return null. If there was no such timer, the developer could easily infer
           whether the user has an account with the [=IDP=] or not, or whether the user closed the UI without granting permission to share the [=IDP=] account information with the [=RP=].
-      1. Let |provider| be |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"][0].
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1083,32 +1083,34 @@ requests.
...
 <div algorithm>
-To <dfn>potentially create IdentityCredential</dfn>, given an {{IdentityProviderConfig}} |provider|:
-    1. Let |manifest| be the result of running the [=fetch the manifest=]
-        algorithm with |provider|.
-    1. If |manifest| is null, return null.
-    1. Let |accountsList| be the result of running the
-        [=fetch the accounts list=] algorithm with |manifest| and |provider|.
+To <dfn>potentially create IdentityCredential</dfn>, given a list of {{IdentityProviderConfig}} |providers|:
+    1. Let |accountsList| be an empty list.
+    1. For each |provider| in |providers|:
```
samuelgoto, line 1102:
```suggestion
    1. In parallel, for each |provider| in |providers|:
```

We could/should do these in parallel right?

spec/index.bs
```diff
@@ -1083,32 +1083,34 @@ requests.
...
+        1. Let |manifest| be the result of running the [=fetch the manifest=]
+            algorithm with |provider|.
+        1. If |manifest| is null, return null.
+        1. Let |providerAccountsList| be the result of running the
+            [=fetch the accounts list=] algorithm with |manifest| and |provider|.
```
samuelgoto, line 1107: Didn't @bvandersloot-mozilla say that he was worried about enabling fetching multiple identity providers before we factored into the IdP Sign-in Status API (because, IIRC, that would allow trackers to be introduced here in this list and fail deliberately invisibly)?

npm1: Maybe we can add an "Issue" here to note that.

npm1 (commented):
Looks good to me modulo Sam's comments

samuelgoto:
Can we try to update this PR to contain the latest proposal for the Multiple IdP API @cbiesinger @npm1 and @tttzach ? A PR may be a useful/concrete way that TAG reviews / Mozilla / Webkit can look at to give guidance on (we already have the explainer, which is pretty good, but a spec PR may make things more concrete).

npm1:
This is kinda outdated but relevant. Do we plan to update this PR or should we close and later send a separate new PR?

cbiesinger:
I think we can keep this closed and we can pick it up later, either from this PR or start over

PR #378, opened 2022-11-22
Queue tasks from in parallel

npm1:
Since DiscoverFromExternalSource is run in parallel, this PR does some changes:

1. Queue tasks when we need to fetch or create a new Credential object.
2. Add 'on' methods which receive the fetch result and continue the execution of the FedCM algorithm with the results from the fetch.

Relevant issue: https://github.com/fedidcg/FedCM/issues/261

***
<a href="https://pr-preview.s3.amazonaws.com/npm1/FedCM/pull/378.html" title="Last updated on Jan 12, 2023, 6:40 PM UTC (080871e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/378/51aced8...npm1:080871e.html" title="Last updated on Jan 12, 2023, 6:40 PM UTC (080871e)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1220,18 +1220,14 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
 To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
 |provider| and a |globalObject|, run the following steps. This returns an {{IdentityCredential}}.
     1. Assert: These steps are running [=in parallel=].
-    1. Let |check|, |config| be variables initially null.
-    1.  [=Queue a global task=] on the [=DOM manipulation task source=]  given |globalObject| and
-        run the following steps:
-        1. Set |check| to the result of [=check the well-known file=] with |provider|.
-        1. Set |config| to the result of [=fetch the config file=] with |provider|.
-    1. Wait for |check| and |config| to become non-null.
-    1. Assert: |check| is true and |config| is an {{IdentityProviderAPIConfig}}.
-    1. Let |accountsList| be null.
+    1. [=Queue a global task=] on the [=DOM manipulation task source=]  given |globalObject| to
+        set |check| to the result of [=check the well-known file=] with |provider|.
+    1. [=Queue a global task=] on the [=DOM manipulation task source=]  given |globalObject| to
+        set |config| to the result of [=fetch the config file=] with |provider|.
+    1. Wait until |check| and |config| are set.
     1. [=Queue a global task=] on the [=DOM manipulation task source=]  given |globalObject| to
         set |accountsList| to the result of [=fetch the accounts list=] with |config| and |provider|.
-    1. Wait for |accountsList| to become non-null.
-    1. Assert: |accountsList| is an {{IdentityProviderAccountList}}.
+    1. Wait until |accountsList| is set.
     1. If |accountsList|'s size is 1:
         1. Let |account| be |accountsList|[0].
         1. Let |accountState| be the result of running the [=compute account state=] algorithm
@@ -1255,12 +1251,10 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
     1. Wait until the [=user agent=]'s dialog is closed and |accountState| has been updated.
     1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then throw a new
         "{{NetworkError}}" {{DOMException}}.
-    1. Let |credential| be null.
     1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to 
         set |credential| to the result of running [=create tokens=] algorithm with |accountState|,
         |account|'s {{IdentityProviderAccount/id}}, |provider|, |config|, and |globalObject|.
-    1. Wait until |credential| is not null.
-    1. Assert: |credential| is an {{IdentityCredential}}.
+    1. Wait until |credential| is set.
     1. Return |credential|.
 </div>
 
@@ -1479,7 +1473,7 @@ an [=AccountState=].
     1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to
         compute |configUrl| as the result of running [=url parser=] with |provider|'s
         {{IdentityProviderConfig/configURL}}.
-    1. Wait for the above task to run, and let |idpOrigin| be the [=origin=] corresponding to
+    1. Wait until |configUrl| is set, and let |idpOrigin| be the [=origin=] corresponding to
         |configUrl|.
     1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
     1. Let |accountId| be |account|'s {{IdentityProviderAccount/id}}.
@@ -1531,11 +1525,10 @@ To <dfn>request permission to sign-up</dfn> the user with a given an {{IdentityP
 [=AccountState=] |accountState|, an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderConfig}}
 |provider|, and a |globalObject|, run the following steps. This returns a boolean.
     1. Assert: These steps are running [=in parallel=].
-    1. Let |metadata| be false (|metadata| could be set to null, hence initialize differently here).
     1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to
         [=fetch the client metadata=] algorithm with |config| and |provider|, and set the result to
         |metadata|.
-    1. Wait until |metadata| is not false.
+    1. Wait until |metadata| is set.
     1. Assert: |metadata| is null or an {{IdentityProviderClientMetadata}}.
     1. If |metadata|["{{IdentityProviderClientMetadata/privacy_policy_url}}"] is defined and the |provider|'s
         {{IdentityProviderConfig/clientId}} is not in the list of
@@ -1700,39 +1693,43 @@ When the {{IdentityCredential/logoutRPs()}} method is invoked given a [=list=] o
 {{IdentityCredentialLogoutRPsRequest}}s |logoutRequests|, the user agent MUST execute the following
 steps. This returns a {{Promise}}.
     1. Let |promise| be a new {{Promise}}.
-    1. For each |request| in |logoutRequests|:
-        1. Let |rpOrigin| be [=this=]'s [=Document/origin=].
-        1. Let |idpOrigin| be |request|'s {{IdentityCredentialLogoutRPsRequest/url}}'s [=origin=].
-        1. Let |account| be |request|'s {{IdentityCredentialLogoutRPsRequest/accountId}}.
-        1. Let |triple| be (|rpOrigin|, |idpOrigin|, |account|).
-        1. If [=state machine map=][|triple|] does not exist, continue.
-        1. Let |accountState| be [=state machine map=][|triple|].
-        1. If the |accountState|'s {{AccountState/registration state}} is [=unregistered=] or
-            |accountState|'s {{AccountState/allows logout}} is false, continue.
-        1. Let |fetchRequest| be a new <a spec=fetch for=/>request</a> as follows:
+    1. Let |globalObject| be <a>this</a>'s <a>relevant global object</a>.
+    1. [=In parallel=], perform the following steps:
+        1. For each |request| in |logoutRequests|:
+            1. Let |rpOrigin| be [=this=]'s [=Document/origin=].
+            1. Let |idpOrigin| be |request|'s {{IdentityCredentialLogoutRPsRequest/url}}'s [=origin=].
+            1. Let |account| be |request|'s {{IdentityCredentialLogoutRPsRequest/accountId}}.
+            1. Let |triple| be (|rpOrigin|, |idpOrigin|, |account|).
+            1. If [=state machine map=][|triple|] does not exist, continue.
+            1. Let |accountState| be [=state machine map=][|triple|].
+            1. If the |accountState|'s {{AccountState/registration state}} is [=unregistered=] or
+                |accountState|'s {{AccountState/allows logout}} is false, continue.
+            1. Let |fetchRequest| be a new <a spec=fetch for=/>request</a> as follows:
 
-            :  [=request/url=]
-            :: |request|'s {{IdentityCredentialLogoutRPsRequest/url}}
-            :  [=request/mode=]
-            :: "GET"
-            :  [=request/redirect mode=]
-            :: "error"
-            :  [=request/client=]
-            :: null
-            :  [=request/window=]
-            :: "no-window"
-            :  [=request/service-workers mode=]
-            :: "none"
-            :  [=request/destination=]
-            :: "webidentity"
-            :  [=request/origin=]
-            :: a unique [=opaque origin=]
-            :  [=request/credentials mode=]
-            :: "include"
+                :  [=request/url=]
+                :: |request|'s {{IdentityCredentialLogoutRPsRequest/url}}
+                :  [=request/mode=]
+                :: "GET"
+                :  [=request/redirect mode=]
+                :: "error"
+                :  [=request/client=]
+                :: null
+                :  [=request/window=]
+                :: "no-window"
+                :  [=request/service-workers mode=]
+                :: "none"
+                :  [=request/destination=]
+                :: "webidentity"
+                :  [=request/origin=]
+                :: a unique [=opaque origin=]
+                :  [=request/credentials mode=]
+                :: "include"
 
-        1. [=Fetch=] |fetchRequest|.
-        1. Set the |accountState| {{AccountState/allows logout}} to false.
-    1. <a for=Promise>Resolve</a> |promise| with [undefined] and return |promise|.
+            1. [=Queue a global task=] on the [=network task source=]  given |globalObject|
+                to [=fetch=] |fetchRequest|.
+            1. Set the |accountState| {{AccountState/allows logout}} to false.
+        1. <a for=Promise>Resolve</a> |promise| with [undefined].
+    1. Return |promise|.
 </div>
 
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1192,7 +1192,7 @@ requests.
 When the {{IdentityCredential}}'s
 <dfn for="IdentityCredential" method>\[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)</dfn>
 algorithm is invoked, the user agent MUST execute the following steps. This returns an
-{{IdentityCredential}}, null, or an error.
+{{IdentityCredential}} (or throws an error to the caller).
 
     1. Assert: These steps are running [=in parallel=].
     1. Assert: |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"] [=map/exists=].
@@ -1206,68 +1206,67 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
         [issue](https://github.com/fedidcg/FedCM/issues/389).
 
         Note: the purpose of having a timer here is to avoid leaking the reason causing this
-        method to return null. If there was no such timer, the developer could easily infer
+        method to throw an error. If there was no such timer, the developer could easily infer
         whether the user has an account with the [=IDP=] or not, or whether the user closed the UI without granting permission to share the [=IDP=] account information with the [=RP=].
     1. Let |provider| be |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"][0].
-    1. Let |outputCredential| be null.
-    1. Run the [=potentially create an IdentityCredential=] algorithm with |provider|,
-        |globalObject|, and |outputCredential|.
+    1. Return the result of running [=create an IdentityCredential=] with |provider| and
+        |globalObject|.
 
         Note: The |globalObject| is not currently passed onto the {{Credential/[[DiscoverFromExternalSource]]}}
         algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
-
-    1. Wait until an error is thrown or |outputCredential| becomes non-null.
-    1. Assert: |outputCredential| is an {{IdentityCredential}}.
-    1. Return |outputCredential|.
 </div>
 
 <div algorithm>
-To <dfn>potentially create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
-|provider|, |globalObject|, and |outputCredential|, run the following steps:
+To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
+|provider| and a |globalObject|, run the following steps. This returns an {{IdentityCredential}}.
     1. Assert: These steps are running [=in parallel=].
-    1. Let |check| be null.
-    1.  [=Queue a global task=] on the [=DOM manipulation task source=]  given |globalObject| to
-        [=check the well-known file=] with |provider| and |check|.
-    1. Let |config| be null.
-    1.  [=Queue a global task=] on the [=DOM manipulation task source=]  given |globalObject| to
-        [=fetch the config file=] with |provider| and |config|.
-    1. Wait for |check| and |config| to become non-null (or for errors to be thrown).
+    1. Let |check|, |config| be variables initially null.
+    1.  [=Queue a global task=] on the [=DOM manipulation task source=]  given |globalObject| and
+        run the following steps:
+        1. Set |check| to the result of [=check the well-known file=] with |provider|.
+        1. Set |config| to the result of [=fetch the config file=] with |provider|.
+    1. Wait for |check| and |config| to become non-null.
     1. Assert: |check| is true and |config| is an {{IdentityProviderAPIConfig}}.
     1. Let |accountsList| be null.
     1. [=Queue a global task=] on the [=DOM manipulation task source=]  given |globalObject| to
-        [=fetch the accounts list=] algorithm with |config|, |provider|, and |accountsList|.
+        set |accountsList| to the result of [=fetch the accounts list=] with |config| and |provider|.
     1. Wait for |accountsList| to become non-null.
     1. Assert: |accountsList| is an {{IdentityProviderAccountList}}.
     1. If |accountsList|'s size is 1:
         1. Let |account| be |accountsList|[0].
         1. Let |accountState| be the result of running the [=compute account state=] algorithm
             given |provider|, |account|, and |globalObject|.
-        1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then run the
+        1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=],
             [=request permission to sign-up=] algorithm with |account|, |accountState|, |config|,
-            |provider|, and |globalObject|.
-        1. Otherwise, show a dialog to request user permission to sign in via |account|.
-        1. If the user grants permission, run the [=sign-in=] algorithm with |accountState|.
+            |provider|, and |globalObject| and set the result in |permission|.
+        1. Otherwise, show a dialog to request user permission to sign in via |account|, and set the
+            result in |permission|.
+        1. If |permission|, [=sign-in=] with |accountState|.
     1. Otherwise:
         1. Let |account| be the result of running the [=select an account=] from the
             |accountsList|.
-        1. If |account| is null, return null.
         1. Let |accountState| be the result of running the [=compute account state=] algorithm
             given |provider| and |account|.
-        1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then run the
-            [=request permission to sign-up=] algorithm with |account|, |accountState|, |config|,
-            |provider|, and |globalObject|.
-        1. Otherwise, run the [=sign-in=] algorithm with |accountState|.
+        1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=]:
+            1. Run the [=request permission to sign-up=] algorithm with |account|, |accountState|,
+                |config|, |provider|, and |globalObject| and set the result in |permission|.
+            1. If |permission|, [=sign-in=] with |accountState|.
+        1. Otherwise, [=sign-in=] with |accountState|.
     1. Wait until the [=user agent=]'s dialog is closed and |accountState| has been updated.
     1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then throw a new
-        "{{AbortError}}" {{DOMException}}.
-    1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to run
-        the [=create tokens=] algorithm with |accountState|, |account|'s
-        {{IdentityProviderAccount/id}}, |provider|, |config|, |globalObject|, and |outputCredential|.
+        "{{NetworkError}}" {{DOMException}}.
+    1. Let |credential| be null.
+    1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to 
+        set |credential| to the result of running [=create tokens=] algorithm with |accountState|,
+        |account|'s {{IdentityProviderAccount/id}}, |provider|, |config|, and |globalObject|.
+    1. Wait until |credential| is not null.
+    1. Assert: |credential| is an {{IdentityCredential}}.
+    1. Return |credential|.
 </div>
 
 <div algorithm>
-To <dfn>check the well-known file</dfn> given an {{IdentityProviderConfig}} |provider| and
-|outputCheck|, run the following steps:
+To <dfn>check the well-known file</dfn> given an {{IdentityProviderConfig}} |provider|, run the
+following steps. This returns a boolean.
     1. Assert: These steps are running on the [=DOM manipulation task source=].
     1. Let |configUrl| be the result of running [=url parser=] with |provider|'s
         {{IdentityProviderConfig/configURL}}.
@@ -1303,33 +1302,36 @@ To <dfn>check the well-known file</dfn> given an {{IdentityProviderConfig}} |pro
         with [=request/mode=] set to "unsafe-no-cors". See the relevant
         [pull request](https://github.com/whatwg/fetch/pull/1533) for details.</p>
 
+    1. Let |check| be null.
     1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the
         following steps given a <a spec=fetch for=/>response</a> |response|:
         1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
         1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderWellKnown}},
             |discovery|.
-        1. Set |outputCheck| as follows:
+        1. Set |check| as follows:
             1. If the [=list/size=] of |discovery|["{{IdentityProviderWellKnown/provider_urls}}"]
-                is greater than 1, set |outputCheck| to false.
+                is greater than 1, set |check| to false.
 
                 Issue: [relax](https://github.com/fedidcg/FedCM/issues/333) the size of the
                 provider_urls array.
 
-            1. Otherwise, set |outputCheck| to |discovery|["{{IdentityProviderWellKnown/provider_urls}}"]
+            1. Otherwise, set |check| to |discovery|["{{IdentityProviderWellKnown/provider_urls}}"]
                 [=string/is=] equal to |provider|'s {{IdentityProviderConfig/configURL}}.
-        1. If |outputCheck| is false, throw a "{{NetworkError}}" {{DOMException}}.
+        1. If |check| is false, throw a "{{NetworkError}}" {{DOMException}}.
+    1. [=In parallel=], wait until |check| becomes non-null, and return it.
 </div>
 
 <div algorithm>
 To <dfn>fetch the config file</dfn> algorithm given an {{IdentityProviderConfig}} |provider|,
-and an |outputConfig|:
+run the following steps. This returns an {{IdentityProviderAPIConfig}}:
     1. Assert: These steps are running on the [=DOM manipulation task source=].
     1. Let |configUrl| be the result of running [=url parser=] with |provider|'s
         {{IdentityProviderConfig/configURL}}.
-    1. If |configUrl| is failure, return null.
+    1. If |configUrl| is failure, throw a new "{{NetworkError}}" {{DOMException}}.
     1. Run a [[!CSP]] check with a [[CSP#directive-connect-src|connect-src]]
         directive on the URL passed as |configUrl|.
-    1. If |configUrl| is not a [=potentially trustworthy URL=], return null.
+    1. If |configUrl| is not a [=potentially trustworthy URL=], throw a new "{{NetworkError}}"
+        {{DOMException}}.
     1. Let |request| be a new <a spec=fetch for=/>request</a> as follows:
 
         :  [=request/url=]
@@ -1354,12 +1356,13 @@ and an |outputConfig|:
         :  [=request/credentials mode=]
         :: "omit"
 
+    1. Let |config| be null.
     1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the following
         steps given a <a spec=fetch for=/>response</a> |response|:
         1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
         1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderAPIConfig}} stored
-            in |outputConfig|.
-        1. If |outputConfig| is null, throw a "{{NetworkError}}" {{DOMException}}.
+            in |config|.
+    1. [=In parallel=], wait until |config| is not null, and return it.
 </div>
 
 NOTE: We use a two-tier file system in order to prevent the [=IDP=] to easily determine the [=RP=]
@@ -1372,7 +1375,8 @@ path manipulation to fingerprint (for instance, by including the RP in the path)
 
 <div algorithm>
 To <dfn>fetch the accounts list</dfn> given an {{IdentityProviderAPIConfig}} |config|, an
-{{IdentityProviderConfig}} |provider|, and |outputAccountsList|:
+{{IdentityProviderConfig}} |provider|, run the following steps. This returns an
+{{IdentityProviderAccountList}}.
     1. Assert: These steps are running on the [=DOM manipulation task source=].
     1. Let |accountsUrl| be the result of [=computing the manifest URL=] given |provider| and
         |config|["{{IdentityProviderAPIConfig/accounts_endpoint}}"].
@@ -1403,25 +1407,28 @@ To <dfn>fetch the accounts list</dfn> given an {{IdentityProviderAPIConfig}} |co
 
     Issue: The credentialed fetch in this algorithm can lead to a timing attack that leaks the user's identities before the user grants permission. This is an active area of investigation that is being explored [here](https://github.com/fedidcg/FedCM/issues/230#issuecomment-1089040953).
 
+    1. Let |accountsList| be null.
     1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the following
         steps given a <a spec=fetch for=/>response</a> |response|:
         1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
         1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderAccountList}}, and
-            store the result in |outputAccountsList|.
+            store the result in |accountsList|.
 
     Issue: We should validate the accounts list returned here for repeated ids, as described [here](https://github.com/fedidcg/FedCM/issues/336).
+
+    1. [=In parallel=], wait for |accountsList| to become non-null, and return it.
 </div>
 
 <div algorithm>
 To <dfn>create tokens</dfn> given an [=AccountState=] |accountState|, a {{USVString}} |accountId|,
-    an {{IdentityProviderConfig}} |provider|, an {{IdentityProviderAPIConfig}} |config|,
-    |globalObject|, and |outputCredential|:
+    an {{IdentityProviderConfig}} |provider|, an {{IdentityProviderAPIConfig}} |config|, and
+    |globalObject|, run the following steps. This returns an {{IdentityCredential}}.
     1. Assert: These steps are running on the [=DOM manipulation task source=].
     1. Assert |accountState|'s {{AccountState/registration state}} is [=registered=].
     1. Assert |accountState|'s {{AccountState/allows logout}} is true.
     1. Let |idTokenUrl| be the result of [=computing the manifest URL=] given |provider| and
         |config|["{{IdentityProviderAPIConfig/id_assertion_endpoint}}"].
-    1. If |idTokenUrl| is failure, return null.
+    1. If |idTokenUrl| is failure, throw a new "{{NetworkError}}" {{DOMException}}.
     1. Let |requestBody| be a [=string=] resulting in concatenating "client_id=",
         |provider|'s {{IdentityProviderConfig/clientId}}, "&nonce=",
         |provider|'s {{IdentityProviderConfig/nonce}}, "&account_id=", and |accountId|.
@@ -1453,13 +1460,15 @@ To <dfn>create tokens</dfn> given an [=AccountState=] |accountState|, a {{USVStr
         :  [=request/credentials mode=]
         :: "include"
 
-      1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the following
-          steps given a <a spec=fetch for=/>response</a> |response|:
-          1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
-          1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderToken}}, |token|.
-          1. Let |outputCredential| be a new {{IdentityCredential}} given |globalObject|'s
-              <a for="global object">realm</a>.
-          1. Set |outputCredential|'s {{IdentityCredential/token}} to |token|.
+    1. Let |credential| be null.
+    1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the following
+        steps given a <a spec=fetch for=/>response</a> |response|:
+        1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
+        1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderToken}}, |token|.
+        1. Let |credential| be a new {{IdentityCredential}} given |globalObject|'s
+            <a for="global object">realm</a>.
+        1. Set |credential|'s {{IdentityCredential/token}} to |token|.
+    1. [=In parallel=], wait for |credential| to be non-null, and return it.
 </div>
 
 <div algorithm>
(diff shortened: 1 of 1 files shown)


npm1:
CC @bvandersloot-mozilla @annevk in case you're interested in reviewing.

npm1:
I added a bunch of 'on' functions to receive the fetch results. @johannhof, could you take a look?

npm1:
Ping @samuelgoto can you take a look please?

npm1:
@domfarolino any chance you could take a look? I believe this PR is not too complicated and improves Fetch integration but Sam didn't feel comfortable taking the first look. Perhaps you can lend us your Fetch expertise :)

spec/index.bs
```diff
@@ -1202,65 +1203,114 @@ requests.
...
-            |provider|.
-        1. Otherwise, run the [=sign-in=] algorithm with |accountState|.
-    1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then return null.
-    1. Let |token| be the result of running the [=create tokens=] algorithm with |accountState|,
-        |account|'s {{IdentityProviderAccount/id}}, |provider|, and |manifest|.
-    1. Let |credential| be a new {{IdentityCredential}}.
-    1. Set |credential|'s {{IdentityCredential/token}} to |token|.
-    1. Return |credential|.
+To <dfn>check the well-known file</dfn> given an {{IdentityProviderConfig}} |provider| and a
+|globalObject|:
```
domfarolino, line 1216: I think a lot of the weaving `globalObject` through to different algorithms is unnecessary. Assuming `provider` comes from script, you can just queue tasks to its relevant settings object. See https://html.spec.whatwg.org/multipage/server-sent-events.html#the-eventsource-interface:relevant-settings-object for example.

spec/index.bs
```diff
@@ -1202,65 +1203,114 @@ requests.
...
+    1. Let |configUrl| be the result of running [=url parser=] with |provider|'s
+        {{IdentityProviderConfig/configURL}}.
+    1. If |configUrl| is failure, return false.
+    1. Let |rootUrl| be a new [=/URL=].
+    1. Set |rootUrl|'s [=url/scheme=] to |configUrl|'s [=url/scheme=].
+    1. Set |rootUrl|'s [=url/host=] to |configUrl|'s [=url/host=]'s [=host/registrable domain=].
+    1. Set |rootUrl|'s [=url/path=] to the <a>list</a> «".well-known", "web-identity"».
+    1. If |rootUrl| is not a [=potentially trustworthy URL=], return false.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |rootUrl|
```
domfarolino, line 1226: Editorial: I think it is cleaner and less polluting to the algorithm to initialize objects/structs like https://github.com/WICG/nav-speculation/blob/main/prefetch.bs#L379-L394

spec/index.bs
```diff
@@ -1202,65 +1203,114 @@ requests.
...
+|globalObject|:
+    1. Let |configUrl| be the result of running [=url parser=] with |provider|'s
+        {{IdentityProviderConfig/configURL}}.
+    1. If |configUrl| is failure, return false.
+    1. Let |rootUrl| be a new [=/URL=].
+    1. Set |rootUrl|'s [=url/scheme=] to |configUrl|'s [=url/scheme=].
+    1. Set |rootUrl|'s [=url/host=] to |configUrl|'s [=url/host=]'s [=host/registrable domain=].
+    1. Set |rootUrl|'s [=url/path=] to the <a>list</a> «".well-known", "web-identity"».
+    1. If |rootUrl| is not a [=potentially trustworthy URL=], return false.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
```
domfarolino, line 1225: I think I'd add an XXX box somewhere around the creation of this request, to track the "TODO" of making this request's `mode=unsafe-no-cors` (https://github.com/whatwg/fetch/pull/1533) once that mode is ready, since I think that's what we landed on. Linking to that PR is probably a good idea too.

spec/index.bs
```diff
@@ -1202,65 +1203,114 @@ requests.
...
+    1. Let |check| be false.
+    1. [=Queue a global task=] on the [=networking task source=] given |globalObject| to perform the
+        following steps:
+        1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the
+            following steps given a <a spec=fetch for=/>response</a> |response|:
+            1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
+            1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderWellKnown}},
+                |discovery|.
+            1. Let |check| be set as follows:
+                1. If the [=list/size=] of |discovery|["{{IdentityProviderWellKnown/provider_urls}}""]
```
domfarolino, line 1245: Is it correct that there are three double quote characters on this line (and a few lines below as well)?

spec/index.bs
```diff
@@ -1202,65 +1203,114 @@ requests.
...
+                    is greater than 1, set |check| to false.
+
+                    Issue: [relax](https://github.com/fedidcg/FedCM/issues/333) the size of the provider_urls array.
+
+                1. Otherwise, set |check| to true if |discovery|["{{IdentityProviderWellKnown/provider_urls}}""]
+                    [=string/is=] equal to |provider|'s {{IdentityProviderConfig/configURL}}.
```
domfarolino, line 1251: Is it possible for _this_ condition to also not be satisfied, and then `check` is uninitialized?

spec/index.bs
```diff
@@ -1202,65 +1203,114 @@ requests.
...
+        1. [=request/header list=] set to a [=list=] containing a single [=header=] with
+            [=header/name=] set to `Accept` and [=header/value=] set to `application/json`
+        1. [=request/referrer policy=] set to "no-referrer"
+        1. [=request/credentials mode=] set to "omit"
+    1. Let |check| be false.
+    1. [=Queue a global task=] on the [=networking task source=] given |globalObject| to perform the
+        following steps:
+        1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the
+            following steps given a <a spec=fetch for=/>response</a> |response|:
+            1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
```
domfarolino, line 1241: So I think these steps want to be running on the "main thread" (since it calls "On well-known check completed" which throws errors to JS), but I don't think that happens in practice. Note that fetch's [task destination](https://fetch.spec.whatwg.org/#ref-for-fetch-params-task-destination) will be null in this case if I understand correctly.

The unsafe-no-cors work needs to partially [fix this](https://github.com/whatwg/fetch/pull/1533/#issuecomment-1325762475) so that the task destination is at least non-null for this kind of request (made in parallel, detached from the main thread), but even once fixed, your response handling code will end up back on the "in parallel" queue that you originally ran on. If you want to go back to the main thread from here you will need to post another task to the main thread, since there doesn't seem to be a way to manipulate fetch's "fetch params" directly from the caller.

spec/index.bs
```diff
@@ -1202,65 +1203,114 @@ requests.
...
-    1. If [=state machine map=][|triple|] does not exist, set [=state machine map=][|triple|] to a
-        new [=AccountState=].
-    1. Let |accountState| be [=state machine map=][|triple|].
-    1. Return |accountState|.
+    1. If |configUrl| is failure, return null.
+    1. Run a [[!CSP]] check with a [[CSP#directive-connect-src|connect-src]]
+        directive on the URL passed as |configUrl|.
+    1. If |configUrl| is not a [=potentially trustworthy URL=], return null.
+    1. Let |request| be a new <a spec=fetch for=/>request</a> with the following properties:
+        1. [=request/url=] set to |configUrl|
```
domfarolino, line 1275: same comment regarding initialization

domfarolino (commented):
Here a few comments from skimming through this this evening, but I'll give it a more thorough look later.

spec/index.bs
npm1: Ah, thats unfortunate. I just added a queue a task right at the beginning of the `processResponseConsumeBody`, I guess that should fix it for now.

spec/index.bs
npm1: Ah yes this is not clear, I removed the 'if'. It should be check IS the string/is comparison, if that makes sense.

spec/index.bs

spec/index.bs
domfarolino: I think you're right! I guess this is fine then.

spec/index.bs
```diff
@@ -47,6 +47,7 @@ spec: html; urlPrefix: https://html.spec.whatwg.org/multipage/
     type: dfn
         text: origin; for: html-origin-def; url: origin.html#concept-origin
         text: Content-Type Metadata; url: urls-and-fetching.html#content-type
+        text: DOM manipulation task source; url: webappapis.html#dom-manipulation-task-source
```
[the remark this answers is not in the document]
domfarolino: Really weird. https://github.com/whatwg/html/pull/6360 tracks this issue. I suppose we should just add a comment to remove this once it is solved, just like https://github.com/whatwg/notifications/blob/main/notifications.bs#L12 does.

spec/index.bs
domfarolino: I guess just one is fine, but maybe mention that the XXX applies to all other requests, just so we don't have to duplicate it unncessarily?

spec/index.bs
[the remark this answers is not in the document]
npm1: Done

spec/index.bs
npm1: I know, I don't love it either. I wish we could pull it out of something!

spec/index.bs

npm1 (commented):
Thanks for the review so far!

spec/index.bs
```diff
@@ -1189,7 +1191,7 @@ requests.
 <div algorithm>
     When the {{IdentityCredential}}'s
     <dfn for="IdentityCredential" method>\[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)</dfn>\
```
domfarolino, line 1193: What's the trailing "\" for here?

spec/index.bs
```diff
@@ -1189,7 +1191,7 @@ requests.
...
-    algorithm is invoked, the user agent MUST execute the following steps:
+    algorithm is invoked, the user agent MUST execute the following steps (note that these are invoked while [=in parallel=]):
```
domfarolino, line 1194: nit: I'd probably add an `Assert: These steps are running [=in parallel=]` as the first step of the algorithm just for more normative clarity

spec/index.bs
```diff
@@ -1202,27 +1204,190 @@ requests.
           method to return null. If there was no such timer, the developer could easily infer
```
domfarolino, line 1204: It won't let me add a comment on the above lines this PR doesn't change, but I want to note that you shouldn't call `setTimeout()` directly here, for two reasons:
 - You're in parallel but setTimeout() is a web-exposed DOM API
 - We shouldn't call raw DOM APIs from spec prose in general. Some JS could override the `setTimeout()` web-exposed API to do something else, and your spec would call that "something else", whereas you want the spec to always call the intended underlying algorithms. See for example, the XXX box in https://html.spec.whatwg.org/multipage/urls-and-fetching.html#start-intersection-observing-a-lazy-loading-element that says "The intention is to use the original value of the [...]".

Maybe you could keep this as-is for now but add an XXX box and file a follow-up bug?

Also, is the intention to *always* throw the NetworkError from the timeout, or should we be canceling the task if we return an IdentityCredential? It looks like we're not canceling the task anywhere, so I think the error would unconditionally be thrown which is weird. One more thing I wanted to mention is that since you're in parallel, you can probably just synchronously wait for which ever comes first: 60 seconds, or a returned IdentityCredential. I don't think you need to post a timer task anyways, since there isn't any use in "unblocking" the in-parallel context

domfarolino (commented):
One thing I'd recommend is prefacing each algorithm with its return type, so we can more-easily track that all of the return values are appropriately used. I think I spotted some cases where algorithms were returning things like "false" and "an empty list" to nobody that was doing anything with those values, so it might be good to structure the algorithms as I mentioned, and do an audit.

The common way to do this is:

"To _foo_ given a _bar_ and _baz_, run the following steps. They return a _qux_."

spec/index.bs
npm1: The backslash? Not sure, removing.

spec/index.bs

spec/index.bs
npm1: File issue and added note, thanks!

spec/index.bs
```diff
@@ -1202,27 +1204,190 @@ requests.
...
+          Note: The |globalObject| is not currently passed onto the {{Credential/[[DiscoverFromExternalSource]]}}
+          algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
+      1. Wait until an error is thrown or an {{IdentityCredential}} <var ignore=''>credential</var>
+          is returned.
+</div>
+
+<div algorithm>
+To <dfn>check the well-known file</dfn> given an {{IdentityProviderConfig}} |provider| and a
+|globalObject|:
+    1. Let |configUrl| be the result of running [=url parser=] with |provider|'s
```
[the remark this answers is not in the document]
npm1: Done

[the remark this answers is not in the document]
npm1: Ah, since this is the first step I did the queueing before invoking this algorithm.

spec/index.bs
```diff
@@ -1202,27 +1204,190 @@ requests.
...
+    
+    Note: The spec is yet to be updated so that all <a spec=fetch for=/>requests</a> are created
+    with [=request/mode=] set to "unsafe-no-cors". See the relevant
+    [pull request](https://github.com/whatwg/fetch/pull/1533) for details.
+    1. Let |check| be false.
+    1. [=Queue a global task=] on the [=networking task source=] given |globalObject| to perform the
+        following steps:
+        1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the
+            following steps given a <a spec=fetch for=/>response</a> |response|:
+            1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject|
```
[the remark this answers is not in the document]
npm1: Are you sure fetch can be invoked from in parallel? I was told before that it cannot (easily) https://github.com/fedidcg/FedCM/issues/261#issuecomment-1268064407

spec/index.bs
```diff
@@ -1202,27 +1204,190 @@ requests.
...
+    1. Run [=fetch the internal manifest=], passing |provider| and |globalObject|.
+</div>
+
+NOTE: The well-known file and internal manifest fetches could be performed simultaneously, it is
+just simpler to specify these as happening one after the other.
+
+<div algorithm>
+To <dfn>fetch the internal manifest</dfn> algorithm given an {{IdentityProviderConfig}} |provider|
+and a |globalObject|:
+    1. Let |configUrl| be the result of running [=url parser=] with |provider|'s
```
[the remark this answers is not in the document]
npm1: Done

spec/index.bs
```diff
@@ -1202,27 +1204,190 @@ requests.
...
+        :: a unique [=opaque origin=]
+        :  [=request/header list=]
+        :: a [=list=] containing a single [=header=] with [=header/name=] set to `Accept` and
+            [=header/value=] set to `application/json`
+        :  [=request/referrer policy=]
+        :: "no-referrer"
+        :  [=request/credentials mode=]
+        :: "omit"
+    
+    Note: The spec is yet to be updated so that all <a spec=fetch for=/>requests</a> are created
```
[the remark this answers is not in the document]
npm1: I tried `<p class=XXX>...</p>` and `<p class=NOTE>...</p>` but they do not produce any boxes. Help me understand the suggestion please

spec/index.bs
```diff
@@ -1202,27 +1204,190 @@ requests.
...
+        :: a unique [=opaque origin=]
+        :  [=request/header list=]
+        :: a [=list=] containing a single [=header=] with [=header/name=] set to `Accept` and
+            [=header/value=] set to `application/json`
+        :  [=request/referrer policy=]
+        :: "no-referrer"
+        :  [=request/credentials mode=]
+        :: "omit"
+
+    1. [=Queue a global task=] on the [=networking task source=] given |globalObject| to perform the
```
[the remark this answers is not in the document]
npm1: Thanks, removed.

npm1:
> One thing I'd recommend is prefacing each algorithm with its return type, so we can more-easily track that all of the return values are appropriately used. I think I spotted some cases where algorithms were returning things like "false" and "an empty list" to nobody that was doing anything with those values, so it might be good to structure the algorithms as I mentioned, and do an audit.
> 
> The common way to do this is:
> 
> "To _foo_ given a _bar_ and _baz_, run the following steps. They return a _qux_."

Thanks, added return type to the algorithms which had one.

spec/index.bs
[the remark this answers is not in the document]
domfarolino: Yeah I think I'm wrong, we should be fetching from the main thread.

spec/index.bs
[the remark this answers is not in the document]
domfarolino: Ah, I think it is pretty common for WHATWG-style CSS to be copied around bikeshed specs, as in https://github.com/WICG/fenced-frame/blob/master/spec.bs#L133-L138. I was just assuming bs would do this for us, and I forgot about the duplicated CSS. Hmm, in that case I suppose `<p class=issue>` is probably appropriate.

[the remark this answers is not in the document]

npm1 (commented):
Ready for another review! Added asserts and removed some unneeded task queueing.

spec/index.bs
```diff
@@ -1187,42 +1189,212 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
-      1. Let |credential| be the result of running the [=potentially create IdentityCredential=]
-          algorithm with |provider|.
-      1. If |credential| is null or [=potentially create IdentityCredential=] threw a {{TypeError}}, wait for the task that throws a {{NetworkError}}, otherwise return
-          |credential|.
+      1. [=Queue a global task=] on the [=DOM manipulation task source=] to
+          [=check the well-known file=] algorithm with |provider| and |globalObject|.
+
+          Note: The |globalObject| is not currently passed onto the {{Credential/[[DiscoverFromExternalSource]]}}
+          algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
+      1. Wait until an error is thrown or an {{IdentityCredential}} <var ignore=''>credential</var>
```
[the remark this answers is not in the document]
domfarolino: Well what are we doing what the thing that we're waiting for? Are we supposed to return it, or view it in some way? It seems like we're waiting for something to happen but not doing anything with that happens.

domfarolino (commented):
This generally LGTM. I think it's probably fine to clean the hand-wavy "Wait" part up in a follow-up, provided we clarify what we're actually doing with the thing we're waiting for.

spec/index.bs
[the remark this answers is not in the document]
npm1: We return it here yes. Making it clearer.

npm1:
@samuelgoto @domfarolino lemme know if there are any objections to merging this! I want to get this merged to work on other stuff in the spec (this touches everything so easier to have it merged before)

samuelgoto:
Will look at this first thing today!

Thanks for setting this up!

spec/index.bs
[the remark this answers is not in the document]
domfarolino: Am I understanding correctly that we're waiting for the "check the well-known file" algorithm (that is running on the main thread) to return something, and then we're somehow returning it from the in-parallel context? That's kind of odd if I'm reading it correctly.

Also it does not seem that the "check the well-known file" algorithm actually returns anything, nor do the algorithms in the chain of subroutines, if I'm reading correctly.

spec/index.bs
```diff
@@ -1187,42 +1189,209 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+        1. Let |check| be set as follows:
+            1. If the [=list/size=] of |discovery|["{{IdentityProviderWellKnown/provider_urls}}"]
+                is greater than 1, set |check| to false.
+
+                Issue: [relax](https://github.com/fedidcg/FedCM/issues/333) the size of the
+                provider_urls array.
+
+            1. Otherwise, set |check| to |discovery|["{{IdentityProviderWellKnown/provider_urls}}"]
+                [=string/is=] equal to |provider|'s {{IdentityProviderConfig/configURL}}.
+        1. Run [=on well-known check completed=], passing |check|, |provider|, and |globalObject|.
```
samuelgoto, line 1274: Don't we [=fetch the internal manifest=] in parallel with the well-known file rather than after it?

spec/index.bs
```diff
@@ -1187,42 +1189,209 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+</div>
+
+<div algorithm>
+<dfn>On well-known check completed</dfn>, given a boolean |check|, an {{IdentityProviderConfig}}
+|provider| and a |globalObject|:
+    1. If |check| is false, throw a "{{NetworkError}}" {{DOMException}}.
```
samuelgoto, line 1280: Hum, interesting ... so, if you throw an exception inside the [=Queue a global task=], then it manages to pass the exception back to the {{Credential/[[DiscoverFromExternalSource]]}} algorithm? Is that done by magic or is that specified somewhere?

spec/index.bs
samuelgoto: Ah, you do note that later in the spec on "just simpler to specify these as happening one after the other." .... but that seems not great, right? Seems a bit like it would lead to the wrong implementation ... 

Would it make it substantially harder to write the spec in the way that represents the desired implementation? Maybe if we had the [=potentially create IdentityCredential=] top level method we would be able to make both of these fetches in parallel?

spec/index.bs
```diff
@@ -1187,42 +1189,209 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+just simpler to specify these as happening one after the other.
+
+<div algorithm>
+To <dfn>fetch the internal manifest</dfn> algorithm given an {{IdentityProviderConfig}} |provider|
+and a |globalObject|:
+    1. Assert: These steps are running on the [=DOM manipulation task source=].
+    1. Let |configUrl| be the result of running [=url parser=] with |provider|'s
+        {{IdentityProviderConfig/configURL}}.
+    1. If |configUrl| is failure, return null.
+    1. Run a [[!CSP]] check with a [[CSP#directive-connect-src|connect-src]]
```
samuelgoto, line 1294: @cbiesinger can you sanity check if this is correct?

spec/index.bs
```diff
@@ -1187,42 +1189,209 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+        1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
+        1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderAPIConfig}}, |manifest|.
+        1. Run [=on internal manifest fetched=], passing |manifest|, provider|, and |globalObject|.
+</div>
+
+NOTE: We use a two-tier manifest list in order to prevent the [=IDP=] to easily determine the [=RP=]
+that a user is visiting by encoding the information in the manifest path. We solve this issue by
+requiring a manifest list to be on the root of the [=IDP=]. The manifest itself can be anywhere, but
+it will not be used if the user agent does not find it in the manifest list. This allows the [=IDP=]
+to keep their actual manifest on an arbitary path while allowing the user agent to prevent manifest
```
samuelgoto, line 1332:
```suggestion
to keep their actual config files on an arbitary path while allowing the user agent to prevent the config file have information about the RP.
```

spec/index.bs
```diff
@@ -1187,42 +1189,209 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+        1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
+        1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderAccountList}}, |accountsList|.
+        1. Run [=on accounts list fetched=], passing |accountsList|, |manifest|, |provider|, and |globalObject|.
+
+    Issue: We should validate the accounts list returned here for repeated ids, as described [here](https://github.com/fedidcg/FedCM/issues/336).
+</div>
+
+<div algorithm>
+<dfn>On accounts list fetched</dfn>, given an {{IdentityProviderAccountList}} |accountsList|, an
+{{IdentityProviderAPIConfig}} |manifest|, an {{IdentityProviderConfig}} |provider| and a
```
samuelgoto, line 1386:
```suggestion
{{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderConfig}} |provider| and a
```

spec/index.bs
```diff
@@ -1232,24 +1401,85 @@ To <dfn>potentially create IdentityCredential</dfn>, given an {{IdentityProvider
...
-            [=request permission to sign-up=] algorithm with |account|, |accountState|, |manifest|, and
-            |provider|.
+            [=request permission to sign-up=] algorithm with |account|, |accountState|, |manifest|,
+            |provider|, and |globalObject|.
         1. Otherwise, run the [=sign-in=] algorithm with |accountState|.
-    1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then return null.
-    1. Let |token| be the result of running the [=create tokens=] algorithm with |accountState|,
-        |account|'s {{IdentityProviderAccount/id}}, |provider|, and |manifest|.
-    1. Let |credential| be a new {{IdentityCredential}}.
+    1. Wait until the user agent's dialog is closed and |accountState| has been updated.
```
samuelgoto, line 1407:
```suggestion
    1. Wait until the [=user agent=]'s dialog is closed and |accountState| has been updated.
```

spec/index.bs
```diff
@@ -1232,24 +1401,85 @@ To <dfn>potentially create IdentityCredential</dfn>, given an {{IdentityProvider
...
+    1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then throw a new
+        "{{AbortError}}" {{DOMException}}.
+    1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to run
```
samuelgoto, line 1410: Why do we need to enqueue a global task here to fetch the [=create tokens=] fetch? Why isn't this fetch done like the other ones?

spec/index.bs
samuelgoto, line 1407: Can we use the "wait until" pattern that you use here rather than series of callbacks for the fetches? That seems like it would lead to more readable text.

spec/index.bs
```diff
@@ -1232,24 +1401,85 @@ To <dfn>potentially create IdentityCredential</dfn>, given an {{IdentityProvider
...
+</div>
+
+<div algorithm>
+<dfn>On token created</dfn>, given an {{IdentityProviderToken}} |token| and a |globalObject|:
+    1. Assert: These steps are running on the [=DOM manipulation task source=].
+    1. Create a new {{IdentityCredential}} given |globalObject|'s <a for="global object">realm</a>.
     1. Set |credential|'s {{IdentityCredential/token}} to |token|.
-    1. Return |credential|.
+    1. [=In parallel=], pass |credential| to the {{Credential/[[DiscoverFromExternalSource]]}}
+        callback.
```
samuelgoto, line 1469: What does "callback" refer to?

spec/index.bs
```diff
@@ -1612,16 +1704,27 @@ steps:
...
+            :  [=request/destination=]
+            :: "webidentity"
+            :  [=request/origin=]
+            :: a unique [=opaque origin=]
+            :  [=request/credentials mode=]
+            :: "include"
+
         1. [=Fetch=] |fetchRequest|.
         1. Set the |accountState| {{AccountState/allows logout}} to false.
     1. <a for=Promise>Resolve</a> |promise| with [undefined] and return |promise|.
```
samuelgoto, line 1730: How does this promise get resolved?

spec/index.bs
```diff
@@ -1187,42 +1189,209 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+    1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the following
+        steps given a <a spec=fetch for=/>response</a> |response|:
+        1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
+        1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderAPIConfig}}, |manifest|.
+        1. Run [=on internal manifest fetched=], passing |manifest|, provider|, and |globalObject|.
+</div>
+
+NOTE: We use a two-tier manifest list in order to prevent the [=IDP=] to easily determine the [=RP=]
+that a user is visiting by encoding the information in the manifest path. We solve this issue by
+requiring a manifest list to be on the root of the [=IDP=]. The manifest itself can be anywhere, but
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -1187,42 +1189,209 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
-          algorithm with |provider|.
-      1. If |credential| is null or [=potentially create IdentityCredential=] threw a {{TypeError}}, wait for the task that throws a {{NetworkError}}, otherwise return
-          |credential|.
+      1. [=Queue a global task=] on the [=DOM manipulation task source=]  given |globalObject| to
+          [=check the well-known file=] algorithm with |provider| and |globalObject|.
+
+          Note: The |globalObject| is not currently passed onto the {{Credential/[[DiscoverFromExternalSource]]}}
+          algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
+      1. Wait until an error is thrown or an {{IdentityCredential}} <var ignore=''>credential</var>
+          is returned, and return it here.
```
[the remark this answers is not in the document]
npm1: Fixed I think

spec/index.bs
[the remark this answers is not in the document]
npm1: Fixed I think

spec/index.bs
npm1: ?

spec/index.bs
```diff
@@ -1187,42 +1189,62 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
-        algorithm with |provider|.
-    1. If |manifest| is null, return null.
-    1. Let |accountsList| be the result of running the
-        [=fetch the accounts list=] algorithm with |manifest| and |provider|.
-    1. If |accountsList|'s size is 0, return null.
+To <dfn>potentially create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
+|provider|, |globalObject|, and |outputCredential|, run the following steps:
+    1. Assert: These steps are running [=in parallel=].
+    1. Let |check| be null.
+    1.  [=Queue a global task=] on the [=DOM manipulation task source=]  given |globalObject| to
```
samuelgoto, line 1229:
```suggestion
    1.  Let |check| be the result we get from enqueuing the [=check the well-known file=] task into the [=Queue a global task=] on the [=DOM manipulation task source=]  given |globalObject| to
```

spec/index.bs
```diff
@@ -1232,85 +1254,256 @@ To <dfn>potentially create IdentityCredential</dfn>, given an {{IdentityProvider
...
-    1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then return null.
-    1. Let |token| be the result of running the [=create tokens=] algorithm with |accountState|,
-        |account|'s {{IdentityProviderAccount/id}}, |provider|, and |manifest|.
-    1. Let |credential| be a new {{IdentityCredential}}.
-    1. Set |credential|'s {{IdentityCredential/token}} to |token|.
-    1. Return |credential|.
+    1. Wait until the [=user agent=]'s dialog is closed and |accountState| has been updated.
+    1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then throw a new
+        "{{AbortError}}" {{DOMException}}.
+    1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to run
```
samuelgoto, line 1263:
```suggestion
    1. Return the result of the [==create tokens] algorithm once it is queued into the [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to run
```

spec/index.bs
```diff
@@ -1323,11 +1516,58 @@ To <dfn>extract the JSON fetch response</dfn> given a <a spec=fetch for=/>respon
...
-    |provider|:
-    1. Let |metadata| be the result of running the [=fetch the client metadata=]
-        algorithm with |manifest| and |provider|.
+To <dfn>request permission to sign-up</dfn> the user with a given an {{IdentityProviderAccount}} |account|, an
+[=AccountState=] |accountState|, an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderConfig}}
+|provider|, and a |globalObject|:
+    1. Assert: These steps are running [=in parallel=].
+    1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to
+        [=fetch the client metadata=] algorithm with |account|, |accountState|, |config|,
+        |provider|.
```
samuelgoto, line 1525:
```suggestion
        |provider| and let |clientMetadata| be the result.
      1. Prompt the user ....
```

domfarolino:
I haven't had time to look deeply at the changes to the chaining and "async/await" as @samuelgoto as put it, can you summarize it briefly?

npm1:
> I haven't had time to look deeply at the changes to the chaining and "async/await" as @samuelgoto as put it, can you summarize it briefly?

In the latest change, I used the following style. In an algorithm (eg. fetch X):

```
1. Steps in the DOM manipulation task source.
2. Let |returnVar| be null
3. Fetch and in the fetch callback set |returnVar|.
4. In parallel, wait for |returnVar| to become non-null, and return it.
```

That is kinda the closest I could get to the await/async syntax that Sam wanted to simplify the methods / avoid a long chaining of callbacks.

spec/index.bs
```diff
@@ -1187,147 +1189,354 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
-        algorithm with |provider|.
-    1. If |manifest| is null, return null.
-    1. Let |accountsList| be the result of running the
-        [=fetch the accounts list=] algorithm with |manifest| and |provider|.
-    1. If |accountsList|'s size is 0, return null.
+To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
+|provider| and a |globalObject|, run the following steps. This returns an {{IdentityCredential}}.
+    1. Assert: These steps are running [=in parallel=].
+    1. Let |check|, |config| be variables initially null.
+    1.  [=Queue a global task=] on the [=DOM manipulation task source=]  given |globalObject| and
```
samuelgoto, line 1224:
```suggestion
    1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| and
```

spec/index.bs
```diff
@@ -1187,147 +1189,354 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+        run the following steps:
```
samuelgoto, line 1225:
```suggestion
        run the following steps in parallel:
```

spec/index.bs
```diff
@@ -1187,147 +1189,354 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+        1. Set |check| to the result of [=check the well-known file=] with |provider|.
+        1. Set |config| to the result of [=fetch the config file=] with |provider|.
+    1. Wait for |check| and |config| to become non-null.
+    1. Assert: |check| is true and |config| is an {{IdentityProviderAPIConfig}}.
```
samuelgoto, line 1229:
```suggestion
```
Is this really necessary?

spec/index.bs
```diff
@@ -1187,147 +1189,354 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+    1. Let |accountsList| be null.
```
samuelgoto, line 1230: Is there something special about "null"? Should we use "undefined" instead? I'm wondering if some of these algorithms can actually return "null" as a valid result, and these "wait until is null" will never unblock.

If we used "undefined" (as opposed to "null"), then maybe the "wait" could be "wait until the |accountsList| is non-undefined".

Or maybe we could just say "wait until the |accountList| is set" without initializing it (to null)?

spec/index.bs
```diff
@@ -1187,147 +1189,354 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
-            |provider|.
-        1. Otherwise, show a dialog to request user permission to sign in via |account|.
-        1. If the user grants permission, run the [=sign-in=] algorithm with |accountState|.
+            given |provider|, |account|, and |globalObject|.
+        1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=],
+            [=request permission to sign-up=] algorithm with |account|, |accountState|, |config|,
+            |provider|, and |globalObject| and set the result in |permission|.
+        1. Otherwise, show a dialog to request user permission to sign in via |account|, and set the
+            result in |permission|.
+        1. If |permission|, [=sign-in=] with |accountState|.
```
samuelgoto, line 1244:
```suggestion
        1. If |permission|, [=sign-in=] with |accountState|, otherwise throw a "{{NetworkError}}" {{DOMException}}.
```

I'm trying to follow the code here, but shouldn't we introduce a "if not permission throw Error"?

spec/index.bs
```diff
@@ -1187,147 +1189,354 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+
+<div algorithm>
+To <dfn>compute account state</dfn> given an {{IdentityProviderConfig}} |provider|, an
+{{IdentityProviderAccount}} |account|, and a |globalObject|, run the following steps. This returns
+an [=AccountState=].
+    1. Assert: These steps are running [=in parallel=].
+    1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to
+        compute |configUrl| as the result of running [=url parser=] with |provider|'s
+        {{IdentityProviderConfig/configURL}}.
+    1. Wait for the above task to run, and let |idpOrigin| be the [=origin=] corresponding to
```
samuelgoto, line 1482: This pattern here: I like it because it didn't require you to initialize |configURL| to "null" and wait for it to be non-null.

Wouldn't we be able to use this (i.e. text that says "Wait for the above task to run") across the other places where we enqueue a task?

spec/index.bs
```diff
@@ -1187,147 +1189,354 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
-        algorithm with |manifest| and |provider|.
+To <dfn>request permission to sign-up</dfn> the user with a given an {{IdentityProviderAccount}} |account|, an
+[=AccountState=] |accountState|, an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderConfig}}
+|provider|, and a |globalObject|, run the following steps. This returns a boolean.
+    1. Assert: These steps are running [=in parallel=].
+    1. Let |metadata| be false (|metadata| could be set to null, hence initialize differently here).
+    1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to
+        [=fetch the client metadata=] algorithm with |config| and |provider|, and set the result to
+        |metadata|.
+    1. Wait until |metadata| is not false.
```
samuelgoto, line 1538:
```suggestion
    1. Wait until the task above is finished and the |metadata| variable is set.
```

Here is one suggestion that I think would avoid the "Let |metadata| be false" step.

WDYT?

spec/index.bs
samuelgoto: Is this already a "resolved" promise (with the value of "undefined") that we are returning?

samuelgoto (commented):
LGTM

@domfarolino can you take a last look and approve too?

Once @domfarolino LGTMs and this last round of feedback is addressed (I think there is a slightly cleaner way to deal with the async/awaits), I'm planning to squash and merge.

spec/index.bs
npm1: Ah got it, yea I wasn't looking closely at stuff outside of the `Discover` section and unrelated to fetch. I think I fixed this

domfarolino:
> In the latest change, I used the following style. In an algorithm (eg. fetch X):
> ```1. Steps in the DOM manipulation task source.
> 2. Let |returnVar| be null
> 3. Fetch and in the fetch callback set |returnVar|.
> 4. In parallel, wait for |returnVar| to become non-null, and return it.
> ```
> That is kinda the closest I could get to the await/async syntax that Sam wanted to simplify the methods / avoid a long chaining of callbacks.

On first scan, this looks nice but I think it has some issues. First, you'd need to be more explicit about posting tasks. An algorithm that has its control flow broken up by "Waiting in parallel" would need to explicitly come back to the original-calling thread to return a value. The common trend I see for this elsewhere is:

```
1. Do some main-thread stuff
2. Wait in-parallel for some main-thread state to change, then post a task back to the main thread to run the following steps
3. (Back on main thread), do something with the value `foo` that we now know has been changed to what we want
```

That would normally work, depending on what you do with `foo`. The problem with your algorithms, it seems, is that they need to _return_ the value `foo` that you waited for in a non-blocking way. But I don't think you can really return from an algorithm once its control flow has been interrupted by waiting "in-parallel" (as to not block the main thread). By the time we come back to the main thread in step 3, the original caller of the algorithm has already totally finished invoking it, so I don't think we can return a concrete value after that. That's why async/await works in JS: `async` algorithms don't actually return concrete values, they all return promises so that the caller can only retrieve the value asynchronously after the algorithm was invoked, but I don't think we have a spec prose equivalent.

Do your algorithms have to return values, or can they just modify state? That is, can you do the following? :

```
4. In parallel, wait for `returnVar` to become non-null, and do something with it (but not return it)
```

npm1:
Hey @samuelgoto @domfarolino the latest PR has changes that hopefully satisfy both of you. Now most stuff is run in parallel and we only go into a task when necessary, thus making it possible to wait for the results of fetches while still allowing the algorithms to return the awaited values. PTAL and let me know what you think!

spec/index.bs
```diff
@@ -1401,27 +1410,26 @@ To <dfn>fetch the accounts list</dfn> given an {{IdentityProviderAPIConfig}} |co
 
     Issue: The credentialed fetch in this algorithm can lead to a timing attack that leaks the user's identities before the user grants permission. This is an active area of investigation that is being explored [here](https://github.com/fedidcg/FedCM/issues/230#issuecomment-1089040953).
 
-    1. Let |accountsList| be null.
-    1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to the following
-        steps given a <a spec=fetch for=/>response</a> |response|:
+    1. [=Queue fetch=] with |request|, |globalObject|, and <var ignore=>processResponseConsumeBody</var>
```
samuelgoto, line 1413:
```suggestion
    1. [=Fetch=] with |request|, |globalObject|, and <var ignore=>processResponseConsumeBody</var>
```

Can we just call this "fetch" for short (and skip the queueing)? The queuing can be an implementation specific part and I think would make things more readable.

spec/index.bs
```diff
@@ -1485,15 +1490,16 @@ an [=AccountState=].
...
-        {{IdentityProviderConfig/configURL}}.
-    1. Let |manifestUrl| be the result of running [=url parser=] given |manifestString|.
-    1. If |manifestUrl| is failure, let |manifestUrl| be the result of running [=url parser=] given
-        |manifestString| (the relative URL) and |configUrl| (the base URL).
+When <dfn>computing the manifest URL</dfn> given an {{IdentityProviderConfig}} |provider|, a
+[=string=] |manifestString|, and |globalObject|, perform the following steps. This returns a
+<a spec=url for=/>URL</a> or failure.
+    1. Let |configUrl| be the result of running [=queue url parser=] with |provider|'s
+        {{IdentityProviderConfig/configURL}} and |globalObject|.
+    1. Let |manifestUrl| be the result of running [=queue url parser=] given |manifestString| and
```
samuelgoto, line 1498:
```suggestion
    1. Let |manifestUrl| be the result of running [=parse url=] given |manifestString| and
```

Same here, can we skip the implementation that this is going to be queueing things?

samuelgoto (approved):
The last commit makes changes that I think meet the readability bar that I was trying to maintain. If this meets the fetch precision bar, I think we should merge this.

Dominic, WDYT?

spec/index.bs
```diff
@@ -1187,147 +1189,347 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
-    1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then return null.
-    1. Let |token| be the result of running the [=create tokens=] algorithm with |accountState|,
-        |account|'s {{IdentityProviderAccount/id}}, |provider|, and |manifest|.
-    1. Let |credential| be a new {{IdentityCredential}}.
-    1. Set |credential|'s {{IdentityCredential/token}} to |token|.
+        1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=]:
+            1. Run the [=request permission to sign-up=] algorithm with |account|, |accountState|,
+                |config|, |provider|, and |globalObject| and set the result in |permission|.
+            1. If |permission|, [=sign-in=] with |accountState|.
+        1. Otherwise, [=sign-in=] with |accountState|.
```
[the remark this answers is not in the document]
npm1: I don't think so. 1238 runs before this step, and that one prompts the account chooser. If the user is unregistered then it goes to 1242. 1246 would only be run if the user is registered.

spec/index.bs
npm1: I don't think we should name it the same as the actual `[=fetch=]` from the Fetch spec, that feels way more confusing IMO. Happy to rename it to something else though. Maybe `[=Await fetch=]`?

spec/index.bs
npm1: Ditto, can be `[=await url parser=]`?

spec/index.bs
```diff
@@ -1187,147 +1189,350 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+        [pull request](https://github.com/whatwg/fetch/pull/1533) for details.</p>
+
+    1. [=Queue fetch=] with |wellKnownRequest|, |globalObject|, and <var ignore=>processResponseConsumeBody</var>
+        set to the following steps given a <a spec=fetch for=/>response</a> |response|:
+        1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
+        1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderWellKnown}},
+            |discovery|.
+        1. Let |check| be set as follows:
+            1. If the [=list/size=] of |discovery|["{{IdentityProviderWellKnown/provider_urls}}"]
+                is greater than 1, set |check| to false.
```
samuelgoto, line 1329: Can't you avoid the |check| variable and "throw" when these conditions are met? e.g.

```suggestion
                is greater than 1, throw "{{NetworkError}}" {{DOMException}}.
```

And then later, also throw in line 1335?

spec/index.bs
```diff
@@ -1187,147 +1189,350 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+    Issue: We should validate the accounts list returned here for repeated ids, as described [here](https://github.com/fedidcg/FedCM/issues/336).
 
-        Note: This operation is not defined but should be removed when we [integrate](https://github.com/fedidcg/FedCM/issues/261) with Fetch.
+    1. Wait for |accountsList| to be set.
     1. Return |accountsList|.
+</div>
 
-        Issue: We should validate the accounts list returned here for repeated ids, as described [here](https://github.com/fedidcg/FedCM/issues/336).
+<div algorithm>
+To <dfn>create a credential</dfn> given an [=AccountState=] |accountState|, a {{USVString}}
```
samuelgoto, line 1426: Maybe we can use the same terminology for the algorithms that are used primarily for fetching things (this is a POST rather than a GET, but still, a [=fetch=])?
 
```suggestion
To <dfn>fetch an identity assertion</dfn> given an [=AccountState=] |accountState|, a {{USVString}}
```

WDYT?

spec/index.bs
samuelgoto: [=parse url=] already exists?

samuelgoto (approved):
I'm still waiting some confirmation from farolino that this works, but here is another round of feedback while we wait for his approval.

Again, I think this is ready to be merged as far as I'm concerned, and these can be resolved in a separate PR.

spec/index.bs
```diff
@@ -1187,147 +1189,350 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+        :: "no-referrer"
+        :  [=request/credentials mode=]
+        :: "omit"
+
+    1. [=Queue fetch=] with |configRequest|, |globalObject|, and <var ignore=>processResponseConsumeBody</var>
+        set to the following steps given a <a spec=fetch for=/>response</a> |response|:
+        1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
+        1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderAPIConfig}} stored
+            in |config|.
+    1. Wait for both |check| and |config| to be set.
```
domfarolino, line 1367: I think |check| doesn't exist at this point yet because it was defined for the first time in the fetch response handler.

spec/index.bs
```diff
@@ -1187,147 +1189,350 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+    1. Assert: |check| (should have thrown an exception if it is false).
```
domfarolino, line 1368: Are you referring to the following line?

> If |check| is false, throw a "{{NetworkError}}" {{DOMException}}.

This throws an exception on the main thread, but that wouldn't interrupt anything in parallel, so I think you could still get here with |check| being false.

spec/index.bs
```diff
@@ -1187,147 +1189,350 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
-        1. Otherwise, run the [=sign-in=] algorithm with |accountState|.
-    1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=] then return null.
-    1. Let |token| be the result of running the [=create tokens=] algorithm with |accountState|,
-        |account|'s {{IdentityProviderAccount/id}}, |provider|, and |manifest|.
-    1. Let |credential| be a new {{IdentityCredential}}.
-    1. Set |credential|'s {{IdentityCredential/token}} to |token|.
+        1. If |accountState|'s {{AccountState/registration state}} is [=unregistered=]:
+            1. Run the [=request permission to sign-up=] algorithm with |account|, |accountState|,
+                |config|, |provider|, and |globalObject| and set the result in |permission|.
+            1. If |permission|, [=sign-in=] with |accountState|.
```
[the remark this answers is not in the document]
npm1: We do not need to update `accountState` in that case

spec/index.bs
npm1: Removed |check| and now just waiting for 'both fetch responses to be completed'. Let me know if that works?

domfarolino:
Before I do a full scan, here's one example of a red flag I'm seeing kind of frequently: https://github.com/fedidcg/FedCM/pull/378/files#diff-40cc3a1ba233cc3ca7b6d5873260da9676f6ae20bb897b62f7871c80d0bda4e9R1284. Basically we're throwing a DOMException from in parallel at least 4 times in that algorithm, and I think elsewhere as well. But DOMExceptions are platform objects and must be created/thrown with a surrounding agent / global, so I don't think what's happening now is technically valid. This is why I mentioned on chat the other day when we discussed introducing "queue fetch":

> As long as everything besides those steps can run in parallel and doesn't need to be on the main thread
> Which I wasn't under the impression of earlier, but if that's the case that would be great

It seems like there are more things besides Fetch that currently expect to run on the main thread, which is a little worrying. I'm not sure what the full implementation flow is in Chromium so I don't know what to recommend for these algorithms, but it seems like spec-wise they all used to run on the main thread with the exception of "waiting" for fetches to complete, but the "waiting" got complicated and callback-ey, so you switched it so the algorithms ran entirely "in parallel" with the exception of kicking off main-thread tasks for URL parsing and Fetching, but this causes two issues:
 1. There are other things that have to run on the main thread that are currently not
 2. While the way we're kicking off tasks for URL parsing and Fetching is technically correct, it kind of hints to me that there is a design issue. I think it's uncommon to make fetches where the URL is parsed _asynchronously_ with respect to the Fetching, and this can have observable consequences too (i.e., you're allowing the `<base>` URL of the document to change in between parsing and fetching, whereas I'm guessing that is not possible in the implementation?).

I think (1) needs to be fixed regardless, but (2) might just be a symptom of the `unsafe-no-cors` Fetch mode not being ironed out yet. For example, as per https://github.com/whatwg/fetch/pull/1533#issuecomment-1325762475 we're going to want to Fetch these requests from some sort of detached context (I think in-parallel), whereas this PR sticks with today's precedent of only fetching from the main thread, which is best for now, but is in part why we have this weird asynchronous gap between parsing and Fetching and maybe more things. I think (2) can probably be overlooked for now, though (1) shouldn't be. But I don't know the best way to fix (1) without reverting the algorithms to run on the main thread and wait in parallel, but I am happy to hear other ideas @npm1 since you probably have the spec paged into your head more than me right now.

npm1:
> Before I do a full scan, here's one example of a red flag I'm seeing kind of frequently: https://github.com/fedidcg/FedCM/pull/378/files#diff-40cc3a1ba233cc3ca7b6d5873260da9676f6ae20bb897b62f7871c80d0bda4e9R1284. Basically we're throwing a DOMException from in parallel at least 4 times in that algorithm, and I think elsewhere as well. But DOMExceptions are platform objects and must be created/thrown with a surrounding agent / global, so I don't think what's happening now is technically valid.

Hmm yea, perhaps we can a) change those cases to 'return null', and then at the main method just queue a task to throw if the result was null b) add a helper so that any throw becomes 'queue a task that throws', and change all throws to also return too (this might be tricky because the parent will not know, so I think a) is better).

> It seems like there are more things besides Fetch that currently expect to run on the main thread, which is a little worrying. I'm not sure what the full implementation flow is in Chromium

The algorithm in Chromium is implemented in the browser process. Unfortunately there is no such equivalent in the spec? It seems that the closest thing is 'in parallel' but we are forced to go back to main for things that supposedly cannot be handled in parallel, but which we can do in the browser process.

> 2. While the way we're kicking off tasks for URL parsing and Fetching is technically correct, it kind of hints to me that there is a design issue. I think it's uncommon to make fetches where the URL is parsed _asynchronously_ with respect to the Fetching, and this can have observable consequences too (i.e., you're allowing the `<base>` URL of the document to change in between parsing and fetching, whereas I'm guessing that is not possible in the implementation?).

Yea that's true although the fetches are not relative to the document.

> I think (1) needs to be fixed regardless

Is there anything else you see besides the exceptions? We could do what I suggested if that's the only thing, but I may have missed other examples.

npm1:
By the way the exception thing is a problem coming from the caller as well as this already assumes you can throw exceptions from in parallel: https://w3c.github.io/webappsec-credential-management/#ref-for-dom-credential-discoverfromexternalsource-slot%E2%91%A0:~:text=Set%20result%20to%20the%20result%20of%20executing%20result%E2%80%99s

But thinking about it, why is that not allowed? They are thrown and caught from in parallel (and I'm assuming we're always carrying around the |globalObject| anyways).

npm1:
Answering to myself: at least DOMExceptions are Objects so they cannot be created from in parallel. I will return failure except at the end, and there 'queue a task to throw'.

npm1:
I think I fixed exception handling for the most part, let me know what you think!

spec/index.bs
```diff
@@ -1187,201 +1189,417 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
         1. Let |json| be the result of [=extract the JSON fetch response=] from |response|.
-        1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderAccountList}}, |potentialAccountsList|.
-        1. Set |accountsList| to |potentialAccountsList|.
+        1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderAccountList}}, and
+            store the result in |accountsList|.
+
+    Issue: We should validate the accounts list returned here for repeated ids, as described [here](https://github.com/fedidcg/FedCM/issues/336).
 
-        Note: This operation is not defined but should be removed when we [integrate](https://github.com/fedidcg/FedCM/issues/261) with Fetch.
+    1. Wait for |accountsList| to be set.
```
domfarolino, line 1423: Again at this point I don't think this variable exists yet, right? I think we have to set it to null early on and wait for it to be assigned to something, so we can reference it appropriately, even though I understand the intent of this.

npm1: I had it defined like this before and changed based on Sam's feedback :) Since you feel strongly about this, I have tried to ensure the variables we wait for are initialized outside of the task.

spec/index.bs
```diff
@@ -1187,201 +1189,417 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+        :: a [=list=] containing a single [=header=] with [=header/name=] set to `Accept` and
+            [=header/value=] set to `application/json`
+        :  [=request/referrer policy=]
+        :: "no-referrer"
+        :  [=request/credentials mode=]
+        :: "include"
+
+    Issue: The credentialed fetch in this algorithm can lead to a timing attack that leaks the user's identities before the user grants permission. This is an active area of investigation that is being explored [here](https://github.com/fedidcg/FedCM/issues/230#issuecomment-1089040953).
+
+    1. [=Fetch request=] with |request|, |globalObject|, and <var ignore=>processResponseConsumeBody</var>
```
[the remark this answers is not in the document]
npm1: I vehemently oppose manual numbering :) It seems issues/notes require an extra indentation compared with numbered items (that fixed it, here and in a couple other places).

npm1 (commented):
ptal

spec/index.bs
```diff
@@ -1187,201 +1189,425 @@ The {{CredentialRequestOptions/signal}} is used as an abort signal for the
...
+algorithm is invoked, the user agent MUST execute the following steps. This returns an
+{{IdentityCredential}} (or throws an error to the caller).
+
+    1. Assert: These steps are running [=in parallel=].
+    1. Assert: |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"] [=map/exists=].
+    1. Assert: |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"] [=list/size=] is 1.
+
+          Issue: Support choosing accounts from multiple [=Identity Provider=]s, as described [here](https://github.com/fedidcg/FedCM/issues/319).
+    1. Run {{setTimeout}} passing a [=task=] which throws a {{NetworkError}}, after a timeout of
+        60 seconds.
```
domfarolino, line 1203: I think the "network error" here should be a failure like you now have elsewhere, since this is all in parallel.

domfarolino (approved):
OK it looks like the exception problem is fixed now with the exception (pun intended) of one case I think? Otherwise, I think your "failure" propagation works nicely.

spec/index.bs
npm1: Agree but DiscoverFromExternalSource actually accepts exceptions being thrown at it somehow. I filed https://github.com/w3c/webappsec-credential-management/issues/211 since I'm not sure that's acceptable (plus need the `globalObject`). I suspect it is done that way so that error messages can be set by whoever implements the DiscoverFromExternalSource. So yes this method is still throwing exceptions in two places, but I propose waiting for the linked issue to be resolved before modifying.

domfarolino: That seems fine, while slightly unfortunate 👍

spec/index.bs
samuelgoto: I still feel that these variable assignments to "null" are hacky and there has to be a better way, but maybe we can take that to a separate discussion. For now, will resolve this, merge this PR and try to make this better asynchronoulsy.

samuelgoto (approved):

PR #385, opened 2022-12-16
Add an explainer for the sign-in status API

cbiesinger:


cbiesinger:
@bvandersloot-mozilla FYI

PR #391, opened 2023-01-11
clarify token validation is the RP's responsibility

### LINKED ISSUE #318: Need to define token validation steps for the RP
Right now the spec doesn't specifically mention the token validation steps for the RP very well. This could be very small and reference just a single set of claims that the spec defines and highly reference the JWT spec. Alternatively, this could be defined in a way that focuses primarily on the baseline requirements now and allow for an extension point that leaves space for the VC Data model to slot in as an extension to the token data model defined.

kdenhartog:
closes #318

***
<a href="https://pr-preview.s3.amazonaws.com/kdenhartog/FedCM/pull/391.html" title="Last updated on Jan 11, 2023, 1:30 AM UTC (18fe97b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/391/7011888...kdenhartog:18fe97b.html" title="Last updated on Jan 11, 2023, 1:30 AM UTC (18fe97b)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -971,7 +971,11 @@ Every {{IdentityProviderToken}} is expected to have members with the following s
 
 The content of the [=token=] is opaque to the user agent and can contain
 anything that the [=Identity Provider=] would like to pass to the
-[=Relying Party=] to facilitate the login.
+[=Relying Party=] to facilitate the login. For this reason the [=Relying Party=]
+is expected to be the party responsible for validating the [=token=] passed 
+along from the [=Identity Provider=] using the appropriate token validation 
+algorithms defined. One example of how this might be done is defined 
+in [[OIDC-Connect-Core#IDTokenValidation]].
 
 NOTE: For [=Identity Providers=], it is worth considering how
 [portable](https://github.com/fedidcg/FedCM/issues/314) accounts are.


asr-enid (commented):
I'm unsure if it makes sense to even delve into (validation etc.) this given the FedCM API leaves it open what is passed here, or just name this as an general example. The FedCM token could as well not be an ID token type piece of information, but rather something that resembles an auth code, access token, .... 

Keeping it more generic in pointing out that its the RPs role to leverage that token in general in a non FedCM related process makes sense I would think.

spec/index.bs
```diff
@@ -971,7 +971,11 @@ Every {{IdentityProviderToken}} is expected to have members with the following s
 
 The content of the [=token=] is opaque to the user agent and can contain
 anything that the [=Identity Provider=] would like to pass to the
-[=Relying Party=] to facilitate the login.
+[=Relying Party=] to facilitate the login. For this reason the [=Relying Party=]
```
kdenhartog, line 974:
```suggestion
[=Relying Party=] to facilitate the login. For this reason the [=Relying Party=]
is expected to be the party responsible for validating the [=token=] passed 
along and it's out of scope for the browser to validate the tokens.
```

kdenhartog: @asr-enid I see your point about this. What about this as a replacement text to make it more general?

npm1: Looks good although my only comment is that the RP not only 'validates' the token but rather uses it in some way, for instance to create an account for the user by extracting the relevant data from it (and the format of the data plus how to extract such data is out of scope of this spec).

kdenhartog: Ok, yeah I see your point, but not quite sure what the proper way to account for this flow is. 

In the sense of OAuth wouldn't the relying party essentially be the resource server hence the need for it to perform the validation? In some flows I do believe it passes the token back to the IDP for something like an ID_Token in which case the RP is not actually doing the validation the IDP would be. But in that case isn't the IDP actually the resource server and the authorization server and the third party service + browser would be acting as the client?

Just wanting to double check my assumption here so that we can figure out the right way to word it. @timcappalli would be good to get your help on this wording.

samuelgoto: This looks like a step forward in making a clarification. If there is a better variation, feel free to send another PR, for for the sake of making incremental forward progress, I'll merge this PR because I think it stands on its own.

PR #395, opened 2023-01-13
Add a proposal for Webdriver support

cbiesinger:


cbiesinger:
@bvandersloot-mozilla is something along these lines something Mozilla would/could implement?

bvandersloot-mozilla:
> @bvandersloot-mozilla is something along these lines something Mozilla would/could implement?

It seems reasonable, however I can't guarantee prioritization or when we could get to it.

PR #405, opened 2023-01-19
Fix small typo in JSON example

philsmart:


philsmart:
I did not consider this a substantive change (to require a CLA). You can just add the `}` without this PR.

npm1 (approved):
Thanks for the fix!

philsmart:
Thanks!

PR #408, opened 2023-01-25
Add an alternatives considered section for header syntax

cbiesinger:
Also renames signout to signout-all, to make it clearer that this should only be used when no more accounts remain signed in.

### DIFF
--- a/proposals/idp-sign-in-status-api.md
+++ b/proposals/idp-sign-in-status-api.md
@@ -41,12 +41,15 @@ settings page allowing the user to disable certain IDPs for use with FedCM.
 
 ```
 IdP-SignIn-Status: action=signin
-IdP-SignIn-Status: action=signout
+IdP-SignIn-Status: action=signout-all
 ```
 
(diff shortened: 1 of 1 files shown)


npm1 (approved):

yi-gu:
lgtm. agreed that we should go with the first option since IDPs are the pure source of truth of whether there are remaining accounts signed in while it's error-prone for browsers to main such a list.

proposals/idp-sign-in-status-api.md
```diff
@@ -41,12 +41,15 @@ settings page allowing the user to disable certain IDPs for use with FedCM.
...
 ```
 
 These headers can be sent on the toplevel load as well as subresources such as
 XMLHttpRequest (this is necessary for at least one IDP).
 
+The signout-all header should only be sent when no accounts remain signed in
+to the IDP, i.e. when this action has signed out all accounts or if this
+was the last/only account getting signed out.
 
 ### JS API (future)
```
samuelgoto, line 55: Just curious: does the JS API need to change too?

PR #410, opened 2023-01-26
Add spec about account pictures

npm1:
***
<a href="https://pr-preview.s3.amazonaws.com/npm1/FedCM/pull/410.html" title="Last updated on Jan 30, 2023, 9:29 PM UTC (6f5ec16)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/410/3ec5b3e...npm1:6f5ec16.html" title="Last updated on Jan 30, 2023, 9:29 PM UTC (6f5ec16)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -767,6 +767,7 @@ dictionary IdentityProviderAccount {
   required USVString name;
   required USVString email;
   USVString given_name;
+  USVString picture;
   sequence<USVString> approved_clients;
 };
 dictionary IdentityProviderAccountList {
@@ -785,6 +786,8 @@ Every {{IdentityProviderAccount}} is expected to have members with the following
     ::  The user's email address.
     :   <dfn>given_name</dfn>
     ::  The user's given name.
+    :   <dfn>picture</dfn>
+    ::  The account's picture.
     :   <dfn>approved_clients</dfn>
     ::  A list of [=RP=]s (in the form of [=Client ID=]s) this account is already registered with.
         Used in the [=request permission to sign-up=] to allow the [=IDP=] to control whether to show
@@ -1205,6 +1208,10 @@ failure.
     1. If |config| is failure, return failure.
     1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
         and |globalObject|.
+    1. For each |account| in |accountsList|:
+        1. If |account|["{{IdentityProviderAccount/picture}}"] is present,
+            [=fetch the account picture=] with |account| and |globalObject|.
+    1. Wait for the fetches initiated in the above for loop to be completed.
     1. If |accountsList|'s size is 1:
         1. Let |account| be |accountsList|[0].
         1. Let |accountState| be the result of running the [=compute account state=] algorithm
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1405,6 +1412,32 @@ To <dfn>fetch the accounts list</dfn> given an {{IdentityProviderAPIConfig}} |co
     1. Return |accountsList|.
 </div>
 
+<div algorithm>
+To <dfn>fetch the account picture</dfn> given an {{IdentityProviderAccount}} |account| and a
+|globalObject|, run the following steps:
+    1. Let |pictureRequest| be a new <a spec=fetch for=/>request</a> as follows:
+
+        :  [=request/url=]
+        :: |account|["{{IdentityProviderAccount/picture}}"]
```
samuelgoto, line 1421: Don't we do things like resolving relative URLs?

npm1: I didn't see any relative URL resolving in the implementation.

spec/index.bs
```diff
@@ -785,6 +786,8 @@ Every {{IdentityProviderAccount}} is expected to have members with the following
     ::  The user's email address.
     :   <dfn>given_name</dfn>
     ::  The user's given name.
+    :   <dfn>picture</dfn>
+    ::  The account's picture.
```
cbiesinger, line 790: "URL for the account's picture"?

spec/index.bs
```diff
@@ -1205,6 +1208,10 @@ failure.
     1. If |config| is failure, return failure.
     1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
         and |globalObject|.
+    1. For each |account| in |accountsList|:
+        1. If |account|["{{IdentityProviderAccount/picture}}"] is present,
+            [=fetch the account picture=] with |account| and |globalObject|.
```
cbiesinger, line 1213: Hmm I'm not sure we want to be so prescriptive. For example:
- If a login hint is present, we probably don't want to fetch all account pictures
- If a UA wants to show returning accounts first and new accounts behind a button, they probably don't want to fetch all pictures
etc

spec/index.bs
cbiesinger: should we allow them? (maybe no point... I assume IDPs will usually want to serve them from a CDN, so relative URL does not really benefit them)

spec/index.bs
```diff
@@ -1405,6 +1412,32 @@ To <dfn>fetch the accounts list</dfn> given an {{IdentityProviderAPIConfig}} |co
     1. Return |accountsList|.
 </div>
 
+<div algorithm>
+To <dfn>fetch the account picture</dfn> given an {{IdentityProviderAccount}} |account| and a
+|globalObject|, run the following steps:
+    1. Let |pictureRequest| be a new <a spec=fetch for=/>request</a> as follows:
```
cbiesinger, line 1418: Set destination to image, maybe?

spec/index.bs
npm1: * We haven't spec'd login hint.
* That's fair but we have to specify these fetches somewhere, and the user agent dialogs are shortly after this. I added a note and removed the wait to be more lenient

spec/index.bs
npm1: Done. Just to confirm, we do allow redirects in this fetch?

spec/index.bs
npm1: I don't have a strong opinion. Probably not very useful yea...

spec/index.bs
cbiesinger: Pretty sure the answer is yes

cbiesinger (approved):

npm1:
Let me know if you have other comment before I merge this @samuelgoto.

spec/index.bs
```diff
@@ -1205,10 +1208,19 @@ failure.
     1. If |config| is failure, return failure.
     1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
         and |globalObject|.
+    1. For each |account| in |accountsList|:
+        1. If |account|["{{IdentityProviderAccount/picture}}"] is present,
+            [=fetch the account picture=] with |account| and |globalObject|.
```
samuelgoto, line 1213: should we spec out how to deal with fetching errors here?

spec/index.bs
samuelgoto: are you sure we don't resolve relative urls in the implementation? that would catch me by surprise ...

samuelgoto (approved):
LGTM

Just clarifying questions, but otherwise this is good to merge.

spec/index.bs
npm1: We do not. The url is passed as-is to some generic browser process fetcher, which would not be able to resolve relative urls since it has no idea of what the origin would be.

spec/index.bs
npm1: I added some text that the user agent may use some generic placeholder image in that case (that seems to be what we do, i.e. the dialog still shows)

PR #422, opened 2023-02-07
Add clarification about state machine

npm1:
Fixes https://github.com/fedidcg/FedCM/issues/388

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1059,6 +1059,11 @@ the following properties:
         on this account after the last successful {{IdentityCredential}} creation with this account.
 </dl>
 
+Note: The [=state machine map=] allows multiple keys whose {{AccountState/registration state}} is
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1059,6 +1059,11 @@ the following properties:
         on this account after the last successful {{IdentityCredential}} creation with this account.
 </dl>
 
+Note: The [=state machine map=] allows multiple keys whose {{AccountState/registration state}} is
```
samuelgoto, line 1062: Should we make this more precise? Should this make this normative? If non-normative, why is important to make this note?

npm1: I just made the note based on the feedback on the issue. However, since we are leaning more towards being concise and not adding any fluff at all, especially in normative sections, I'm leaning towards closing this. I'll ask in the issue for now.

PR #423, opened 2023-02-07
Use Origin instead of Referer

npm1:
Fixes https://github.com/fedidcg/FedCM/issues/379
cc @bvandersloot-mozilla as not sure what will be the right way to ask the Fetch spec to set the Origin header vs not set it, in the new fetch mode introduced in FedCM.

***
<a href="https://pr-preview.s3.amazonaws.com/npm1/FedCM/pull/423.html" title="Last updated on Feb 9, 2023, 3:52 PM UTC (2b3a614)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/423/97004a2...npm1:2b3a614.html" title="Last updated on Feb 9, 2023, 3:52 PM UTC (2b3a614)">Diff</a>


samuelgoto (approved):
LGTM

@bvandersloot-mozilla LGTY?

spec/index.bs
```diff
@@ -319,7 +319,7 @@ The well-known file is fetched:
 
 (a) **without** cookies,
 (b) **with** the <a http-header>Sec-Fetch-Dest</a> header set to `webidentity`, and
-(c) **without** a [[RFC9110#field.referer|Referer]] header
```
bvandersloot-mozilla, line 322: Genuine question: do we want to include that the Referer also will not be sent here (and in similar lists)?

spec/index.bs
```diff
@@ -1191,12 +1188,10 @@ To <dfn>fetch an identity assertion</dfn> given an [=AccountState=] |accountStat
         :  [=request/destination=]
         :: "webidentity"
         :  [=request/origin=]
-        :: a unique [=opaque origin=]
+        :: [=RP=]'s [=/origin=] (TODO)
```
bvandersloot-mozilla, line 1191: This works for now. Suggestion: replace the `(TODO)` with an inline `ISSUE: ` to switch to `cors-mode: user-agent-no-cors` when available?

bvandersloot-mozilla (approved):
Looks pretty good. One suggestion and one question that can be resolved as you see fit.

spec/index.bs
npm1: There is already an issue noted for this. I updated the string to `user-agent-no-cors`

PR #425, opened 2023-02-08
Clarify how IdentityProvider is exposed

cbiesinger:
This also does not need to be a partial interface.


cbiesinger:
@bvandersloot-mozilla it looks like I can't request a review from you but can you take a look?

bvandersloot-mozilla (approved):
LGTM

PR #432, opened 2023-02-13
Add Identity Provider branding name as an optional field 

bvandersloot-mozilla:
This adds one optional field to the IDP Branding information to provide a name the browser can show to represent the IDP. This shouldn't replace the IDP Site or Origin, because of security properties, but can be useful for UI elements.

***
<a href="https://pr-preview.s3.amazonaws.com/bvandersloot-mozilla/FedCM/pull/432.html" title="Last updated on Feb 15, 2023, 3:51 PM UTC (b96d5bc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/432/10c8d09...bvandersloot-mozilla:b96d5bc.html" title="Last updated on Feb 15, 2023, 3:51 PM UTC (b96d5bc)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -669,6 +669,7 @@ dictionary IdentityProviderBranding {
   USVString background_color;
   USVString color;
   sequence<IdentityProviderIcon> icons;
+  USVString name;
 };
 
 dictionary IdentityProviderAPIConfig {
@@ -1076,6 +1077,8 @@ Its members have the following semantics:
     ::  [=color=] for text on [=IDP=] branded widgets.
     :   <dfn>icons</dfn>
     ::  A list of {{IdentityProviderIcon}} objects.
+    :   <dfn>name</dfn>
+    ::  A user-recognizable name for the [=IDP=].
 </dl>
 
 Note: The branding preferences are deliberately designed to be high level
@@ -1117,7 +1120,8 @@ For example:
     "icons": [{
       "url": "https://idp.example/icon.ico",
       "size": 25
-    }]
+    }],
+    "name": "IDP Example"
   }
 }
 ```


npm1:
This could be useful... but as you mention, the user agent is already forced to show the IDP site/origin. Is this field required in the UI Firefox has been crafting? I'm not sure if Chrome would use this just because of the attacker threat.

samuelgoto:
> This could be useful... 

+1. This seems like it would be useful to me too, but ...

> but as you mention, the user agent is already forced to show the IDP site/origin. 

... right, chrome's implementation strongly needs the site/origin to be shown (i.e. we could use `name` "in addition to" `url` but not "in replacement of" ).

> Is this field required in the UI Firefox has been crafting? I'm not sure if Chrome would use this just because of the attacker threat.

Are you actively intending to use `name`? If so, adding this property here seems fine, but if you aren't (and neither is chrome), then we'd want to avoid writing something that no implementer is planning to use.

bvandersloot-mozilla:
> > but as you mention, the user agent is already forced to show the IDP site/origin.
> 
> ... right, chrome's implementation strongly needs the site/origin to be shown (i.e. we could use `name` "in addition to" `url` but not "in replacement of" ).

Strong agree. Firefox has the same constraint.

> > Is this field required in the UI Firefox has been crafting? I'm not sure if Chrome would use this just because of the attacker threat.
> 
> Are you actively intending to use `name`? If so, adding this property here seems fine, but if you aren't (and neither is chrome), then we'd want to avoid writing something that no implementer is planning to use.

There are places we would like to include `name`, assuming it isn't something that Google thinks is too dangerous to include.

samuelgoto:
> assuming it isn't something that Google thinks is too dangerous to include.

We can ask around to confirm, but I doubt that we'd arrive at a different position: seems safe (maybe comparable to `<title>`: content-provided string that is used in conjunction with the `origin`). If this is something that Firefox is planning to use, I think we should merge (and even if Chrome doesn't at first, I think we would at some point).

@npm1 wdyt?

bvandersloot-mozilla:
Rebase'D!

yi-gu:
Just want to confirm that the feedback we got from Chrome stakeholders matches the analysis in this issue. i.e. it's OK to show IDP names but the domain must be displayed as well AND it has to be as prominent as the IDP name.

samuelgoto:
> Just want to confirm that the feedback we got from Chrome stakeholders matches the analysis in this issue. i.e. it's OK to show IDP names but the domain must be displayed as well AND it has to be as prominent as the IDP name.

Neat, thanks for confirming!

PR #433, opened 2023-02-13
Convert required email field in user account to optional

bvandersloot-mozilla:
This makes the email field of the IdentityProviderAccount optional. This can give a little flexibility to the IDP where accounts may not have an email address tied to them or be a useful way for the user to identify the account.

This patch is meant to land after https://github.com/fedidcg/FedCM/pull/424, and as such only the changes in https://github.com/fedidcg/FedCM/commit/3ed01b5bd473cb667eb584439e3033c4de236c72 are relevant to this PR.


npm1:
Is there an issue associated with this PR? Also what use-case is this trying to solve? In Chrome we currently do require the email. If it was not required, I guess we could show the name plus account ID or just the name when the email is not present. But it's worth thinking about how the user agent would be able to present enough information to the user so they can pick their account correctly.

samuelgoto:
> Is there an issue associated with this PR? Also what use-case is this trying to solve? In Chrome we currently do require the email. If it was not required, I guess we could show the name plus account ID or just the name when the email is not present. But it's worth thinking about how the user agent would be able to present enough information to the user so they can pick their account correctly.

Yeah, this is indeed normative, in the sense that it is, by choice, that we wrote email as a `required` field.

We are certainly interested in making this not the case, but we should think this through.

bvandersloot-mozilla:
> Is there an issue associated with this PR? 

Not currently. I'll file one and we can have the discussion there. (#435)

>Also what use-case is this trying to solve? 

per the PR, "This can give a little flexibility to the IDP where accounts may not have an email address tied to them or be a useful way for the user to identify the account."

samuelgoto:
> Is there an issue associated with this PR?

There are at least two issues that are somewhat related.

One of the them, came up in our TAG reviews:

https://github.com/w3ctag/design-reviews/issues/718#issuecomment-1171733526

The other is this one here:

https://github.com/fedidcg/FedCM/issues/242#issuecomment-1146493676

In both of these issues, my intuition / suggestion is that we should give RPs the ability to select what they need to be disclosed (e.g. optionality of email) and have the browser be able to honor that (the idp also has a say in what can be optional and what can't).

I don't know exactly what the API looks like, but here is what I responded to the TAG when that came up:

```js
await navigator.credentials.get({
      federated: {
        claims: {
          name: {essential: true}, // required
          email: null, // optional
          emailVerified: null, // optional
          picture: {essential: true}, // required
        },
        providers: [{
          url: "https://idp.example",
          clientId: "123"
        }]
      }
});
```

WDYT?

PR #436, opened 2023-02-14
Add the IDP sign-in status API to the spec

### LINKED ISSUE #430: Finalize naming for the IDP sign-in status API
https://github.com/fedidcg/FedCM/blob/main/proposals/idp-sign-in-status-api.md

We currently use these names:
```
IdP-SignIn-Status: action=signin
IdP-SignIn-Status: action=signout-all

interface IdentityProvider {
  static void login();
  static void logout();
}
```

We should make sure these are the names we want to use. I believe @bvandersloot-mozilla / @martinthomson  had thoughts on this.

cbiesinger:
Explainer: https://github.com/fedidcg/FedCM/blob/main/proposals/idp-sign-in-status-api.md

Fixes #430, #447

***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/436.html" title="Last updated on Oct 13, 2023, 3:28 PM UTC (48e4d1a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/436/273ccb0...cbiesinger:48e4d1a.html" title="Last updated on Oct 13, 2023, 3:28 PM UTC (48e4d1a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -40,6 +40,9 @@ spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
         text: no such alert; url: dfn-no-such-alert
         text: error code; url: dfn-error-code
         text: validating capabilities; url: dfn-validate-capabilities
+spec: webappsec-fetch-metadata; urlPrefix: https://w3c.github.io/webappsec-fetch-metadata/
+    type: dfn
+        text: Directly User-Initiated Requests; url: directly-user-initiated
 </pre>
 
 <pre class=link-defaults>
@@ -344,6 +347,8 @@ value |value|:
 
 ### HTTP header API ### {#login-status-http}
 
+[=IDPs=] can set the login status using an HTTP [=response=] [=header=] as follows.
+
 Issue: The HTTP header checking should move into the Fetch spec, since it
     affects all resource loads.
 
@@ -353,14 +358,19 @@ list with name "<dfn><code>Set-Login</code></dfn>" and type "`item`". If |value|
 process this header as follows:
 
 <div algorithm="process the login status header">
-* Let |origin| be the response's [=response/URL=]'s [=/origin=].
-* If |origin| is not [=same-origin with its ancestors=], ignore the header.
-* Otherwise:
-    * Assert that |value| is a tuple.
-    * Let |token| be the first entry of |value|.
-    * If |token| is `logged-in`, [=set the login status=] for |origin|
+1. Let |origin| be the response's [=response/URL=]'s [=/origin=].
+1. Let |client| be the [=/request=]'s [=request/client=].
+1. If the request is not a [=directly user-initiated request=]:
+    1. If |client| is null, return.
+    1. If |origin| is not [=same origin=] with the [=/request=]'s
+        [=request/origin=], return.
+    1. If |client| is not [=same-origin with its ancestors=], return.
+1. Otherwise:
+    1. Assert that |value| is a tuple.
+    1. Let |token| be the first entry of |value|.
+    1. If |token| is `"logged-in"`, [=set the login status=] for |origin|
         to [=logged-in=].
-    * If |token| is `logged-out`, [=set the login status=] for |origin|
+    1. If |token| is `"logged-out"`, [=set the login status=] for |origin|
         to [=logged-out=].
 
 </div>
@@ -388,10 +398,10 @@ partial interface Navigator {
 
 <div algorithm="setStatus">
 When {{NavigatorLogin/setStatus()}} is called with argument |status|:
+1. If the [=current settings object=] is not [=same-origin with its ancestors=],
+    throw a {{SecurityError}} {{DOMException}}.
 1. Let |origin| be the [=current settings object=]'s
     [=environment settings object/origin=].
-1. If |origin| is not [=same-origin with its ancestors=], throw a
-    {{SecurityError}} {{DOMException}}.
 1. Let |value| be [=logged-in=] if |status| is `"logged-in"` or [=logged-out=]
     if |status| is `"logged-out"`.
 1. [=Set the login status=] for |origin| to |value|.
@@ -403,13 +413,15 @@ When {{NavigatorLogin/setStatus()}} is called with argument |status|:
 User agents MUST also clear the [=Login Status map=] data when:
     :   the user clears all cookies or site settings data
     ::  The user agent MUST clear the entire map.
-    :   the user clears cookies or site data for a specific origin
+    :   the user clears all cookies or all site data for a specific origin
     ::  The user agent MUST remove all entries that would be affected
             by the deleted cookies, that is, any entry with an origin
             to which a deleted cookie could be sent to.
 
             Note: For example, domain cookies may affect subdomains of
-                the deleted origin.
+                the deleted origin, e.g. clearing cookies for `google.com`
+                should also reset the login status for `accounts.google.com`,
+                since it may rely on a domain cookie for google.com.
     :   the user deletes individual cookies (if allowed by the user agent)
     ::  the behavior is user agent-defined.
 
@@ -417,10 +429,15 @@ User agents MUST also clear the [=Login Status map=] data when:
                 since is impossible to know whether this cookie affects
                 authorization state.
     : the user agent receives a <a http-header>Clear-Site-Data</a> header with a
-        value of `"cookies"` or `"*"`
+        value of `"cookies"` or `"*"`, and the [=/request=]'s [=request/client=] is
+        not null, and the client's [=environment settings object/origin=] is [=same
+        origin=] with the [=top-level origin=]
     :: while [$clear cookies for origin|clearing cookies for
         origin$] it MUST remove any entries in the [=Login Status Map=] where
         the [=map/key=] is the input origin.
+
+        Issue: Once Clear-Site-Data [supports partitioned cookies](https://github.com/w3c/webappsec-clear-site-data/issues/72),
+            this wording should be updated.
 
 Note: Other website-initiated cookie changes should not affect this map. When
     [=IDP=] login state changes, it should send an explicit [=Set-Login=] header.
@@ -679,8 +696,15 @@ the exception thrown.
         * Return (failure, false).
         * Prompt the user whether to continue. If the user continues, the user
             agent SHOULD set |loginStatus| to [=unknown=]. This MAY include an
-            affordance to [=show an IDP login dialog=]. If the user cancels
-            or that algorithm returns failure, return (failure, true).
+            affordance to [=show an IDP login dialog=].
+
+            * If the user cancels this dialog, return (failure, true).
+            * If the user triggers this affordance:
+                1. Let |config| be the result of running [=fetch the config file=]
+                    with |provider| and |globalObject|.
+                1. If |config| is failure, return (failure, true).
+                1. [=Show an IDP login dialog=] with |config|.
+                1. If that algorithm returns failure, return (failure, true).
 
         Issue: We should perhaps provide a way to let the [=RP=] request that
             the second option is provided, possibly gated on a user gesture.
@@ -689,7 +713,7 @@ the exception thrown.
         [=requires user mediation=].
     1. Let |mediation| be |options|'s {{CredentialRequestOptions/mediation}}.
     1. If |requiresUserMediation| is true and |mediation| is
-        "{{CredentialMediationRequirement/silent}}", return (failure, false).
+        "{{CredentialMediationRequirement/silent}}", return (failure, true).
     1. Let |config| be the result of running [=fetch the config file=] with
         |provider| and |globalObject|.
     1. If |config| is failure, return (failure, false).
@@ -709,8 +733,8 @@ the exception thrown.
         1. <dfn>Mismatch dialog step</dfn>: If |loginStatus| is [=logged-in=], show a
             dialog to the user. The contents of this dialog are defined by the user
             agent. This dialog SHOULD provide an affordance for the user to trigger
-            the [=show an IDP login dialog=] algorithm; this dialog is the
-            <dfn>confirm IDP login dialog</dfn>.
+            the [=show an IDP login dialog=] algorithm with |config|; this dialog
+            is the <dfn>confirm IDP login dialog</dfn>.
 
             Note: This situation happens when the browser expects the user
                 to be signed in, but the accounts fetch indicated that the user
@@ -733,7 +757,7 @@ the exception thrown.
                     1. Otherwise, go back to the [=fetch accounts list step=].
 
     1. Assert: |accountsList| is not failure and the size of |accountsList| is not 0.
-    1. [=Set the login status=] for the [=/origin] of the
+    1. [=Set the login status=] for the [=/origin=] of the
         {{IdentityProviderConfig/configURL}} to [=logged-in=].
     1. If |provider|'s {{IdentityProviderConfig/loginHint}} is not empty:
         1. For every |account| in |accountList|, remove |account| from |accountList| if |account|'s
@@ -761,7 +785,7 @@ the exception thrown.
             [=compute the connection status=] algorithm given |provider| and |account|. When doing this,
             the user agent MAY show some UI to the user indicating that they are being
             <dfn>auto-reauthenticated</dfn>.
-    1. Otherwise, if |mediation| is "{{CredentialMediationRequirement/silent}}", return (failure, false).
+    1. Otherwise, if |mediation| is "{{CredentialMediationRequirement/silent}}", return (failure, true).
     1. Otherwise, if |accountsList|'s size is 1:
         1. Set |account| to |accountsList|[0].
         1. Set |accountState| to the result of running the [=compute the connection status=] algorithm
@@ -934,7 +958,7 @@ dictionary IdentityProviderAPIConfig {
   required USVString accounts_endpoint;
   required USVString client_metadata_endpoint;
   required USVString id_assertion_endpoint;
-  USVString login_url;
+  required USVString login_url;
   IdentityProviderBranding branding;
 };
 </xmp>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -345,7 +345,7 @@ Issue: Should the header checking be integrated into the fetch spec instead,
 
 User agents must also clear the [=IDP Sign-in Status map=] data when:
     :   the user clears all cookies or site settings data
-    ::  The user agent must clear the entire map
+    ::  The user agent must clear the entire map.
     :   the user clears cookies or site data for a specific origin
     ::  The user agent must remove all entries that would be affected
             by the deleted cookies.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -348,7 +348,8 @@ User agents must also clear the [=IDP Sign-in Status map=] data when:
     ::  The user agent must clear the entire map.
     :   the user clears cookies or site data for a specific origin
     ::  The user agent must remove all entries that would be affected
-            by the deleted cookies.
+            by the deleted cookies, that is, any entry with an origin
+            to which a deleted cookie could be sent to.
 
             Note: For example, domain cookies may affect subdomains of
                 the deleted origin.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -678,8 +678,9 @@ the exception thrown.
 
             Note: Since this situation happens when the browser expects the user
                 to be signed in, but the accounts fetch indicated that the user
-                is signed out, it is recommended that this dialog provides a
-                "sign in" button allowing the user to sign in to the IDP using
+                is signed out, it is recommended that this dialog is a <dfn>confirm
+                IDP signin dialog</dfn> and provides a "sign in" button allowing
+                the user to sign in to the IDP using
                 {{IdentityProviderAPIConfig/signin_url}}.
 
             Note: This dialog ensures that silent tracking of the user is
@@ -1874,6 +1875,36 @@ The [=remote end steps=] are:
 
 1. Return [=success=] with data `null`.
 
+## Confirm IDP signin ## {#webdriver-confirmidpsignin}
+
+<figure id="table-webdriver-canceldialog" class="table">
+    <table class="data">
+        <thead>
+            <tr>
+                <th>HTTP Method</th>
+                <th>URI Template</th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr>
+                <td>POST</td>
+                <td>`/session/{session id}/fedcm/confirm_idp_signin`</td>
+            </tr>
+        </tbody>
+    </table>
+</figure>
+
+The [=remote end steps=] are:
+
+1. If no FedCM dialog is currently open, or the dialog is not a [=confirm IDP
+    signin dialog=] return a [=error|WebDriver error=] with [=error code=]
+    [=no such alert=].
+
+1. Act as if the user had clicked the "continue" button in the dialog and
+    initiate the signin flow.
+
+1. Return [=success=] with data `null`.
+
 ## Account list ## {#webdriver-accountlist}
 
 <figure id="table-webdriver-accountlist" class="table">
@@ -1985,7 +2016,8 @@ The [=remote end steps=] are:
     [=error code=] [=no such alert=].
 
 1. Let |type| be a string that is "`AutoReauthn`" if the user is being [=auto-reauthenticated=],
-    or "`AccountChooser`" otherwise.
+    or "`AccountChooser`" is the dialog is an account chooser, or "`ConfirmIdpSignin`" if the
+    dialog is a [=confirm IDP signin dialog=].
 
 1. Return [=success=] with data |type|.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -319,20 +319,18 @@ for an [=IDP=] to adopt the FedCM API. It doesn't introduce security issues on t
 ## The Login Status API ## {#browser-api-login-status}
 <!-- ============================================================ -->
 
-Issue: Reconcile this section with the
-    [Login Status API](https://github.com/privacycg/is-logged-in).
-    See also [this PR](https://github.com/privacycg/is-logged-in/pull/54).
-
-### HTTP header API ### {#login-status-http}
-
-Issue: The HTTP header checking should move into the Fetch spec, since it
-    affects all resource loads.
+### Login Status Map ### {#hdr-login-status-map}
 
 Each [=user agent=] keeps a global, persistent <dfn>Login Status
 map</dfn>, an initially empty [=map=]. The [=map/keys=] in this map are
 [=/origin=] (of [=IDPs=]), and the [=map/values=] are enums that can be one of
 "<dfn><code>unknown</code></dfn>", "<dfn><code>logged-in</code></dfn>",
-and "<dfn><code>logged-out</code></dfn>"
+and "<dfn><code>logged-out</code></dfn>".
+
+### HTTP header API ### {#login-status-http}
+
+Issue: The HTTP header checking should move into the Fetch spec, since it
+    affects all resource loads.
 
 For each [=http-redirect fetch=] and [=http fetch=]'s [=response=], let |value|
 be the result of [=get a structured field value=] from the response's header
@@ -340,17 +338,15 @@ list with name "<dfn><code>Set-Login</code></dfn>" and type "`item`". If |value|
 process this header as follows:
 
 <div algorithm="process the login status header">
-* If the request's [=request/current url=]'s [=/origin=] is not [=same origin=]
-    with the [=environment/top-level origin=], ignore the header.
+* Let |origin| be the response's [=response/URL=]'s [=/origin=].
+* If |origin| is not [=same-origin with its ancestors=], ignore the header.
 * Otherwise:
     * Assert that |value| is a tuple.
     * Let |token| be the first entry of |value|.
-    * If |token| is `login`, [=map/set=] an entry in the
-        [=Login Status map=] with the key being the origin of the resource
-        and the value being [=logged-in=].
-    * If |token| is `logout`, [=map/set=] an entry in the
-        [=Login Status map=] with the key being the origin of the resource
-        and the value being [=logged-out=].
+    * If |token| is `logged-in`, [=map/set=] [=Login Status map=][|origin|]
+        to [=logged-in=].
+    * If |token| is `logged-out`, [=map/set=] [=Login Status map=][|origin|]
+        to [=logged-out=].
 
 </div>
 
@@ -377,23 +373,23 @@ partial interface Navigator {
 
 <div algorithm="setStatus">
 When {{NavigatorLogin/setStatus()}} is called with argument |status|:
-1. If the [=/origin=] of the [=current settings object=] is not [=same origin=]
-    with the [=top-level origin=], throw a {{SecurityError}} {{DOMException}}.
+1. Let |origin| be the [=current settings object=]'s
+    [=environment settings object/origin=].
+1. If |origin| is not [=same-origin with its ancestors=], throw a
+    {{SecurityError}} {{DOMException}}.
 1. Let |value| be [=logged-in=] if |status| is `"logged-in"` or [=logged-out=]
     if |status| is `"logged-out"`.
-1. [=map/set|Set=] the entry in the [=Login Status Map=] with the [=map/key=]
-    being [=/this=]'s [=/relevant settings object=]'s
-    [=environment settings object/origin=] and the value being |value|.
+1. [=map/set|Set=] [=Login Status Map=][|origin|] to |value|.
 
 </div>
 
-### Clearing map data ### {#login-status-clear-data}
+### Clearing the Login Status Map data ### {#login-status-clear-data}
 
-User agents must also clear the [=Login Status map=] data when:
+User agents MUST also clear the [=Login Status map=] data when:
     :   the user clears all cookies or site settings data
-    ::  The user agent must clear the entire map.
+    ::  The user agent MUST clear the entire map.
     :   the user clears cookies or site data for a specific origin
-    ::  The user agent must remove all entries that would be affected
+    ::  The user agent MUST remove all entries that would be affected
             by the deleted cookies, that is, any entry with an origin
             to which a deleted cookie could be sent to.
 
@@ -402,13 +398,13 @@ User agents must also clear the [=Login Status map=] data when:
     :   the user deletes individual cookies (if allowed by the user agent)
     ::  the behavior is user agent-defined.
 
-            Note: The user agent may want to reset the state to [=unknown=],
+            Note: The user agent MAY want to reset the state to [=unknown=],
                 since is impossible to know whether this cookie affects
                 authorization state.
     : the user agent receives a <a http-header>Clear-Site-Data</a> header with a
-        value of `"cookies"`
+        value of `"cookies"` or `"*"`
     :: while [$clear cookies for origin|clearing cookies for
-        origin$] it must remove any entries in the [=Login Status Map=] where
+        origin$] it MUST remove any entries in the [=Login Status Map=] where
         the [=map/key=] is the input origin.
 
 Note: Other website-initiated cookie changes should not affect this map. When
@@ -665,7 +661,7 @@ the exception thrown.
         {{IdentityProviderConfig/configURL}}. If there is no such entry, set it
         to a user-agent specific default (either [=unknown=] or
         [=logged-out=]).
-    1. If |loginStatus| is [=logged-out=], the user agent must do one of the following:
+    1. If |loginStatus| is [=logged-out=], the user agent MUST do one of the following:
 
         * Return (failure, false).
         * Prompt the user whether to continue. If the user continues, the user
@@ -681,8 +677,8 @@ the exception thrown.
     1. Let |mediation| be |options|'s {{CredentialRequestOptions/mediation}}.
     1. If |requiresUserMediation| is true and |mediation| is
         "{{CredentialMediationRequirement/silent}}", return (failure, false).
-    1. <dfn>Fetch config step</dfn>: Let |config| be the result of running
-        [=fetch the config file=] with |provider| and |globalObject|.
+    1. Let |config| be the result of running [=fetch the config file=] with
+        |provider| and |globalObject|.
     1. If |config| is failure, return (failure, false).
     1. <dfn>Fetch accounts list step</dfn>: Let |accountsList| be the result of
         [=fetch the accounts list=] with |config|, |provider|, and |globalObject|.
@@ -721,9 +717,7 @@ the exception thrown.
                     1. If |result| is failure, return (failure, true). The user
                         agent MAY show a dialog to the user before or after
                         returning failure indicating this failure.
-                    1. Otherwise, go back to the [=fetch config step=]. As an
-                        optimization, the user agent MAY instead go to the
-                        [=fetch accounts list step=].
+                    1. Otherwise, go back to the [=fetch accounts list step=].
 
     1. Assert: |accountsList| is not failure and the size of |accountsList| is not 0.
     1. [=map/Set=] an entry in the [=login status map=] with the key being the
@@ -892,6 +886,9 @@ or failure.
         1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderAPIConfig}} stored
             in |config|.
         1. If one of the previous two steps threw an exception, set |config| to failure.
+        1. Set |config|.{{IdentityProviderAPIConfig/login_url}} to the result of  [=computing
+            the manifest URL=] with |provider|, |config| and |globalObject|.
+        1. If |config|.{{IdentityProviderAPIConfig/login_url}} is null, return failure.
     1. Wait for both |config| and |configInWellKnown| to be set.
     1. If |configInWellKnown| is true, return |config|. Otherwise, return failure.
 </div>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -360,25 +360,27 @@ IDPs can also use a JavaScript API to update the stored sign-in status:
 
 
 <pre class="idl">
-  dictionary SignInStatusOptions {
-    boolean idp;
-  };
+enum LoginStatus {
+  "logged-in",
+  "logged-out",
+};
 
-  partial interface Navigator {
-    static Promise&lt;undefined&gt; recordSignIn(optional SignInStatusOptions options = {});
-    static Promise&lt;undefined&gt; recordSignOut(optional SignInStatusOptions options = {});
-  };
+[Exposed=Window, SecureContext] 
+interface NavigatorLogin {
+  Promise&lt;undefined&gt; setStatus(LoginStatus status);
+};
+
+partial interface Navigator {
+  [SecureContext] readonly attribute NavigatorLogin login;
+};
 </pre>
 
-When {{Navigator/recordSignIn}} or {{Navigator/recordSignOut}} is
-called:
-  * If no |options| dictionary is provided, or there is no |idp| member in the
-    |options|, or the |idp| member is false, then the promise is rejected with
-    a {{TypeError}}.
-  * Otherwise, the user agent [=map/set|sets=] the entry in the [=Sign-in Status
+When {{NavigatorLogin/setStatus()}} is called with argument |status|:
+* Let |value| be [=signed-in=] if |status| is `"logged-in"` or [=signed-out=]
+    if |status| is `"logged-out"`.
+* Otherwise, the user agent [=map/set|sets=] the entry in the [=Sign-in Status
     Map=] with the [=map/key=] being [=/this=]'s [=/relevant settings object=]'s
-    [=environment settings object/origin=] and the value being [=signed-in=] (for
-    `recordSignIn`) or [=signed-out=] (for `recordSignOut`).
+    [=environment settings object/origin=] and the value being |value|.
 
 ### Clearing map data ### {#signin-status-clear-data}
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -546,6 +546,9 @@ failure.
         no such entry, set it to a user-agent specific default (either [=unknown=] or
         [=signed-out=]).
     1. If |signinStatus| is [=signed-out=], return failure.
+        Issue: Should the user agent instead show UI to let the user sign
+            in to the IDP? This could be especially useful in the multi-IDP
+            case.
     1. Let |config| be the result of running [=fetch the config file=] with |provider| and
         |globalObject|.
     1. If |config| is failure, return failure.
@@ -564,8 +567,10 @@ failure.
         1. If |signinStatus| is [=signed-in=], show a dialog to the user and return
             failure.
 
-            Note: It is recommended that this dialog provides a "sign in" button
-                allowing the user to sign in to the IDP using
+            Note: Since this situation happens when the browser expects the user
+                to be signed in, but the accounts fetch indicated that the user
+                is signed out, it is recommended that this dialog provides a
+                "sign in" button allowing the user to sign in to the IDP using
                 {{IdentityProviderAPIConfig/signin_url}}.
 
             Note: This dialog ensures that silent tracking of the user is
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -315,35 +315,39 @@ for an [=IDP=] to adopt the FedCM API. It doesn't introduce security issues on t
 [=RP=] cannot inspect the results from the fetches in any way.
 
 <!-- ============================================================ -->
-## IDP Sign-in Status ## {#browser-api-idp-sign-in-status}
+## The Sign-in Status API ## {#browser-api-sign-in-status}
 <!-- ============================================================ -->
 
-Each [=user agent=] keeps a global, persistent <dfn>IDP Sign-in Status
+Each [=user agent=] keeps a global, persistent <dfn>Sign-in Status
 map</dfn>, an initially empty [=map=]. The [=map/keys=] in this map are
 [=/origin=] (of IDPs), and the [=map/values=] are enums that can be one of
 "<dfn><code>unknown</code></dfn>", "<dfn><code>signed-in</code></dfn>",
 and "<dfn><code>signed-out</code></dfn>"
 
 The user agent checks all page and subresource loads for an HTTP response
-[=header=] named <dfn><code>IdP-SignIn-Status</code></dfn>, whose value is
+[=header=] named <dfn><code>SignIn-Status</code></dfn>, whose value is
 parsed as HTTP [[RFC9110#section-5.6.6|Parameter]]s.
-If that header exists and any parameter has a name of `action`,
-the user agent must process it as follows:
+If that header exists, there is a parameter with name `type` and value `idp`,
+and any parameter has a name of `action`, the user agent must process it as
+follows:
 
+<div algorithm="process the signin status header">
 * If this is a subresource request, and the response does not have the
     same [=/origin] as the toplevel page, ignore the header.
 * Otherwise:
     * If the parameter value is `signin`, [=map/set=] an entry in the
-        [=IDP Sign-in Status map=] with the key being the origin of the resource
+        [=Sign-in Status map=] with the key being the origin of the resource
         and the value being [=signed-in=]
     * If the parameter value is `signout-all`, [=map/set=] an entry in the
-        [=IDP Sign-in Status map=] with the key being the origin of the resource
+        [=Sign-in Status map=] with the key being the origin of the resource
         and the value being [=signed-out=]
+
+</div>
 
 Issue: Should the header checking be integrated into the fetch spec instead,
     since it affects all resources?
 
-User agents must also clear the [=IDP Sign-in Status map=] data when:
+User agents must also clear the [=Sign-in Status map=] data when:
     :   the user clears all cookies or site settings data
     ::  The user agent must clear the entire map.
     :   the user clears cookies or site data for a specific origin
@@ -362,17 +366,38 @@ User agents must also clear the [=IDP Sign-in Status map=] data when:
 
 Note: Website-initiated cookie changes should not affect this map. When
     IDP sign-in state changes, it should send an explicit
-    [=IDP-SignIn-Status=].
+    [=SignIn-Status=].
     RP state should not affect this map since it only reflects IDP state.
 
-IDPs can also use a JavaScript API to update the stored sign-in status, see
-{{IdentityProvider}}.
+IDPs can also use a JavaScript API to update the stored sign-in status:
 
-### Interaction with `Clear-Site-Data`
+
+<pre class="idl">
+  dictionary SignInStatusOptions {
+    boolean idp;
+  };
+
+  partial interface Navigator {
+    static Promise&lt;undefined&gt; recordSignIn(optional SignInStatusOptions options);
+    static Promise&lt;undefined&gt; recordSignOut(optional SignInStatusOptions options);
+  };
+</pre>
+
+When {{Navigator/recordSignIn}} or {{Navigator/recordSignOut}} is
+called:
+  * If no |options| dictionary is provided, or there is no |idp| member in the
+    dictionary, or the |idp| member is false, then the promise is rejected with
+    a {{TypeError}}.
+  * Otherwise, the user agent [=map/set|sets=] the entry in the [=Sign-in Status
+    Map=] with the [=map/key=] being  [=/this=]'s [=/relevant settings object=]'s
+    [=environment settings object/origin=] and the value being [=signed-in=] (for
+    `recordSignIn`) or [=signed-out=] (for `recordSignOut`).
+
+### Interaction with `Clear-Site-Data` ### {#browser-clear-site-data}
 
 When the user agent receives a <a http-header>Clear-Site-Data</a> header with a
 value of `"cookies"`, while [$clear cookies for origin|clearing cookies for
-origin$] it must remove any entries in the [=IDP Sign-in Status Map=] where the
+origin$] it must remove any entries in the [=Sign-in Status Map=] where the
 [=map/key=] is the input origin.
 
 <!-- ============================================================ -->
@@ -620,7 +645,7 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
 or a pair (failure, bool), where the bool indicates whether to skip delaying
 the exception thrown.
     1. Assert: These steps are running [=in parallel=].
-    1. Let |signinStatus| be the value of the entry in [=IDP Sign-in Status map=]
+    1. Let |signinStatus| be the value of the entry in [=Sign-in Status map=]
         with the key being the origin of |provider|'s configURL. If there is
         no such entry, set it to a user-agent specific default (either [=unknown=] or
         [=signed-out=]).
@@ -639,7 +664,7 @@ the exception thrown.
     1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
         and |globalObject|.
     1. If the fetch failed, or the size of |accountsList| is 0:
-        1. [=map/set|Set=] an entry in the [=IDP Sign-in Status map=] with the
+        1. [=map/set|Set=] an entry in the [=Sign-in Status map=] with the
             key being the origin of the configURL and the value being [=signed-out=].
             A user agent may decide to skip this step if no credentials were sent to
             server.
@@ -660,9 +685,9 @@ the exception thrown.
             Note: This dialog ensures that silent tracking of the user is
                 impossible by always showing UI of some kind when credentials
                 were sent to the server.
-    1. If |signinStatus| is [=unknown=], [=map/set=] an entry in the [=IDP
-        sign-in status map=] with the key being the origin of the configURL and
-        the value being [=signed-in=].
+    1. If |signinStatus| is [=unknown=], [=map/set=] an entry in the [=sign-in
+        status map=] with the key being the origin of the configURL and the value being
+        [=signed-in=].
     1. If |provider|'s {{IdentityProviderConfig/loginHint}} is not empty:
         1. For every |account| in |accountList|, remove |account| from |accountList| if |account|'s
             {{IdentityProviderAccount/login_hints}} does not [=list/contain=] |provider|'s
@@ -1234,26 +1259,12 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
   };
 
   [Exposed=Window, SecureContext] interface IdentityProvider {
-      static undefined recordSignIn();
-      static undefined recordSignOut();
       static Promise&lt;sequence&lt;IdentityUserInfo&gt;&gt; getUserInfo(IdentityProviderConfig config);
   };
 </pre>
 
 Issue: [Decide](https://github.com/fedidcg/FedCM/issues/476) whether {{IdentityProvider}} is the
 correct location for the {{IdentityProvider/getUserInfo()}} method.
-
-Issue: Should the recordSign{In,Out} functions return `Promise<undefined>`, so the caller can know
-    when the call is finished?
-
-Issue: [Reconcile](https://github.com/speced/bikeshed/issues/2636) how these
-    record* functions relate to the isLoggedIn API.
-
-When {{IdentityProvider/recordSignIn}} or {{IdentityProvider/recordSignOut}} is
-called, the user agent [=map/set|sets=] the entry in the [=IDP Sign-in Status
-Map=] with the [=map/key=] being  [=/this=]'s [=/relevant settings object=]'s
-[=environment settings object/origin=] and the value being [=signed-in=] (for
-`recordSignIn`) or [=signed-out=] (for `recordSignOut`).
 
 An {{IdentityUserInfo}} represents user account information from a user. This information is exposed
 to the [=IDP=] once the user has already used the FedCM API to login in the [=RP=]. That is, it is
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -347,7 +347,7 @@ User agents must also clear the [=IDP Sign-in Status map=] data when:
     :   the user clears all cookies or site settings data
     ::  The user agent must clear the entire map
     :   the user clears cookies or site data for a specific origin
-    ::  The user agent must remote all entries that would be affected
+    ::  The user agent must remove all entries that would be affected
             by the deleted cookies.
 
             Note: For example, domain cookies may affect subdomains of
@@ -364,7 +364,8 @@ Note: Website-initiated cookie changes should not affect this map. When
     [=IDP-SignIn-Status=].
     RP state should not affect this map since it only reflects IDP state.
 
-IDPs can also use a JavaScript API to update the stored sign-in, see {{IdentityProvider}}.
+IDPs can also use a JavaScript API to update the stored sign-in status, see
+{{IdentityProvider}}.
 
 ### Interaction with `Clear-Site-Data`
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -333,7 +333,7 @@ follows:
 
 <div algorithm="process the signin status header">
 * If this is a subresource request, and the response does not have the
-    same [=/origin] as the toplevel page, ignore the header.
+    same [=/origin=] as the toplevel page, ignore the header.
 * Otherwise:
     * If the parameter value is `signin`, [=map/set=] an entry in the
         [=Sign-in Status map=] with the key being the origin of the resource
@@ -360,7 +360,7 @@ User agents must also clear the [=Sign-in Status map=] data when:
     :   if the user agent allows deleting individual cookies
     ::  the behavior is user agent-defined.
 
-            Note: The user agent may want to reset the state to undefined,
+            Note: The user agent may want to reset the state to [=unknown=],
                 since is impossible to know whether this cookie affects
                 authorization state.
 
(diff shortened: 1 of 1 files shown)


cbiesinger:
@bvandersloot-mozilla , can you take a look and see if this makes sense?

samuelgoto (commented):
Looks roughly right to me. Just some minor editorial suggestions, but the algorithm seems right.

martinthomson (changes requested):
State clearing?

spec/index.bs
```diff
@@ -301,6 +301,30 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+## IDP sign-in status ## {#browser-api-idp-sign-in-status}
+<!-- ============================================================ -->
+
+Each [=user agent=] keeps a global, persistent <dfn>IDP sign-in status
+map</dfn>, an initially empty [=map=]. The [=map/keys=] in this map are
+[=/origin=] (of IDPs), and the [=map/values=] are enums that can be one of
+"unknown", "signed-in", and "signed-out".
+
+The user agent checks all page and subresource loads for an HTTP response
+header named `IdP-SignIn-Status` containing [[RFC9110#parameter|Parameter]]s.
```
npm1, line 314: I think header response names need to be defined. Other than that this checking of all headers in everything sounds like something that needs to be integrated into the Fetch spec (this is a monkeypatch). Perhaps note that as an issue.

spec/index.bs
```diff
@@ -301,6 +301,30 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+If that header exists and the first parameter has a name of `action`,
```
npm1, line 315: Presumably this needs to be parsed? Can we perhaps encapsulate this in an algorithm

spec/index.bs
```diff
@@ -301,6 +301,30 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+the user agent must process it as follows:
+
+* If this is a subresource request, and the subresource does not have the
+    same [=/origin] as the toplevel page, ignore the header.
```
npm1, line 319: What do you mean by the origin of the subresource? Presumably the final origin? I think this needs to be pulled from the response itself (I imagine the request may only have the initial origin)

spec/index.bs
```diff
@@ -301,6 +301,30 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+* Otherwise:
+    * If the parameter value is `signin`, add (or replace) an entry in the
```
npm1, line 321: Use https://infra.spec.whatwg.org/#map-set

spec/index.bs
```diff
@@ -301,6 +301,30 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+        [=IDP sign-in status map=] with the key being the origin of the resource
+        and the value being "signed-in"
+    * If the parameter value is `signout-all`, add (or replace) an entry in the
```
npm1, line 324: ditto

spec/index.bs
```diff
@@ -301,6 +301,30 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
 [=RP=] cannot inspect the results from the fetches in any way.
 
+<!-- ============================================================ -->
+## IDP sign-in status ## {#browser-api-idp-sign-in-status}
+<!-- ============================================================ -->
+
+Each [=user agent=] keeps a global, persistent <dfn>IDP sign-in status
+map</dfn>, an initially empty [=map=]. The [=map/keys=] in this map are
+[=/origin=] (of IDPs), and the [=map/values=] are enums that can be one of
+"unknown", "signed-in", and "signed-out".
```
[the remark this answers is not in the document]
bvandersloot-mozilla: "<dfn><code>unknown</code></dfn>" is even better

spec/index.bs
```diff
@@ -490,11 +514,28 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
 |provider| and a |globalObject|, run the following steps. This returns an {{IdentityCredential}} or
 failure.
     1. Assert: These steps are running [=in parallel=].
+    1. Let |signinStatus| be the value of the entry in [=IDP sign-in status map=]
+        with the key being the origin of |provider|'s configURL. If there is
+        no such entry, set it to a user-agent specific default (either "unknown" or
+        "signed-out").
+    1. If |signinStatus| is "signed-out", return failure.
```
bvandersloot-mozilla, line 521: Am I following it right that if a user is "signed out", any use of the API fails without a dialog?

spec/index.bs
```diff
@@ -490,11 +514,28 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
     1. Let |config| be the result of running [=fetch the config file=] with |provider| and
         |globalObject|.
     1. If |config| is failure, return failure.
     1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
         and |globalObject|.
+    1. If the fetch failed, or the size of |accountsList| is 0:
+        1. Add (or replace) an entry in the [=IDP sign-in status map=] with the
+            key being the origin of the configURL and the value being "signed-out".
+        1. If |signinStatus| is "signed-in", show a dialog to the user and return
```
bvandersloot-mozilla, line 530: does this mean that the sign-in status map and the actual state is out of sync at this point?

bvandersloot-mozilla:
@cbiesinger - what happened to the Javascript API?

spec/index.bs
cbiesinger: Done

cbiesinger (commented):
Martin: I added text that user agents should clear the sign-in state data when cookies or site data is cleared, is that what you meant? If not, please elaborate a bit on your concern

Ben: I was going to do that one separately, do you prefer if I add it to this PR?

bvandersloot-mozilla:
> Ben: I was going to do that one separately, do you prefer if I add it to this PR?
Yes, that would be great!

spec/index.bs
bvandersloot-mozilla: It would be nice to have a way to cleanly recover from a signed-out state without having to navigate to the identity provider independently. This has something to do with how to integrate this with the Multi-IDP patch, since I think their integration is non-trivial.

spec/index.bs
bvandersloot-mozilla: I was confused because showing a sign in link in the dialog when the status was signed-in is non-intuitive to me. I think this makes sense.

cbiesinger:
I've added the JS API, please take a look!

At a high-level, I think integration with #438 would happen by showing a "sign in" link for each IDP that we have a signed out (or mismatch) status for. I am not sure how to best handle the case when the user is signed out of all IDPs, since the API design so far assumes that no dialog would be shown in such a case.

I'm actually not sure how prescriptive the spec should be with regards to what a sign-in flow should look like, but this PR does add a signin_url field to the config manifest which can be used for that purpose.

spec/index.bs
cbiesinger: I added a "Issue:" for now. I think the multi-IDP PR is maybe a better place to discuss how to integrate these two?

spec/index.bs
cbiesinger: I reworded the note a bit to hopefully make this more clear.

spec/index.bs
```diff
@@ -301,6 +301,57 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+        and the value being [=signed-in=]
+    * If the parameter value is `signout-all`, [=map/set=] an entry in the
+        [=IDP Sign-in Status map=] with the key being the origin of the resource
+        and the value being [=signed-out=]
+
+Issue: Should the header checking be integrated into the fetch spec instead,
+    since it affects all resources?
+
+User agents must also clear the [=IDP Sign-in Status map=] when the user clears
+cookies or stored site data.
```
martinthomson, line 334: This isn't sufficiently precise for me.  Yes, if you nuke all state, this needs to go too, but there are other events that cause state for an origin to be removed.  What if only the IdP clears state?  What if only the RP clears state?

cbiesinger: I elaborated a bit. Not sure I understand the concern about RP state clearing especially but please let me know if the additional text addresses your concern?

spec/index.bs
```diff
@@ -330,8 +331,26 @@ the user agent must process it as follows:
...
+                the deleted origin.
+    :   if the user agent allows deleting individual cookies
+    ::  the behavior is user agent-defined.
+
+            Note: The user agent may want to reset the state to undefined,
+                since is impossible to know whether this cookie affects
+                authorization state.
+
+Note: Website-initiated cookie changes should not affect this map. When
+    IDP sign-in state changes, it should send an explicit
```
martinthomson, line 351: I don't think that this is true: https://w3c.github.io/webappsec-clear-site-data/

spec/index.bs
```diff
@@ -330,8 +331,26 @@ the user agent must process it as follows:
 Issue: Should the header checking be integrated into the fetch spec instead,
     since it affects all resources?
 
-User agents must also clear the [=IDP Sign-in Status map=] when the user clears
-cookies or stored site data.
+User agents must also clear the [=IDP Sign-in Status map=] data when:
+    :   the user clears all cookies or site settings data
+    ::  The user agent must clear the entire map
```
martinthomson, line 336: Missing a period, though I'm not sure whether this is supposed to be a sentence, as it is really a fragment.

spec/index.bs
```diff
@@ -330,8 +331,26 @@ the user agent must process it as follows:
...
+    :   the user clears cookies or site data for a specific origin
+    ::  The user agent must remote all entries that would be affected
+            by the deleted cookies.
```
martinthomson, line 339: This is the part that I think needs to be clearer.  It might even need some more thought.

When you clear state for an RP, you probably just forget those identities (and the associated IdPs) that were used.

However, when you clear state for an IdP, there will be a bunch of RPs that have received identity information from the IdP.  I see a two options, each with drawbacks.

1. You could clear ALL state for any RPs as well on the basis that retaining state at the RP might lead to information the RP has could propagate to the IdP, re-establishing same-site recognition when it shouldn't (see below for an attack on this).  The drawback here is that clearing state in one place has an wider effect than users might expect.
2. You could clear just the login state, so that the RP would need to ask again.  The attack here is that, if the user chooses to use the IdP again, the RP could remind the IdP of the user identity.  The drawback is that now clearing state for the IdP ends up having unintended consequences.  And though the reminding thing is a real problem, that only happens if the user gives permission to the RP to use the IdP again AND the RP state is not cleared AND the user intends to have a clean break.  I don't believe that there is anything in FedCM that might leak the information prior to the user performing that grant (with the usual caveats about the effectiveness of cross-site information leakage protections).

Note that in the first case, if you allow for the possibility that each RP might also have established links to other IdPs, then you might also need to clear state for any of those RPs as well.  It's a connected graph and the only way to make clean break is to burn out all the links.  Of course, that might be even more surprising.  At some point though, you need to recognize that this sort of purging needs to consider all forms of linkability, so you might want to stop before the only action you permit involves clearing all browser state.

I don't think that there is an easy cut to make between these two.  I can see how different browsers might choose to take different approaches on the basis that they take different perspectives on how to empower their users.  One option favours user control over privacy, the other favours convenience.  For a specification, we generally try to document the considerations carefully and clearly, then leave these decisions to implementations.  It is OK to do that because there are no guarantees on retention of site-level state, just as there are no firm rules about how private information is managed, so each browser gets to choose their own approach.

Note that I'm talking about a problem with broader scope than your change.  You're the lucky one to have to contend with this.

martinthomson: Hmm, the more I think on this, the more I realize that this cuts in both directions.  The only thing that might make the RP clearing case simpler is that we have a different perception of the IdP role in this process.

* If the user deliberately chooses to reuse an identity, then the connection is remade, which is essentially an explicit request to re-establish state.

* If the user chooses a different identity, they may or may not have an expectation that the IdP not share the fact that this RP previously had access to the previous identity.

For the latter, we already rely fairly heavily on the user making a trust decision with respect to the information the IdP might have about them.  That is not a symmetric relationship with the RP.  We don't assume that the user trusts the RP in the same way.

(Lots of stuff to write...)

spec/index.bs
cbiesinger: I have added a section on interactions with Clear-Site-Data.

spec/index.bs
cbiesinger: I have elaborated on the line you commented on to describe the intent I had.

As you say, I think a lot of what you brought up here is only tangentially related to the IDP signin status API (this PR) and applies to FedCM in general.

However, the way I think about it is that the connection between RP and IDP is more akin to a permission. Permissions in general don't get deleted when the user clears cookies... To be honest, I think it would be surprising to users if deleting (e.g.) google.com cookies logged them out of (e.g.) Pinterest and Tripit?

I think I'm going to file a github issue to discuss all this, because it should probably involve not just you and me :)

cbiesinger: I filed https://github.com/fedidcg/FedCM/issues/493

spec/index.bs
```diff
@@ -314,6 +314,67 @@ value (which is used when a resource loaded as a third-party, not first-party).
 for an [=IDP=] to adopt the FedCM API. It doesn't introduce security issues on the API because the
 [=RP=] cannot inspect the results from the fetches in any way.
 
+<!-- ============================================================ -->
+## IDP Sign-in Status ## {#browser-api-idp-sign-in-status}
```
samuelgoto, line 318:
```suggestion
## The Sign-in Status API ## {#browser-api-sign-in-status}
```

spec/index.bs
```diff
@@ -559,6 +620,14 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
 or a pair (failure, bool), where the bool indicates whether to skip delaying
 the exception thrown.
     1. Assert: These steps are running [=in parallel=].
+    1. Let |signinStatus| be the value of the entry in [=IDP Sign-in Status map=]
```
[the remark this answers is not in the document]
samuelgoto: Also need to get by "type=idp" here.

spec/index.bs
```diff
@@ -1139,13 +1234,27 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
   };
 
   [Exposed=Window, SecureContext] interface IdentityProvider {
+      static undefined recordSignIn();
+      static undefined recordSignOut();
```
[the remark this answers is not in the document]
samuelgoto: Or maybe we should just have one option or another?

spec/index.bs
```diff
@@ -1139,13 +1234,27 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
...
       static Promise&lt;sequence&lt;IdentityUserInfo&gt;&gt; getUserInfo(IdentityProviderConfig config);
   };
 </pre>
 
 Issue: [Decide](https://github.com/fedidcg/FedCM/issues/476) whether {{IdentityProvider}} is the
 correct location for the {{IdentityProvider/getUserInfo()}} method.
 
+Issue: Should the recordSign{In,Out} functions return `Promise<undefined>`, so the caller can know
```
samuelgoto, line 1246:
```suggestion
```

Yes, I think it should return Promise<void>.

spec/index.bs
```diff
@@ -1139,13 +1234,27 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
...
+    when the call is finished?
+
+Issue: [Reconcile](https://github.com/speced/bikeshed/issues/2636) how these
```
samuelgoto, line 1249:
```suggestion
Issue: [Reconcile](https://github.com/speced/bikeshed/issues/2636) how these
```

Done, I think if you address these comments it will have reconciled with the Login Status API.

spec/index.bs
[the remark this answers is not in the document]
cbiesinger: I don't think we need two dictionaries at this point. If we decide we want a second one we can easily do so later.

cbiesinger (commented):
I think this should be ready for review now

spec/index.bs
```diff
@@ -362,17 +366,38 @@ User agents must also clear the [=IDP Sign-in Status map=] data when:
...
+
+  partial interface Navigator {
+    static Promise&lt;undefined&gt; recordSignIn(optional SignInStatusOptions options = {});
+    static Promise&lt;undefined&gt; recordSignOut(optional SignInStatusOptions options = {});
+  };
+</pre>
+
+When {{Navigator/recordSignIn}} or {{Navigator/recordSignOut}} is
+called:
+  * If no |options| dictionary is provided, or there is no |idp| member inthe
```
TallTed, line 388:
```suggestion
  * If no |options| dictionary is provided, or there is no |idp| member in the
```

spec/index.bs
```diff
@@ -362,17 +366,38 @@ User agents must also clear the [=IDP Sign-in Status map=] data when:
...
+    |options|, or the |idp| member is false, then the promise is rejected with
+    a {{TypeError}}.
+  * Otherwise, the user agent [=map/set|sets=] the entry in the [=Sign-in Status
+    Map=] with the [=map/key=] being  [=/this=]'s [=/relevant settings object=]'s
```
TallTed, line 392:
```suggestion
    Map=] with the [=map/key=] being [=/this=]'s [=/relevant settings object=]'s
```

TallTed (changes requested):
Tiny, human-facing

spec/index.bs
```diff
@@ -314,6 +314,92 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+The user agent checks all page and subresource loads for an HTTP response
+[=header=] named <dfn><code>SignIn-Status</code></dfn>, whose value is
+parsed as HTTP [[RFC9110#section-5.6.6|Parameter]]s.
+If that header exists, there is a parameter with name `type` and value `idp`,
+and any parameter has a name of `action`, the user agent must process it as
+follows:
+
+<div algorithm="process the signin status header">
+* If this is a subresource request, and the response does not have the
+    same [=/origin] as the toplevel page, ignore the header.
```
npm1, line 336:
```suggestion
    same [=/origin=] as the toplevel page, ignore the header.
```

spec/index.bs
```diff
@@ -314,6 +314,92 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+    ::  The user agent must remove all entries that would be affected
+            by the deleted cookies, that is, any entry with an origin
+            to which a deleted cookie could be sent to.
+
+            Note: For example, domain cookies may affect subdomains of
+                the deleted origin.
+    :   if the user agent allows deleting individual cookies
+    ::  the behavior is user agent-defined.
+
+            Note: The user agent may want to reset the state to undefined,
```
npm1, line 363:
```suggestion
            Note: The user agent may want to reset the state to [=unknown=],
```

spec/index.bs
```diff
@@ -559,6 +645,14 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
 or a pair (failure, bool), where the bool indicates whether to skip delaying
 the exception thrown.
     1. Assert: These steps are running [=in parallel=].
+    1. Let |signinStatus| be the value of the entry in [=Sign-in Status map=]
+        with the key being the origin of |provider|'s configURL. If there is
+        no such entry, set it to a user-agent specific default (either [=unknown=] or
+        [=signed-out=]).
+    1. If |signinStatus| is [=signed-out=], return failure.
+        Issue: Should the user agent instead show UI to let the user sign
```
npm1, line 653: This is not rendering as an issue (maybe needs a blank line before.

spec/index.bs
```diff
@@ -569,6 +663,32 @@ the exception thrown.
     1. If |config| is failure, return (failure, false).
     1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
         and |globalObject|.
+    1. If the fetch failed, or the size of |accountsList| is 0:
```
npm1, line 666: I think fetch failing is being passed as |accountsList| being failure

spec/index.bs
```diff
@@ -569,6 +663,32 @@ the exception thrown.
...
+        1. [=map/set|Set=] an entry in the [=Sign-in Status map=] with the
+            key being the origin of the configURL and the value being [=signed-out=].
```
npm1, line 668: Link origin, link configURL

spec/index.bs
```diff
@@ -569,6 +663,32 @@ the exception thrown.
...
+            A user agent may decide to skip this step if no credentials were sent to
+            server.
+
+            Note: For example, if the fetch failed due to a DNS error, no
+                credentials were sent. In this situation, we do not know whether
+                the user is signed in or not and so we may not want to reset
+                the status.
+        1. If |signinStatus| is [=signed-in=], show a dialog to the user and return
```
npm1, line 676: We (Chrome) don't want to return failure right away here? In this case, we wait until the sign in dialog to resolve, and if we see signin then we refetch accounts?

cbiesinger:
I've addressed all comments so far. In particular, I have now specified an IDP sign-in flow based on the signin_url. Please take a look, especially @bvandersloot-mozilla and @npm1

spec/index.bs
```diff
@@ -659,21 +659,26 @@ the exception thrown.
         with the key being the origin of |provider|'s configURL. If there is
         no such entry, set it to a user-agent specific default (either [=unknown=] or
         [=signed-out=]).
-    1. If |signinStatus| is [=signed-out=], return failure.
+    1. If |signinStatus| is [=signed-out=], the user agent must either:
```
TallTed, line 662:
```suggestion
    1. If |signinStatus| is [=signed-out=], the user agent must do one of the following:
```

spec/index.bs
```diff
@@ -659,21 +659,26 @@ the exception thrown.
...
 
-        Issue: Should the user agent instead show UI to let the user sign
-            in to the IDP? This could be especially useful in the multi-IDP
-            case. See https://github.com/fedidcg/FedCM/issues/442 for discussion.
+        1. Return failure.
+        1. Prompt the user whether to continue. This MAY include an affordance
```
TallTed, line 665:
```suggestion
        * Return failure.
        * Prompt the user whether to continue. This MAY include an affordance
```

spec/index.bs
```diff
@@ -1255,6 +1273,28 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+        the [=/origin=] of the {{IdentityProviderAPIConfig/signin_url}} and
+        the value is [=signed-in=]. If such a change happens, set
+        |status_changed| to true.
+
+        Note: The IDP sign-in flow may trigger this change using either
+            the [[#signin-status-javascript|JavaScript]] or
+            [[#signin-status-http|HTTP header]] API.
+    1. Wait for one of the following conditions:
+        1. The user closes the browsing context: return failure.
+        1. {{IdentityProvider}}.{{IdentityProvider/close}} is called:
```
TallTed, line 1292:
```suggestion
        * The user closes the browsing context: return failure.
        * {{IdentityProvider}}.{{IdentityProvider/close}} is called:
```

TallTed (changes requested):
Minor but helpful for comprehension: use bullets (`<ul>`) for either/or; use numbers (`<ol>`) for steps/sequences.

npm1 (commented):
Some nits up to "show sign-in dialog"

spec/index.bs
```diff
@@ -1139,13 +1311,23 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
...
 correct location for the {{IdentityProvider/getUserInfo()}} method.
 
+A {{IdentityProvider/close}} function is provided to signal to the browser that
+the signin flow is finished. The reason for this function in addition to the
+header is that even when the user is already logged in, the signin flow may not
+be finished yet; for example, an IDP may want to prompt the user to verify
+their phone number. To allow for such flows, the IDP must call
+{{IdentityProvider/close}} when the flow is fully done.
+
+See the [=show an IDP sign-in dialog=] algorithm for more details.
```
npm1, line 1329: Hmm ok so this does nothing if it is invoked and we were not waiting for it to be called, i.e. step "show an IDP sign-in dialog"?

bvandersloot-mozilla:
Looking at this and a lot has changed since February. Thoughts to start out:
- you probably want a `navigatable` inside of `show an IDP sign-in dialog`.
- the steps for "If signinStatus is[ signed-out ](https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/436/273ccb0...cbiesinger:654908f.html#signed-out), the user agent must either:" are very broad. I _think_ that this will end up working out because the second option includes a user-cancel to failure. 
- Integration with Login Status API is still pending as a big issue. 

Overall, I believe this represents what we agreed would be reasonable in February (although, admittedly that was a long time ago). I would hope we can resolve the log in status integration in person this week at TPAC.

cbiesinger:
While I work on npm's and Ted's issues, and the TPAC changes:

> * you probably want a `navigatable` inside of `show an IDP sign-in dialog`.

This is done now

> * the steps for "If signinStatus is[ signed-out ](https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/436/273ccb0...cbiesinger:654908f.html#signed-out), the user agent must either:" are very broad. I _think_ that this will end up working out because the second option includes a user-cancel to failure.

Do you want it to say something else? I'm not 100% sure on what your UI will look like here.

> * Integration with Login Status API is still pending as a big issue.

Gladly that was resolved during TPAC. I will comment again once I have updated this PR accordingly. (See also https://github.com/fedidcg/login-status)

spec/index.bs
```diff
@@ -649,43 +659,62 @@ the exception thrown.
...
+    1. If |signinStatus| is [=signed-out=], the user agent must either:
+
+        1. Return failure.
+        1. Prompt the user whether to continue. This MAY include an affordance
+            to [=show an IDP sign-in dialog=]. If the user cancels or that
+            algorithm returns failure, return failure.
+
+        Issue: We should perhaps provide a way to let the RP request that
+            the second option is provided, possibly gated on a user gesture.
+            See https://github.com/fedidcg/FedCM/issues/442 for discussion.
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -649,43 +659,62 @@ the exception thrown.
...
     1. Let |requiresUserMediation| be |provider|'s {{IdentityProviderConfig/configURL}}'s [=/origin=]'s
         [=requires user mediation=].
     1. Let |mediation| be |options|'s {{CredentialRequestOptions/mediation}}.
     1. If |requiresUserMediation| is true and |mediation| is
         "{{CredentialMediationRequirement/silent}}", return failure.
     1. Let |config| be the result of running [=fetch the config file=] with |provider| and
-        |globalObject|.
+        |globalObject| (this is the <dfn>fetch config step</dfn>).
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -649,43 +659,62 @@ the exception thrown.
...
     1. If |config| is failure, return (failure, false).
     1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
-        and |globalObject|.
-    1. If the fetch failed, or the size of |accountsList| is 0:
+        and |globalObject| (this is the <dfn>fetch accounts list step</dfn>).
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -649,43 +659,62 @@ the exception thrown.
...
-            key being the origin of the configURL and the value being [=signed-out=].
-            A user agent may decide to skip this step if no credentials were sent to
-            server.
+            key being the [=/origin=] of the {{IdentityProviderConfig/configURL}}
+            and the value being [=signed-out=].  A user agent may decide to
+            skip this step if no credentials were sent to server.
 
             Note: For example, if the fetch failed due to a DNS error, no
                 credentials were sent. In this situation, we do not know whether
                 the user is signed in or not and so we may not want to reset
```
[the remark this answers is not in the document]
cbiesinger: Is that not redundant with "no credentials were sent" in the previous line?

spec/index.bs
```diff
@@ -649,43 +659,62 @@ the exception thrown.
...
                 the status.
-        1. If |signinStatus| is [=signed-in=], show a dialog to the user and return
-            failure.
+        1. If |signinStatus| is [=signed-in=], show a dialog to the user. The
+            contents of this dialog are defined by the user agent. This dialog
+            SHOULD provide an affordance for the user to trigger the [=show an
+            IDP sign-in dialog=] algorithm; this dialog is then the <dfn>confirm
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -649,43 +659,62 @@ the exception thrown.
...
                 impossible by always showing UI of some kind when credentials
                 were sent to the server.
+
+            If the user closes the dialog, return failure.
+
+            If the [=show an IDP sign-in dialog=] algorithm was triggered:
+
+                1. Let |result| be the result of that algorithm.
+                1. If |result| is failure, return failure. The user agent MAY
+                    show a dialog to the user before or after returning failure
```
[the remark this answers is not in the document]
cbiesinger: I wanted the flexibility to show an error dialog in this case, e.g. if we make this work: https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/ui/webid/identity_dialog_controller.cc;l=87;drc=b74163d3d63e781ce9899ec6a4e5a7c113d960b9;bpv=1;bpt=1

I am not aware of specific plans from either us or Mozilla, though I also did not specifically ask them.

spec/index.bs
cbiesinger: Yeah, do you have a different suggestion? We could make it return a promise and reject it in other cases, not sure that's necessary though.

spec/index.bs
```diff
@@ -649,43 +659,62 @@ the exception thrown.
...
-                IDP signin dialog</dfn> and provides a "sign in" button allowing
-                the user to sign in to the IDP using
-                {{IdentityProviderAPIConfig/signin_url}}.
+                is signed out.
 
             Note: This dialog ensures that silent tracking of the user is
                 impossible by always showing UI of some kind when credentials
                 were sent to the server.
+
+            If the user closes the dialog, return failure.
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -649,43 +659,62 @@ the exception thrown.
...
+
+            If the [=show an IDP sign-in dialog=] algorithm was triggered:
+
+                1. Let |result| be the result of that algorithm.
+                1. If |result| is failure, return failure. The user agent MAY
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -649,43 +659,62 @@ the exception thrown.
...
+                    show a dialog to the user before or after returning failure
+                    indicating this failure.
+                1. Otherwise, go back to the [=fetch config step=]. As an
```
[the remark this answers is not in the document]
cbiesinger: Good question. What do you think we should do?

It's not that bad if we keep going, since the user can always cancel.

spec/index.bs
```diff
@@ -1244,6 +1273,28 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
     1. Return |json|.
 </div>
 
+<div algorithm>
+To <dfn>show an IDP sign-in dialog</dfn> given an {{IdentityProviderAPIConfig}} |config|, run
+the following steps. This returns success or failure.
+    1. Open a new [=/browsing context=].
+    1. Load |config|.{{IdentityProviderAPIConfig/signin_url}} in this new browsing context.
+    1. Let |status_changed| be false.
+    1. Listen for changes to the [=sign-in status map=] where the key is the
```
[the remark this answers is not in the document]
cbiesinger: I added a note.

spec/index.bs
```diff
@@ -1244,6 +1273,28 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+        the [=/origin=] of the {{IdentityProviderAPIConfig/signin_url}} and
+        the value is [=signed-in=]. If such a change happens, set
+        |status_changed| to true.
```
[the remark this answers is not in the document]
cbiesinger: I've simplified this section and that case is now handled.

cbiesinger (commented):
Addressed the Ted and npm comments, still need to update the header/JS API names

cbiesinger:
OK I think this should be fully ready to be reviewed again!

spec/index.bs
```diff
@@ -334,24 +334,23 @@ map</dfn>, an initially empty [=map=]. The [=map/keys=] in this map are
...
-and any parameter has a name of `action`, the user agent must process it as
-follows:
+For each [=http-redirect fetch=] and [=http fetch=]'s [=response=], let |value|
+be the result of [=get a structured field value=] from the response's header
+list with name "<dfn><code>Set-Login</code></dfn>" and type "`item`". If |value| is not null,
+process this header as follows:
 
 <div algorithm="process the signin status header">
 * If this is a subresource request, and the response does not have the
     same [=/origin=] as the toplevel page, ignore the header.
```
npm1, line 344: I think we need to be more precise about what is the origin. Is this request's currentURL's origin, or something else?

npm1, line 344: "toplevel page" is a bit vague, maybe link https://html.spec.whatwg.org/multipage/webappapis.html#concept-environment-top-level-origin

spec/index.bs
```diff
@@ -334,24 +334,23 @@ map</dfn>, an initially empty [=map=]. The [=map/keys=] in this map are
...
-If that header exists, there is a parameter with name `type` and value `idp`,
-and any parameter has a name of `action`, the user agent must process it as
-follows:
+For each [=http-redirect fetch=] and [=http fetch=]'s [=response=], let |value|
+be the result of [=get a structured field value=] from the response's header
+list with name "<dfn><code>Set-Login</code></dfn>" and type "`item`". If |value| is not null,
+process this header as follows:
 
 <div algorithm="process the signin status header">
 * If this is a subresource request, and the response does not have the
```
npm1, line 343: "this is a subresource request" is a bit unclear for inside an algorithm. I wonder if this should use https://fetch.spec.whatwg.org/#concept-request-mode, or something else?

spec/index.bs
```diff
@@ -334,24 +334,23 @@ map</dfn>, an initially empty [=map=]. The [=map/keys=] in this map are
...
     same [=/origin=] as the toplevel page, ignore the header.
 * Otherwise:
-    * If the parameter value is `signin`, [=map/set=] an entry in the
+    * Assert that |value| is a tuple.
```
npm1, line 346: The algorithm could return null for example, so this assert seems incorrect.

spec/index.bs
```diff
@@ -334,24 +334,23 @@ map</dfn>, an initially empty [=map=]. The [=map/keys=] in this map are
...
+    * Let |token| be the first entry of |value|.
+    * If |token| is `login`, [=map/set=] an entry in the
         [=Sign-in Status map=] with the key being the origin of the resource
-        and the value being [=signed-in=]
-    * If the parameter value is `signout-all`, [=map/set=] an entry in the
+        and the value being [=signed-in=].
```
npm1, line 350: Perhaps we change this to `logged-in` / `logged-out`?

spec/index.bs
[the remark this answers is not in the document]
npm1: Yea, I suppose that is fine

spec/index.bs
```diff
@@ -1276,22 +1277,24 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+    1. [=Create a fresh top-level traversable=] with URL |config|.
+        {{IdentityProviderAPIConfig/signin_url}}.
+    1. The user agent MAY [=set up browsing context features=] or otherwise
+        affect the presentation of this traversable in an implementation-defined
+        way.
     1. Wait for one of the following conditions:
-        1. The user closes the browsing context: return failure.
-        1. {{IdentityProvider}}.{{IdentityProvider/close}} is called:
+        * The user closes the browsing context: return failure.
+        * {{IdentityProvider}}.{{IdentityProvider/close}} is called:
```
npm1, line 1287: Called within the context of this new traversable I guess?

npm1 (commented):
Thanks for working on this!

spec/index.bs
cbiesinger: I ended up dropping the subresource part because it's not necessary IMO

spec/index.bs

spec/index.bs
cbiesinger: We only run this algorithm if value is not null, see above.

spec/index.bs

spec/index.bs

spec/index.bs
```diff
@@ -559,16 +656,69 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
+
+                1. Let |result| be the result of that algorithm.
+                1. If |result| is failure, return (failure, true). The user
+                    agent MAY show a dialog to the user before or after
+                    returning failure indicating this failure.
+                1. Otherwise, go back to the [=fetch config step=]. As an
+                    optimization, the user agent MAY instead go to the
+                    [=fetch accounts list step=].
+
+    1. If |loginStatus| is [=unknown=], [=map/set=] an entry in the [=login
```
npm1, line 719: Why not just set the value to logged-in regardless, for simplicity? A user agent can optmize this if needed in their code.

spec/index.bs
```diff
@@ -1123,6 +1274,31 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
     1. Return |json|.
 </div>
 
+<div algorithm>
+To <dfn>show an IDP sign-in dialog</dfn> given an {{IdentityProviderAPIConfig}} |config|, run
+the following steps. This returns success or failure.
+    1. [=Create a fresh top-level traversable=] with URL |config|.
+        {{IdentityProviderAPIConfig/signin_url}}.
```
npm1, line 1281: space between the . and signin_url is weird. perhaps change to "|config|'s"

spec/index.bs
npm1: I think this is a bit unconventional but probably fine.

spec/index.bs
```diff
@@ -1139,13 +1315,23 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
...
   };
 </pre>
 
 Issue: [Decide](https://github.com/fedidcg/FedCM/issues/476) whether {{IdentityProvider}} is the
 correct location for the {{IdentityProvider/getUserInfo()}} method.
 
+A {{IdentityProvider/close}} function is provided to signal to the browser that
+the signin flow is finished. The reason for this function in addition to the
+header is that even when the user is already logged in, the signin flow may not
+be finished yet; for example, an IDP may want to prompt the user to verify
```
npm1, line 1329: nit: link IDPs here and below

npm1 (commented):
Did a full pass again. Added a bunch of nits but the core algorithms generally look good to me!

spec/index.bs
```diff
@@ -314,6 +315,102 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+        and the value being [=logged-in=].
+    * If |token| is `logout`, [=map/set=] an entry in the
+        [=Login Status map=] with the key being the origin of the resource
+        and the value being [=logged-out=].
+
+</div>
+
+### JavaScript API ### {#signin-status-javascript}
+
+IDPs can also use a JavaScript API to update the stored sign-in status:
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -314,6 +315,102 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+interface NavigatorLogin {
+  Promise&lt;undefined&gt; setStatus(LoginStatus status);
+};
+
+partial interface Navigator {
+  [SecureContext] readonly attribute NavigatorLogin login;
+};
+</pre>
+
+When {{NavigatorLogin/setStatus()}} is called with argument |status|:
```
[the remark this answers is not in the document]
cbiesinger: Done

spec/index.bs
cbiesinger: Fixed by wrapping the line differently

spec/index.bs

spec/index.bs
```diff
@@ -559,16 +656,69 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
+        no such entry, set it to a user-agent specific default (either [=unknown=] or
+        [=logged-out=]).
+    1. If |loginStatus| is [=logged-out=], the user agent must do one of the following:
+
+        * Return failure.
+        * Prompt the user whether to continue. This MAY include an affordance
+            to [=show an IDP sign-in dialog=]. If the user cancels or that
+            algorithm returns failure, return (failure, true).
+
+        Issue: We should perhaps provide a way to let the RP request that
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -559,16 +656,69 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
 or a pair (failure, bool), where the bool indicates whether to skip delaying
 the exception thrown.
     1. Assert: These steps are running [=in parallel=].
+    1. Let |loginStatus| be the value of the entry in [=Login Status map=]
+        with the key being the origin of |provider|'s configURL. If there is
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -559,16 +656,69 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
+        no such entry, set it to a user-agent specific default (either [=unknown=] or
+        [=logged-out=]).
+    1. If |loginStatus| is [=logged-out=], the user agent must do one of the following:
+
+        * Return failure.
```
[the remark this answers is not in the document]
cbiesinger: Pretty sure this should be false; done with that.

spec/index.bs
```diff
@@ -559,16 +656,69 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
+        * Prompt the user whether to continue. This MAY include an affordance
```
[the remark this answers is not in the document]

spec/index.bs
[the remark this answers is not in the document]
cbiesinger: Done

spec/index.bs
```diff
@@ -559,16 +656,69 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
+
+            Note: This situation happens when the browser expects the user
+                to be signed in, but the accounts fetch indicated that the user
+                is signed out.
+
+            Note: This dialog ensures that silent tracking of the user is
+                impossible by always showing UI of some kind when credentials
+                were sent to the server.
+
+            If the user closes the dialog, return (failure, true).
```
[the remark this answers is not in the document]
cbiesinger: Done, but using * instead of 1. because the two steps are mutually exclusive.

spec/index.bs
```diff
@@ -559,16 +656,69 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
+
+            If the [=show an IDP sign-in dialog=] algorithm was triggered:
+
+                1. Let |result| be the result of that algorithm.
+                1. If |result| is failure, return (failure, true). The user
+                    agent MAY show a dialog to the user before or after
+                    returning failure indicating this failure.
+                1. Otherwise, go back to the [=fetch config step=]. As an
```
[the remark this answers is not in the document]
cbiesinger: Because the implementation is easier that way? It's what Chrome does...

spec/index.bs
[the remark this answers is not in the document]

cbiesinger: Done. Also linked origin and configURL in the line below.

spec/index.bs
```diff
@@ -314,6 +315,102 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+
+partial interface Navigator {
+  [SecureContext] readonly attribute NavigatorLogin login;
+};
+</pre>
+
+When {{NavigatorLogin/setStatus()}} is called with argument |status|:
+* Let |value| be [=logged-in=] if |status| is `"logged-in"` or [=logged-out=]
+    if |status| is `"logged-out"`.
+* [=map/set|Set=] the entry in the [=Login Status Map=] with the [=map/key=]
```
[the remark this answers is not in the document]
cbiesinger: Great catch, done.

cbiesinger (commented):
Please take another look!

spec/index.bs
```diff
@@ -375,13 +375,18 @@ partial interface Navigator {
 };
 </pre>
 
+<div algorithm="setStatus">
 When {{NavigatorLogin/setStatus()}} is called with argument |status|:
+* If the [=/origin=] of the [=current settings object=] is not [=same origin=]
```
npm1, line 380: nit: prefer "1."  to "*" as these read sequential?

spec/index.bs
```diff
@@ -687,7 +694,8 @@ the exception thrown.
             skip this step if no credentials were sent to server.
 
             Note: For example, if the fetch failed due to a DNS error, no
-                credentials were sent. In this situation, we do not know whether
+                credentials were sent and therefore the IDP did not learn
```
npm1, line 697: [=IDP=]

spec/index.bs
```diff
@@ -704,9 +712,9 @@ the exception thrown.
                 impossible by always showing UI of some kind when credentials
                 were sent to the server.
 
-            If the user closes the dialog, return (failure, true).
+            * If the user closes the dialog, return (failure, true).
```
npm1, line 715: I think we need to add to the previous step, before the notes (or a new step): "Wait until one of the following occurs".

spec/index.bs
```diff
@@ -704,9 +712,9 @@ the exception thrown.
...
 
-            If the [=show an IDP sign-in dialog=] algorithm was triggered:
+            * If the [=show an IDP sign-in dialog=] algorithm was triggered:
```
npm1, line 717: Then in these we can remove the "If"

spec/index.bs
```diff
@@ -716,14 +724,14 @@ the exception thrown.
...
                     [=fetch accounts list step=].
 
-    1. If |loginStatus| is [=unknown=], [=map/set=] an entry in the [=login
-        status map=] with the key being the origin of the configURL and the value being
+    1. Assert: |accountsList| is not failure and the size of |accountsList| is not 0.
+    1. [=map/Set=] an entry in the [=login status map=] with the key being the
+        [=/origin] of the {{IdentityProviderConfig/configURL}} and the value being
         [=logged-in=].
     1. If |provider|'s {{IdentityProviderConfig/loginHint}} is not empty:
         1. For every |account| in |accountList|, remove |account| from |accountList| if |account|'s
```
npm1, line 732: Note that we may remove all accounts if there is none matching the login hint. I think here we need to check if accountsList becomes 0 => then send to the step that creates the confirm idp signin dialog?

npm1 (commented):
We should reconcile signin/login somehow.... Not ideal having both sprinkled all over. Any ideas for that?

spec/index.bs

spec/index.bs

spec/index.bs
cbiesinger: Done, although I think it makes more sense after the notes, because the notes are referencing the paragraph directly above them

spec/index.bs

spec/index.bs
cbiesinger: Done

cbiesinger (commented):
I changed most signin to login, except for the RP context

spec/index.bs
```diff
@@ -412,8 +412,7 @@ User agents must also clear the [=Login Status map=] data when:
         the [=map/key=] is the input origin.
 
 Note: Other website-initiated cookie changes should not affect this map. When
-    IDP sign-in state changes, it should send an explicit
-    [=Set-Login=] header.
+    IDP login state changes, it should send an explicit [=Set-Login=] header.
```
npm1, line 415: nit: [=IDP=], [=RP=] in this note

spec/index.bs
```diff
@@ -712,17 +711,19 @@ the exception thrown.
                 impossible by always showing UI of some kind when credentials
                 were sent to the server.
 
-            * If the user closes the dialog, return (failure, true).
+            1. Wait until one of the following occurs:
```
npm1, line 714: Does this need to be a substep of the mismatch dialog step? Or can it just be step 3 (the step after)

npm1 (approved):
Some other places where signin/signup is used (no strong opinion on which is correct but figured I would flag:

- Lots of text (intro and first sections) use signin instead of login (though this is probably fine)
- "Request permission to sign-up" (not webexposed, so probably ok?)
- Section 5.5 loginState is set to "SignIn" or "SignUp" (this is returned in the method)

spec/index.bs

spec/index.bs
cbiesinger: Yes, it needs to be a substep, because all of this only happens if loginStatus is logged-in.

cbiesinger (commented):
I did not want to update section 5.5 (webdriver accountlist command) because:
1. That is unrelated to IDP signin status
2. We've shipped those for a while
3. Changing this also requires changing Selenium (https://github.com/SeleniumHQ/selenium/blob/trunk/java/src/org/openqa/selenium/federatedcredentialmanagement/FederatedCredentialManagementAccount.java#L54)

spec/index.bs
```diff
@@ -1123,6 +1284,31 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+            1. Let |loginStatus| be the value of the entry in [=Login Status map=]
+                with the key being the origin of the {{IdentityProviderAPIConfig/login_url}}.
+
+                Note: The IDP login flow may set this value to logged-in using
+                    either the [[#login-status-javascript|JavaScript]] or
+                    [[#login-status-http|HTTP header]] API during the login
+                    flow. It is also possible that this change happened in
+                    a different browsing context.
+            1. If |loginStatus| is [=logged-in=], return success.
+            1. Otherwise, return failure.
```
samuelgoto, line 1309: Just wondering: shouldn't we restart the process instead of returning a failure?

spec/index.bs
```diff
@@ -559,21 +660,80 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
+                to be signed in, but the accounts fetch indicated that the user
+                is signed out.
+
+            Note: This dialog ensures that silent tracking of the user is
+                impossible by always showing UI of some kind when credentials
+                were sent to the server.
+
+            1. Wait until one of the following occurs:
+
+                * If the user closes the dialog, return (failure, true).
```
bvandersloot-mozilla, line 716: Why does this return failure immediately?

This allows a user-visible read of the logged in bit.

spec/index.bs
```diff
@@ -559,21 +660,80 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
+
+                * If the [=show an IDP login dialog=] algorithm was triggered:
+
+                    1. Let |result| be the result of that algorithm.
+                    1. If |result| is failure, return (failure, true). The user
+                        agent MAY show a dialog to the user before or after
+                        returning failure indicating this failure.
+                    1. Otherwise, go back to the [=fetch config step=]. As an
```
bvandersloot-mozilla, line 724: Let's pick one here.

spec/index.bs
```diff
@@ -314,6 +315,106 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+};
+
+partial interface Navigator {
+  [SecureContext] readonly attribute NavigatorLogin login;
+};
+</pre>
+
+<div algorithm="setStatus">
+When {{NavigatorLogin/setStatus()}} is called with argument |status|:
+1. If the [=/origin=] of the [=current settings object=] is not [=same origin=]
```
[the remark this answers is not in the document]
cbiesinger: I think you're right that we should since in that situation the inner A only gets partitioned (or no) cookies and thus is not aware of the "real" toplevel login state. I have updated this (and also the HTTP header part)

spec/index.bs
```diff
@@ -314,6 +315,106 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+Issue: Reconcile this section with the
+    [Login Status API](https://github.com/privacycg/is-logged-in).
+    See also [this PR](https://github.com/privacycg/is-logged-in/pull/54).
+
+### HTTP header API ### {#login-status-http}
+
+Issue: The HTTP header checking should move into the Fetch spec, since it
+    affects all resource loads.
+
+Each [=user agent=] keeps a global, persistent <dfn>Login Status
```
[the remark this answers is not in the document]
cbiesinger: Yes, done.

spec/index.bs
```diff
@@ -314,6 +315,106 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+};
+</pre>
+
+<div algorithm="setStatus">
+When {{NavigatorLogin/setStatus()}} is called with argument |status|:
+1. If the [=/origin=] of the [=current settings object=] is not [=same origin=]
+    with the [=top-level origin=], throw a {{SecurityError}} {{DOMException}}.
+1. Let |value| be [=logged-in=] if |status| is `"logged-in"` or [=logged-out=]
+    if |status| is `"logged-out"`.
+1. [=map/set|Set=] the entry in the [=Login Status Map=] with the [=map/key=]
```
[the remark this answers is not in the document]
cbiesinger: Done (with a helper variable for origin)

spec/index.bs
```diff
@@ -314,6 +315,106 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+* If the request's [=request/current url=]'s [=/origin=] is not [=same origin=]
+    with the [=environment/top-level origin=], ignore the header.
+* Otherwise:
+    * Assert that |value| is a tuple.
+    * Let |token| be the first entry of |value|.
+    * If |token| is `login`, [=map/set=] an entry in the
+        [=Login Status map=] with the key being the origin of the resource
+        and the value being [=logged-in=].
+    * If |token| is `logout`, [=map/set=] an entry in the
+        [=Login Status map=] with the key being the origin of the resource
```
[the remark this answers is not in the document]

spec/index.bs

spec/index.bs
```diff
@@ -1123,6 +1284,31 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
     1. Return |json|.
 </div>
 
+<div algorithm>
+To <dfn>show an IDP login dialog</dfn> given an {{IdentityProviderAPIConfig}} |config|, run
+the following steps. This returns success or failure.
+    1. [=Create a fresh top-level traversable=] with URL
```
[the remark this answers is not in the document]
cbiesinger: I added a validation step to fetching the config file to ensure that signin_url is also valid here. I also think that this represents the desired semantics better (don't accept the config file if there is no login_url even if login_url is not triggered during this specific request)

spec/index.bs
```diff
@@ -314,6 +315,106 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+    if |status| is `"logged-out"`.
+1. [=map/set|Set=] the entry in the [=Login Status Map=] with the [=map/key=]
+    being [=/this=]'s [=/relevant settings object=]'s
+    [=environment settings object/origin=] and the value being |value|.
+
+</div>
+
+### Clearing map data ### {#login-status-clear-data}
+
+User agents must also clear the [=Login Status map=] data when:
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -559,21 +660,80 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
 or a pair (failure, bool), where the bool indicates whether to skip delaying
 the exception thrown.
     1. Assert: These steps are running [=in parallel=].
+    1. Let |loginStatus| be the value of the entry in [=Login Status map=]
+        with the key being the [=/origin=] of |provider|'s
+        {{IdentityProviderConfig/configURL}}. If there is no such entry, set it
```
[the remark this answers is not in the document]

cbiesinger (commented):
Comments addressed/responded to, please take another looo

spec/index.bs
bvandersloot-mozilla: Rejecting in a user-specified time is different than rejecting after a fixed 120s. So by seeing a rejection before that point, the IDP can determine they were not "logged-in".

spec/index.bs
```diff
@@ -1123,6 +1284,31 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
     1. Return |json|.
 </div>
 
+<div algorithm>
+To <dfn>show an IDP login dialog</dfn> given an {{IdentityProviderAPIConfig}} |config|, run
```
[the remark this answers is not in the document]
bvandersloot-mozilla: I meant for instances where the login dialog is in the same navigatable, and then navigate the user back to this page. I think I remember that being out of scope here.

spec/index.bs
cbiesinger: Ah, as of https://github.com/fedidcg/FedCM/commit/c8609ec14b5dde7a4f087e065378b8e457aadbd5 the 120s has been replaced with a random timer.

spec/index.bs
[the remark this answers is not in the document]
cbiesinger: I'm not positive I understand you correctly but I think that's out of scope yeah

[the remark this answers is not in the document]
asr-enid: I guess this refers to flow that require cross origin navigation, for technical reasons or cases with authentication intermediaries which are currently not properly covered by this

spec/index.bs
```diff
@@ -314,6 +315,110 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+### Login Status Map ### {#hdr-login-status-map}
+
+Each [=user agent=] keeps a global, persistent <dfn>Login Status
+map</dfn>, an initially empty [=map=]. The [=map/keys=] in this map are
+[=/origin=] (of [=IDPs=]), and the [=map/values=] are enums that can be one of
+"<dfn><code>unknown</code></dfn>", "<dfn><code>logged-in</code></dfn>",
+and "<dfn><code>logged-out</code></dfn>".
+
+<div algorithm>
+To <dfn>get a login status entry</dfn> for an [=/origin=] |origin|:
```
samuelgoto, line 331: Can we move the setter to this list of algorithms too? That way, we can easily just move this section out of this spec into an independent spec, such that Safari/Firefox could implement this spec without necessarily having to implement FedCM.

spec/index.bs
```diff
@@ -314,6 +315,110 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+1. If [=Login Status map=][|origin|] exists, return it.
+1. Otherwise, return a user-agent specific default (either [=unknown=] or
```
samuelgoto, line 333:
```suggestion
1. Otherwise, return [=unknown=].
```

Can we move the "the user-agent may choose to assume that "unknown" means "logged-out" to the place FedCM consumes this bit? This feels FedCM-specific, e.g. the "user-agent specific value for "unknown" can be different between FedCM and SAA?

spec/index.bs
```diff
@@ -314,6 +315,110 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+list with name "<dfn><code>Set-Login</code></dfn>" and type "`item`". If |value| is not null,
+process this header as follows:
+
+<div algorithm="process the login status header">
+* Let |origin| be the response's [=response/URL=]'s [=/origin=].
+* If |origin| is not [=same-origin with its ancestors=], ignore the header.
+* Otherwise:
+    * Assert that |value| is a tuple.
+    * Let |token| be the first entry of |value|.
+    * If |token| is `logged-in`, [=map/set=] [=Login Status map=][|origin|]
```
samuelgoto, line 354: Capture this in a "set" algorithm and use here instead?

spec/index.bs
samuelgoto, line 331:
```suggestion
To <dfn>get the login status</dfn> for an [=/origin=] |origin|:
```

spec/index.bs
```diff
@@ -559,21 +660,80 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
     1. Assert: These steps are running [=in parallel=].
+    1. Let |loginStatus| be the value of the entry in [=Login Status map=]
+        with the key being the [=/origin=] of |provider|'s
+        {{IdentityProviderConfig/configURL}}. If there is no such entry, set it
+        to a user-agent specific default (either [=unknown=] or
+        [=logged-out=]).
+    1. If |loginStatus| is [=logged-out=], the user agent must do one of the following:
+
+        * Return (failure, false).
+        * Prompt the user whether to continue. If the user continues, the user
```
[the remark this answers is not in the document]
samuelgoto: This is inside the "create an IdentityCredential" algorithm: nothing happens before this, right? That is, "UA prompts user to select an IDP before it does anything" isn't captured **before** we get to this point, right?

spec/index.bs
```diff
@@ -559,21 +664,75 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
+        "{{CredentialMediationRequirement/silent}}", return (failure, false).
+    1. Let |config| be the result of running [=fetch the config file=] with
+        |provider| and |globalObject|.
     1. If |config| is failure, return (failure, false).
-    1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
-        and |globalObject|.
+    1. <dfn>Fetch accounts list step</dfn>: Let |accountsList| be the result of
+        [=fetch the accounts list=] with |config|, |provider|, and |globalObject|.
+    1. If |accountsList| is failure, or the size of |accountsList| is 0:
+        1. [=map/set|Set=] an entry in the [=Login Status map=] with the
```
samuelgoto, line 691: Use a setter algorithm here instead.

spec/index.bs
samuelgoto: Just curious:  does our implementation do? Do we retry?

spec/index.bs

spec/index.bs

spec/index.bs
cbiesinger: I did not want to move the mapping of the HTTP token to the enum value to the algorithm, but I did add the set algorithm.

spec/index.bs

spec/index.bs
```diff
@@ -559,21 +671,75 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
 the exception thrown.
     1. Assert: These steps are running [=in parallel=].
+    1. Let |loginStatus| be the result of [=get the login status=] with
+        the [=/origin=] of |provider|'s {{IdentityProviderConfig/configURL}}.
+    1. If |loginStatus| is [=unknown=], a user agent MAY set it to [=logged-out=].
+    1. If |loginStatus| is [=logged-out=], the user agent MUST do one of the following:
+
+        * Return (failure, false).
+        * Prompt the user whether to continue. If the user continues, the user
+            agent SHOULD set |loginStatus| to [=unknown=]. This MAY include an
```
samuelgoto, line 681:
```suggestion
            agent SHOULD set |loginStatus| to [=unknown=]. This MUST include an
```

Isn't this a MUST? Otherwise, how can we move forward with the algorithm if the loginStatus is 'logged-out'?

samuelgoto: Don't we also have to "wait" somewhere here until the user becomes 'logged-in'?

samuelgoto (approved):
LGTM

spec/index.bs
cbiesinger: That happens if the user triggers the affordance to "show an IDP login dialog" and is encapsulated in that algorithm.

I don't think the SHOULD needs to be a MUST; the algorithm should work fine as-is, no? I left it that way because some user agents don't really want to use an unknown state.

spec/index.bs
```diff
@@ -314,6 +315,117 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
+
+            Note: For example, domain cookies may affect subdomains of
+                the deleted origin.
+    :   the user deletes individual cookies (if allowed by the user agent)
+    ::  the behavior is user agent-defined.
+
+            Note: The user agent MAY want to reset the state to [=unknown=],
+                since is impossible to know whether this cookie affects
+                authorization state.
+    : the user agent receives a <a http-header>Clear-Site-Data</a> header with a
```
npm1, line 419: What happens if this is sent from a subresource or iframe that does not match the top-level? I was trying to find an answer in Clear-Site-Data and it seems the spec is not updated at all to account for any sort of partitioning, but maybe we can at least add an issue to make sure we handle that at some point...

spec/index.bs
```diff
@@ -559,21 +671,75 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
+            the second option is provided, possibly gated on a user gesture.
+            See [this issue](https://github.com/fedidcg/FedCM/issues/442) for discussion.
     1. Let |requiresUserMediation| be |provider|'s {{IdentityProviderConfig/configURL}}'s [=/origin=]'s
         [=requires user mediation=].
     1. Let |mediation| be |options|'s {{CredentialRequestOptions/mediation}}.
     1. If |requiresUserMediation| is true and |mediation| is
-        "{{CredentialMediationRequirement/silent}}", return failure.
-    1. Let |config| be the result of running [=fetch the config file=] with |provider| and
-        |globalObject|.
+        "{{CredentialMediationRequirement/silent}}", return (failure, false).
```
npm1, line 692: Why should this be delayed? It seems to me like an issue with the developer call, so no harm rejecting immediately?

spec/index.bs
```diff
@@ -559,21 +671,75 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
+                * If the [=show an IDP login dialog=] algorithm was triggered:
+
+                    1. Let |result| be the result of that algorithm.
+                    1. If |result| is failure, return (failure, true). The user
+                        agent MAY show a dialog to the user before or after
+                        returning failure indicating this failure.
+                    1. Otherwise, go back to the [=fetch accounts list step=].
+
+    1. Assert: |accountsList| is not failure and the size of |accountsList| is not 0.
+    1. [=Set the login status=] for the [=/origin] of the
```
npm1, line 736: nit:  [=/origin=]

spec/index.bs
```diff
@@ -595,7 +761,7 @@ the exception thrown.
             [=compute the connection status=] algorithm given |provider| and |account|. When doing this,
             the user agent MAY show some UI to the user indicating that they are being
             <dfn>auto-reauthenticated</dfn>.
-    1. Otherwise, if |mediation| is "{{CredentialMediationRequirement/silent}}", return failure.
+    1. Otherwise, if |mediation| is "{{CredentialMediationRequirement/silent}}", return (failure, false).
```
npm1, line 764: In this case, we do not delay the rejection so (failure, true)

spec/index.bs
```diff
@@ -765,6 +934,7 @@ dictionary IdentityProviderAPIConfig {
   required USVString accounts_endpoint;
   required USVString client_metadata_endpoint;
   required USVString id_assertion_endpoint;
+  USVString login_url;
```
npm1, line 937: Why is this not required like the other ones? Not that it matters since we don't really use this in JS but still

spec/index.bs
```diff
@@ -559,21 +671,75 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
     1. Assert: These steps are running [=in parallel=].
+    1. Let |loginStatus| be the result of [=get the login status=] with
+        the [=/origin=] of |provider|'s {{IdentityProviderConfig/configURL}}.
+    1. If |loginStatus| is [=unknown=], a user agent MAY set it to [=logged-out=].
+    1. If |loginStatus| is [=logged-out=], the user agent MUST do one of the following:
+
+        * Return (failure, false).
+        * Prompt the user whether to continue. If the user continues, the user
+            agent SHOULD set |loginStatus| to [=unknown=]. This MAY include an
+            affordance to [=show an IDP login dialog=]. If the user cancels
```
npm1, line 682: nit: this requires parameter

spec/index.bs
```diff
@@ -559,21 +671,75 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
...
+
+            Note: For example, if the fetch failed due to a DNS error, no
+                credentials were sent and therefore the [=IDP=] did not learn
+                the user's identity. In this situation, we do not know whether
+                the user is signed in or not and so we may not want to reset
+                the status.
+        1. <dfn>Mismatch dialog step</dfn>: If |loginStatus| is [=logged-in=], show a
+            dialog to the user. The contents of this dialog are defined by the user
+            agent. This dialog SHOULD provide an affordance for the user to trigger
+            the [=show an IDP login dialog=] algorithm; this dialog is the
```
npm1, line 712: nit: parameter for show

npm1 (commented):
Took another pass. Looks great! Some nits

spec/index.bs
```diff
@@ -1123,6 +1293,31 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+        |config|.{{IdentityProviderAPIConfig/login_url}}.
+    1. The user agent MAY [=set up browsing context features=] or otherwise
+        affect the presentation of this traversable in an implementation-defined
+        way.
+    1. Wait for one of the following conditions:
+        * The user closes the browsing context: return failure.
+        * {{IdentityProvider}}.{{IdentityProvider/close}} is called in the
+            context of this new traversable:
+            1. Close the traversable.
+            1. Let |loginStatus| be the result of [=get the login status=]
```
npm1, line 1309: Ah, this is a bit weird in the implementation. Because we do not know whether we process the header before the idp.close, we allow header processing to also work. But I guess the spec is sync enough that it's not possible...

cbiesinger (commented):
oops, just realized I did not publish my responses

bvandersloot-mozilla (approved):

spec/index.bs
```diff
@@ -353,14 +358,19 @@ list with name "<dfn><code>Set-Login</code></dfn>" and type "`item`". If |value|
...
 <div algorithm="process the login status header">
-* Let |origin| be the response's [=response/URL=]'s [=/origin=].
-* If |origin| is not [=same-origin with its ancestors=], ignore the header.
-* Otherwise:
-    * Assert that |value| is a tuple.
-    * Let |token| be the first entry of |value|.
-    * If |token| is `logged-in`, [=set the login status=] for |origin|
+1. Let |origin| be the response's [=response/URL=]'s [=/origin=].
+1. Let |client| be the [=/request=]'s [=request/client=].
+1. If the request is not a [=directly user-initiated request=]:
```
npm1, line 363: I don't think this is what you want, but rather request's destination = document https://fetch.spec.whatwg.org/#concept-request-destination.

Also the flow in this method is now broken, need to get rid of the "Otherwise".

npm1 (approved):

PR #438, opened 2023-02-14
Add multiple IDP support

npm1:
Relevant issue: https://github.com/fedidcg/FedCM/issues/319

This PR adds support for multiple IDPs. It does so by adding some state to the document to allow bundling multiple `get()` calls with potentially multiple providers each. The bundling follows the pattern explained in https://github.com/fedidcg/FedCM/issues/319#issuecomment-1270753874: all calls before onload are bundled together while after onload we just use a task for bundling. There are also checks added to ensure that dupes in configURLs are not allowed. A few methods now return pairs so as to enable tracking the `provider` to which the selected account belongs to, which in turn enables determining which `get()` call to provide the credential to, in case of a successful flow.

***
<a href="https://pr-preview.s3.amazonaws.com/npm1/FedCM/pull/438.html" title="Last updated on Feb 15, 2023, 9:42 PM UTC (9bdfe44)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/438/10c8d09...npm1:9bdfe44.html" title="Last updated on Feb 15, 2023, 9:42 PM UTC (9bdfe44)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -490,10 +490,10 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
                 : <a spec=dom for="event listener">callback</a>
                 :: new {{EventListener}} whose {{EventListener/handleEvent(event)}} invokes
                     [=run pending fedID requests=] with |document|.
-                : <a spec=dom for="event listener">once</a>                
+                : <a spec=dom for="event listener">once</a>
                 :: true
             1. [=Add an event listener=] passing |document| and |fedCMListener|.
-            
+
                 Issue: determine if this is the best way to spec the fact that we need to invoke the
                 method after the {{Document}} has loaded.
 
@@ -515,7 +515,7 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
     1. Wait until |credentialPromise| becomes set, either as an {{IdentityCredential}} or as an
         exception.
     1. Return |credentialPromise|.
-</div> 
+</div>
 
 <div algorithm>
 When asked to <dfn>run pending fedID requests</dfn> with a {{Document}} |document|, run the
@@ -526,7 +526,7 @@ following steps [=in parallel=]:
     1. Set |document|'s [=fedID task status=] to "running".
     1. Let |allProviders| be an initially empty [=list=].
     1. For each (|providers|, |credentialPromise|) in |document|'s [=pending fedID requests=]:
-        1. [=Extend=] |allProviders| with |providers|.        
+        1. [=Extend=] |allProviders| with |providers|.
     1. Run {{WindowOrWorkerGlobalScope/setTimeout()}} passing a [=task=] which throws a
         {{NetworkError}}, after a timeout of 60 seconds.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -453,10 +453,10 @@ NOTE: The {{CredentialRequestOptions/mediation}} flag is currently not used.
 The {{CredentialRequestOptions/signal}} is used as an abort signal for the
 requests.
 
-Each {{Document}} has an associated <dfn>pending fedID requests</dfn>, an initially empty [=list=].
+Each {{Document}} has an associated <dfn>pending fedID request list</dfn>, an initially empty [=list=].
 
 Each {{Document}} has an associated <dfn>fedID task status</dfn>, which is initially set to
-"no task". Its possible values are "no task", "pending", or "running".
+"<code>[=no task=]</code>". Its possible values are "<code><dfn>no task</dfn></code>", "<code><dfn>pending</dfn></code>", or "<code><dfn>running</dfn></code>".
 
 <div algorithm>
 When the {{IdentityCredential}}'s
@@ -472,16 +472,16 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
         {{Credential/[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}}
         algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
 
-    1. Let |credentialPromise| be a variable set to null.
-    1. If |document|'s [=fedID task status=] is "running":
+    1. Let |credentialOrError| be a variable set to null.
+    1. If |document|'s [=fedID task status=] is "<code>[=running=]</code>":
         1. [=Queue a global task=] with |globalObject| on the [=DOM manipulation task source=] to
-            set |credentialPromise| to a new "{{NotAllowedError}}" {{DOMException}}.
-        1. Wait until |credentialPromise| is set, and return it.
-    1. If |document|'s [=fedID task status=] is "no task":
+            set |credentialOrError| to a new "{{NotAllowedError}}" {{DOMException}}.
+        1. Wait until |credentialOrError| is set, and return it.
+    1. If |document|'s [=fedID task status=] is "<code>[=no task=]</code>":
         1. If |document|'s {{Document/readyState}} is "complete":
             1. [=queue a global task=] with |globalObject| on the [=DOM manipulation task source=] to
                 [=run pending fedID requests=] with |document|.
-            1. Set |document|'s [=fedID task status=] to "pending".
+            1. Set |document|'s [=fedID task status=] to "<code>[=pending=]</code>".
         1. Otherwise:
             1. Let |fedCMListener| be an [=event listener=] set as follows:
 
@@ -494,27 +494,27 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
                 :: true
             1. [=Add an event listener=] passing |document| and |fedCMListener|.
 
-                Issue: determine if this is the best way to spec the fact that we need to invoke the
-                method after the {{Document}} has loaded.
+                Issue: [better](https://github.com/fedidcg/FedCM/issues/440) invoke the method after
+                the {{Document}} has loaded.
 
-            1. Set |document|'s [=fedID task status=] to "pending".
+            1. Set |document|'s [=fedID task status=] to "<code>[=pending=]</code>".
     1. Let |providers| be |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"].
     1. Let |allConfigURLs| be an [=ordered set=].
-    1. For each (|provs|, <var ignore="">cred</var>) in |document|'s [=pending fedID requests=]:
-        1. For each |prov| in |provs|:
-            1. [=set/Append=] |prov|'s {{IdentityProviderConfig/configURL}} to
+    1. For each (|providerList|, <var ignore="">cred</var>) in |document|'s [=pending fedID request list=]:
+        1. For each |provider| in |providerList|:
+            1. [=set/Append=] |provider|'s {{IdentityProviderConfig/configURL}} to
                 |allConfigURLs|.
     1. For each |provider| in |providers|:
         1. If |allConfigURLs| [=list/contains=] |provider|'s {{IdentityProviderConfig/configURL}}:
             1. [=Queue a global task=] with |globalObject| on the [=DOM manipulation task source=] to
-                set |credentialPromise| to a new "{{NotAllowedError}}" {{DOMException}}.
-            1. Wait until |credentialPromise| is set, and return it.
+                set |credentialOrError| to a new "{{NotAllowedError}}" {{DOMException}}.
+            1. Wait until |credentialOrError| is set, and return it.
         1. [=set/Append=] |provider|'s {{IdentityProviderConfig/configURL}} to
             |allConfigURLs|.
-    1. [=list/Append=] (|providers|, |credentialPromise|) to |document|'s [=pending fedID requests=].
-    1. Wait until |credentialPromise| becomes set, either as an {{IdentityCredential}} or as an
+    1. [=list/Append=] (|providers|, |credentialOrError|) to |document|'s [=pending fedID request list=].
+    1. Wait until |credentialOrError| becomes set, either as an {{IdentityCredential}} or as an
         exception.
-    1. Return |credentialPromise|.
+    1. Return |credentialOrError|.
 </div>
 
 <div algorithm>
@@ -523,9 +523,9 @@ following steps [=in parallel=]:
     1. The [=user agent=] MAY wait some time before proceeding with the following steps. This
         provides the [=user agent=] with flexibility on how to bundle together multiple federated
         identity requests from a site.
-    1. Set |document|'s [=fedID task status=] to "running".
+    1. Set |document|'s [=fedID task status=] to "<code>[=running=]</code>".
     1. Let |allProviders| be an initially empty [=list=].
-    1. For each (|providers|, |credentialPromise|) in |document|'s [=pending fedID requests=]:
+    1. For each (|providers|, |credentialOrError|) in |document|'s [=pending fedID request list=]:
         1. [=Extend=] |allProviders| with |providers|.
     1. Run {{WindowOrWorkerGlobalScope/setTimeout()}} passing a [=task=] which throws a
         {{NetworkError}}, after a timeout of 60 seconds.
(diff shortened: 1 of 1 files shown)


npm1:
@bvandersloot-mozilla I was hoping you could take a look at this before you go on leave! I know it is a bit late, so if you're unable to get a chance then I will bug your teammates :)

spec/index.bs
```diff
@@ -453,6 +453,11 @@ NOTE: The {{CredentialRequestOptions/mediation}} flag is currently not used.
 The {{CredentialRequestOptions/signal}} is used as an abort signal for the
 requests.
 
+Each {{Document}} has an associated <dfn>pending fedID requests</dfn>, an initially empty [=list=].
```
cbiesinger, line 456: s/an//, since "requests" is plural?

spec/index.bs
```diff
@@ -939,13 +1008,15 @@ dictionary IdentityProviderClientMetadata {
...
-To <dfn>select an account</dfn> given an |accountsList|, run the following steps. This returns an
-{{IdentityProviderAccount}} or failure.
-    1. Assert |accountsList|'s [=list/size=] is greater than 1.
-    1. Display an account chooser displaying the options from |accountsList|.
-    1. Let |account| be the {{IdentityProviderAccount}} of the account that the user
-        manually selects from the accounts chooser, or failure if no account is selected.
-    1. Return |account|.
+To <dfn>select an account</dfn> given |allAccountsAndProviders|, run the following steps. This
+returns an ({{IdentityProviderAccount}}, {{IdentityProviderAPIConfig}}) [=tuple=] or the [=tuple=]
+(failure, failure).
```
cbiesinger, line 1013: You are doing this so that you always return a 2-tuple? I don't think this is necessary... I'd just say "or failure"

spec/index.bs
```diff
@@ -939,13 +1008,15 @@ dictionary IdentityProviderClientMetadata {
...
+    1. Assert |allAccountsAndProviders|'s [=list/size=] is greater than 1.
+    1. Display an account chooser displaying the options from |allAccountsAndProviders|.
+    1. Let (|account|, |provider|) be the {{IdentityProviderAccount}} and associated
+        {{IdentityProviderConfig}} of the account that the user manually selects from the accounts
+        chooser, or (failure, failure) if no account is selected.
+    1. Return (|account|, |provider|.
```
cbiesinger, line 1019: missing ')'

spec/index.bs
```diff
@@ -461,9 +466,67 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
+    1. Let |document| be |globalObject|'s [=associated Document=].
+
+        Note: The |globalObject| is not currently passed onto the
+        {{Credential/[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}}
+        algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
+
+    1. Let |credentialPromise| be a variable set to null.
+    1. If |document|'s [=fedID task status=] is "running":
+        1. [=Queue a global task=] with |globalObject| on the [=DOM manipulation task source=] to
+            set |credentialPromise| to a new "{{NotAllowedError}}" {{DOMException}}.
```
cbiesinger, line 478: Hm, is that how this is supposed to be written? Not something like "to reject credentialPromise with a NotAllowedError"?

spec/index.bs
```diff
@@ -461,9 +466,67 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
+        1. Wait until |credentialPromise| is set, and return it.
+    1. If |document|'s [=fedID task status=] is "no task":
+        1. If |document|'s {{Document/readyState}} is "complete":
```
cbiesinger, line 481: FYI, longer term we should add a callback from the HTML spec to FedCM as described here: https://github.com/whatwg/html/pull/8382#issuecomment-1279929852

(and also get rid of the DOM event listener)

spec/index.bs
```diff
@@ -461,9 +466,67 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
+                :  <a spec=dom for="event listener">type</a>
+                :: "load"
+                : <a spec=dom for="event listener">callback</a>
+                :: new {{EventListener}} whose {{EventListener/handleEvent(event)}} invokes
+                    [=run pending fedID requests=] with |document|.
+                : <a spec=dom for="event listener">once</a>
+                :: true
+            1. [=Add an event listener=] passing |document| and |fedCMListener|.
+
+                Issue: determine if this is the best way to spec the fact that we need to invoke the
```
cbiesinger, line 497: Ah, I see, maybe link that issue I mentioned above here (or I guess we should file a new FedCM issue for it)

spec/index.bs
```diff
@@ -461,9 +466,67 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
+                method after the {{Document}} has loaded.
+
+            1. Set |document|'s [=fedID task status=] to "pending".
+    1. Let |providers| be |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"].
+    1. Let |allConfigURLs| be an [=ordered set=].
+    1. For each (|provs|, <var ignore="">cred</var>) in |document|'s [=pending fedID requests=]:
```
bvandersloot-mozilla, line 503: Can we rename "provs"?

spec/index.bs
```diff
@@ -461,9 +466,67 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
+                |allConfigURLs|.
+    1. For each |provider| in |providers|:
+        1. If |allConfigURLs| [=list/contains=] |provider|'s {{IdentityProviderConfig/configURL}}:
+            1. [=Queue a global task=] with |globalObject| on the [=DOM manipulation task source=] to
+                set |credentialPromise| to a new "{{NotAllowedError}}" {{DOMException}}.
+            1. Wait until |credentialPromise| is set, and return it.
+        1. [=set/Append=] |provider|'s {{IdentityProviderConfig/configURL}} to
+            |allConfigURLs|.
+    1. [=list/Append=] (|providers|, |credentialPromise|) to |document|'s [=pending fedID requests=].
+    1. Wait until |credentialPromise| becomes set, either as an {{IdentityCredential}} or as an
```
bvandersloot-mozilla, line 515: We probably want to use resolve, reject, and settled rather than set for |credentialPromise|.

bvandersloot-mozilla (changes requested):
I have one major structural complaint: I'd like to be able to have a provider-chooser before the account chooser. Could we have an optional provider list filter step after the config fetches but before the account list fetches? This would enable both UI designs

I haven't gone through for bugs or deep-dived on the Bikeshed, and think @cbiesinger had some good points about figuring out how to handle the HTML integration and async events better. But other than the line above, this matches my expectations.

spec/index.bs
```diff
@@ -461,9 +466,67 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
 
     1. Assert: These steps are running [=in parallel=].
     1. Assert: |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"] [=map/exists=].
-    1. Assert: |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"] [=list/size=] is 1.
+    1. Let |document| be |globalObject|'s [=associated Document=].
```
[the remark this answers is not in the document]
npm1: Good question. I think for now we do not want to mix requests from different frames.

spec/index.bs
npm1: No because `DiscoverFromExternalSource` does not expect to receive a promise. Maybe the variable name is confusing, I can change it.

spec/index.bs
npm1: Yea I should probably do that, for the other part (the one using event listener)

spec/index.bs
```diff
@@ -461,9 +466,67 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
+        algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
+
+    1. Let |credentialPromise| be a variable set to null.
+    1. If |document|'s [=fedID task status=] is "running":
+        1. [=Queue a global task=] with |globalObject| on the [=DOM manipulation task source=] to
+            set |credentialPromise| to a new "{{NotAllowedError}}" {{DOMException}}.
+        1. Wait until |credentialPromise| is set, and return it.
+    1. If |document|'s [=fedID task status=] is "no task":
+        1. If |document|'s {{Document/readyState}} is "complete":
+            1. [=queue a global task=] with |globalObject| on the [=DOM manipulation task source=] to
```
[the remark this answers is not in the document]
npm1: It could also be the networking task source since those steps will produce a bunch of fetches. I'm open to change this but not sure if it makes a big difference (as long as similar fedcm steps run on the same task source)

spec/index.bs
npm1: Not a promise (renamed)

spec/index.bs
npm1: It's not necessary but it was cleaner to always return a tuple here instead of maybe returning failure and then having to do checks about it in the caller.

npm1:
> I have one major structural complaint: I'd like to be able to have a provider-chooser before the account chooser. Could we have an optional provider list filter step after the config fetches but before the account list fetches? This would enable both UI designs

Added

> 
> I haven't gone through for bugs or deep-dived on the Bikeshed, and think @cbiesinger had some good points about figuring out how to handle the HTML integration and async events better. But other than the line above, this matches my expectations.

Ah yea I think for now we should monkeypatch the HTML spec instead of what I did. Haven't fixed this part but ptal at the rest

spec/index.bs
```diff
@@ -939,13 +1015,15 @@ dictionary IdentityProviderClientMetadata {
...
-    1. Assert |accountsList|'s [=list/size=] is greater than 1.
-    1. Display an account chooser displaying the options from |accountsList|.
-    1. Let |account| be the {{IdentityProviderAccount}} of the account that the user
-        manually selects from the accounts chooser, or failure if no account is selected.
-    1. Return |account|.
+To <dfn>select an account</dfn> given |allAccountsAndProviders|, run the following steps. This
+returns an ({{IdentityProviderAccount}}, {{IdentityProviderAPIConfig}}) [=tuple=] or the [=tuple=]
+(failure, failure).
+    1. Assert |allAccountsAndProviders|'s [=list/size=] is greater than 1.
+    1. Display an account chooser displaying the options from |allAccountsAndProviders|.
```
asr-enid, line 1022: There is no mentioning here how this dialogue is constructed w.r.t. to ordering suggestions by the RP - This was discussed at length - https://github.com/fedidcg/FedCM/issues/319#issuecomment-1352319800 

@npm1 
> All of these suggestions sound good to me. Having the bar 'show at least after onload' achieves the goal of enabling all IDPs to register while preserving flexibility on the timing. **This coupled with RP provided ordering seems like a good path forward**. I like how this is shaping up!

@bvandersloot-mozilla 
> Given this, I think that is probably the way forward. How about instead of "the onload event" we talk about "the display of IDP chooser, which must occur after the onload event"? This leaves more flexibility for quiet UI and doesn't artificially constrain the "early calls" to before onload when there may be more time. **Also, we could add an option to the RP's ordering suggestion call** that can suggest to the browser to use a louder UI.

The general context was discussed here https://github.com/fedidcg/FedCM/issues/348 and in a fedccg meeting see here - https://github.com/fedidcg/meetings/blob/main/2022/2022-10-10-notes.md

spec/index.bs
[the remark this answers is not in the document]
bvandersloot-mozilla: My thoughts on this depend on the behavior of major IDPs that I don't know off of the top of my head... Do the APIs included via script tags usually iframe themselves off before trying to establish a session on page load? Put another way: does the equivalent of the `google.accounts.id.initialize` call do the meat of its work inside of an iframe for any major IDP? 

If so, I think mixing requests from frames is important to this extension's utility.

bvandersloot-mozilla:
> > I have one major structural complaint: I'd like to be able to have a provider-chooser before the account chooser. Could we have an optional provider list filter step after the config fetches but before the account list fetches? This would enable both UI designs
> 
> Added

👍 TY!

> 
> > I haven't gone through for bugs or deep-dived on the Bikeshed, and think @cbiesinger had some good points about figuring out how to handle the HTML integration and async events better. But other than the line above, this matches my expectations.
> 
> Ah yea I think for now we should monkeypatch the HTML spec instead of what I did. Haven't fixed this part but ptal at the rest

Agreed. Then once this is graduated to a WG (or more stable) we can upstream the monkeypatch. 
I've looked at the rest and it makes logical sense and represents the vision I have.

I don't have the time to deep-dive on the Bikeshed before my leave, unfortunately. This is mostly because of my general unfamiliarity with the weeds of what is allowed "in parallel" and how to play nice with WHATWG HTML.  I think that a lot of the complication here stems from the Credential internal methods being run `in parallel`. Like the "wait for X to be set" lines where promises are natural. However, I don't spot any bugs in the constructions you make and the algorithms you add. Great work here- there is creativity in this I hadn't gotten far enough to see needed in my sketches.

spec/index.bs
bvandersloot-mozilla: I think it would be good to include here given the interest and our initial commitments. 

If we're hanging it off of `navigator.credentials` we should generalize it to be applicable to all credentials.

However, I prefer to put it somewhere like `window.IdentityProvider.setPromptOrdering`, and maybe move the Sign-In status API to `window.IdentityProvider.status.*`. This makes it clear that it is tied to Identity Provider concerns, rather than general credentials.

samuelgoto: > navigator.credentials.setProviderOrdering(["configURL1", "configURL2"]) 

If we start from the assumption that the relying party is going to be making a JS call (which I'm not sure is a good idea), then we can revisit the entirety of this API design here and take one of the other alternatives that don't rely on batching API calls, that is the [`Register` and `Prompt` alternative](https://github.com/fedidcg/FedCM/blob/main/proposals/multi-idp-api.md#1-register-and-prompt-implementation) we considered. 

We could do:

```js
// Called from JS SDKs
let {token} = await IdentityProvider.register({
  configURL: // ...
});
```

And then later the RP can call:

```js
await navigator.credentials.get({
  identity: {
    providers: [{
      order: ["configUrl1", "configUrl2", ...]
    }]
  }
});
```

npm1: > However, I prefer to put it somewhere like `window.IdentityProvider.setPromptOrdering`, and maybe move the Sign-In status API to `window.IdentityProvider.status.*`. This makes it clear that it is tied to Identity Provider concerns, rather than general credentials.

I'm fine with `IdentityProvider.setPromptOrdering()` too, as we don't have use-cases for making this more general to credentials. Although we might in the future.

> If we start from the assumption that the relying party is going to be making a JS call (which I'm not sure is a good idea), then we can revisit the entirety of this API design here and take one of the other alternatives that don't rely on batching API calls, that is the [`Register` and `Prompt` alternative](https://github.com/fedidcg/FedCM/blob/main/proposals/multi-idp-api.md#1-register-and-prompt-implementation) we considered.

I don't think that suggestion works, and the reasons are stated in the section you linked. We need to fulfill these requirements:

1. The API should be easily invocable from IDP SDKs, independently.
2. The IDP ordering should be customizable by an RP that wishes to do so. Note that this customization is entirely optional: an RP does not have to specify an IDP ordering.

Your suggestion does not satisfy 1), a key requirement. We are not assuming that the RP is going to directly make a JS call. I think however it is fair to assume that if the RP wants to specify IDP ordering, then they should be capable of adding a JS call. Note that this JS call only requires very limited info about the IDPs (their origins or configURLs), not the full set of things required from a `get()` call. But this JS call is also not required: the user agent can decide an IDP ordering if the RP provides no signal about their preference.

An alternative I considered is passing an order parameter directly in the `get()` call. I worry that this means the IDP is the one setting the value since they are the ones invoking the FedCM API, and there's not a lot of incentive to respect the desired RP value (an IDP would always want to be shown as the first one).

asr-enid: > An alternative I considered is passing an order parameter directly in the get() call. I worry that this means the IDP is the one setting the value since they are the ones invoking the FedCM API, and there's not a lot of incentive to respect the desired RP value (an IDP would always want to be shown as the first one).

I agree, this could also lead to conflicting values passed to the get calls. Having some sort of independent way for RPs to define their preferences in general seems sensible outside of some defined default behaviour. Making this optional makes sense to me too. I guess this API surface should then also not be available in IDP IFrames.

> However, I prefer to put it somewhere like window.IdentityProvider.setPromptOrdering, and maybe move the Sign-In status API to window.IdentityProvider.status.*. 

Agree - this would also leave room for catering to additional needs of RPs (https://github.com/fedidcg/FedCM/issues/348) by adding additional means to define this via api surfaces to window.IdentityProvider.  (for example a means to differentiate if they want to trigger a sign-in vs. a sign-up flows or don't care (default), which is not really a multi-idp question thus not in scope for this PR)

PR #452, opened 2023-04-11
Adjust links to RFC9110 sections

tidoust:
The cross-reference database used in Bikeshed used to contain the `httpwg.org` version of RFC9110. It now uses the `www.rfc-editor.org` version, consistent with the URL used in the References section.

This update adjusts the links to RFC9110 sections accordingly.

(Note: the cross-reference database contains the actual section IDs, e.g. `section-10.2.2`, not the section heading's IDs such as `name-location`)


cbiesinger (approved):
a bit disappointing that we can't link directly to the right field. numbers don't seem as stable. anyway, thanks!

PR #453, opened 2023-04-12
Replace the 60 second timeout with a random delay

### LINKED ISSUE #389: Do not use setTimeout directly
We should fix the way we say that an error is to be thrown after a certain wait time. See https://github.com/fedidcg/FedCM/pull/378/files#r1052333153

cbiesinger:
This is a better user experience and also gives user agents mores
flexibility for their UI. It also makes the spec match Chrome's current
implementation.

This change does not add a delay when the user closes the dialog
explicitly. This is not necessary because the timing is effectively
random and therefore indistinguishable from other failures that do
have this delay, and it allows RPs to fall back to different types
of authentication immediately in this case.

Closes #389

***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/453.html" title="Last updated on May 9, 2023, 5:16 PM UTC (ae94804)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/453/6ca279e...cbiesinger:ae94804.html" title="Last updated on May 9, 2023, 5:16 PM UTC (ae94804)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -461,22 +461,25 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
         {{Credential/[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}}
         algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
 
-    1. If |credential| is failure, [=queue a global task=] on the [=DOM manipulation task source=]
-        to throw a new "{{NetworkError}}" {{DOMException}}. If |throwImmediately| is true, this
-        MUST happen after a random delay between 0 and 60 seconds.
+    1. If |credential| is a pair:
+        1. Let |throwImmediately| be the value of the second element of the pair.
+        1. [=Queue a global task=] on the [=DOM manipulation task source=]
+            to throw a new "{{NetworkError}}" {{DOMException}}. If |throwImmediately| is true, this
+            MUST happen after a random delay between 0 and 60 seconds.
+    1. Otherwise, return |credential|.
 </div>
 
 <div algorithm>
 To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
 |provider| and a |globalObject|, run the following steps. This returns an {{IdentityCredential}} or
-failure with an optional boolean |throwImmediately|, defaulting to false.
+a pair (failure, bool), where the bool indicates whether to skip delaying the exception throw.
     1. Assert: These steps are running [=in parallel=].
     1. Let |config| be the result of running [=fetch the config file=] with |provider| and
         |globalObject|.
-    1. If |config| is failure, return failure.
+    1. If |config| is failure, return (failure, false).
     1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
         and |globalObject|.
-    1. If |accountsList| is failure, return failure.
+    1. If |accountsList| is failure, return (failure, false).
     1. For each |account| in |accountsList|:
         1. If |account|["{{IdentityProviderAccount/picture}}"] is present,
             [=fetch the account picture=] with |account| and |globalObject|.
@@ -493,23 +496,21 @@ failure with an optional boolean |throwImmediately|, defaulting to false.
             with |account|, |accountState|, |config|, |provider|, and |globalObject|.
         1. Otherwise, show a dialog to request user permission to sign in via |account|, and set the
             result in |permission|.
-        1. If |permission|, [=sign-in=] with |accountState|; otherwise, return failure with
-            |throwImmediately| set to true.
+        1. If |permission|, [=sign-in=] with |accountState|; otherwise, return (failure, true).
     1. Otherwise:
         1. Let |account| be the result of running the [=select an account=] from the
             |accountsList|.
-        1. If |account| is failure, return failure with |throwImmediately| set to true.
+        1. If |account| is failure, return (failure, true).
         1. Let |accountState| be the result of running the [=compute account state=] algorithm
             given |provider| and |account|.
         1. If |accountState|'s {{AccountState/registration state}} is {{unregistered}}:
             1. Let |permission| be the result of running the [=request permission to sign-up=]
                 algorithm with |account|, |accountState|, |config|, |provider|, and |globalObject|.
-            1. If |permission|, [=sign-in=] with |accountState|; otherwise, return failure with
-                |throwImmediately| set to true.
+            1. If |permission|, [=sign-in=] with |accountState|; otherwise, return (failure, true).
         1. Otherwise, [=sign-in=] with |accountState|.
     1. Wait until the [=user agent=]'s dialog is closed.
     1. If |accountState|'s {{AccountState/registration state}} is {{unregistered}} then return
-        failure.
+        (failure, false).
     1. Let |credential| be the result of running the [=fetch an identity assertion=] algorithm with
         |accountState|, |account|'s {{IdentityProviderAccount/id}}, |provider|, |config|, and
         |globalObject|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -463,16 +463,19 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
 
     1. If |credential| is a pair:
         1. Let |throwImmediately| be the value of the second element of the pair.
+        1. If |throwImmediately| is false, the user agent SHOULD wait a random
+            amount of time between 0 and 60 seconds before the next step, unless
+            it has implemented another way to prevent exposing to the RP whether
+            the user has an account logged in to the RP.
         1. [=Queue a global task=] on the [=DOM manipulation task source=]
-            to throw a new "{{NetworkError}}" {{DOMException}}. If |throwImmediately| is true, this
-            MUST happen after a random delay between 0 and 60 seconds.
+            to throw a new "{{NetworkError}}" {{DOMException}}.
     1. Otherwise, return |credential|.
 </div>
 
 <div algorithm>
 To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
 |provider| and a |globalObject|, run the following steps. This returns an {{IdentityCredential}} or
-a pair (failure, bool), where the bool indicates whether to skip delaying the exception throw.
+a pair (failure, bool), where the bool indicates whether to skip delaying the exception thrown.
     1. Assert: These steps are running [=in parallel=].
     1. Let |config| be the result of running [=fetch the config file=] with |provider| and
         |globalObject|.
@@ -509,8 +512,6 @@ a pair (failure, bool), where the bool indicates whether to skip delaying the ex
             1. If |permission|, [=sign-in=] with |accountState|; otherwise, return (failure, true).
         1. Otherwise, [=sign-in=] with |accountState|.
     1. Wait until the [=user agent=]'s dialog is closed.
-    1. If |accountState|'s {{AccountState/registration state}} is {{unregistered}} then return
-        (failure, false).
     1. Let |credential| be the result of running the [=fetch an identity assertion=] algorithm with
         |accountState|, |account|'s {{IdentityProviderAccount/id}}, |provider|, |config|, and
         |globalObject|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -467,6 +467,14 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
             amount of time between 0 and 60 seconds before the next step, unless
             it has implemented another way to prevent exposing to the RP whether
             the user has an account logged in to the RP.
+
+            Note: The intention here is as follows. If we resolved the promise immediately,
+                then an RP could infer that no dialog was shown because the promise was
+                resolved quickly, after just the network delay. A shown dialog implies
+                that the user is logged in to one or more accounts of the IDP. To prevent
+                this information leakage, especially without user confirmation, we specify
+                this delay. However, UAs may have different UI approaches here and prevent
+                it in a different way.
         1. [=Queue a global task=] on the [=DOM manipulation task source=]
             to throw a new "{{NetworkError}}" {{DOMException}}.
     1. Otherwise, return |credential|.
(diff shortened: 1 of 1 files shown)


cbiesinger:
@rbyers @samuelgoto @yi-gu

spec/index.bs
```diff
@@ -484,6 +475,11 @@ failure.
     1. If |config| is failure, return failure.
     1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
         and |globalObject|.
+    1. If |accountsList|'s size is 0 or if the fetch failed:
```
yi-gu, line 478: or the account list response cannot be parsed (malformed or missing required fields). I guess it's all about "if the browser does not have sufficient user information to display on the UI". WDYT?

cbiesinger: I changed the text to be more precise. Most of the processing steps already threw an exception if something is malformed, so I updated the one step I could find that did not do so, to do so as well. WDYT?

yi-gu: Sounds good!

cbiesinger: For the benefit of anyone else -- if a required field is missing the exception is thrown here:
https://webidl.spec.whatwg.org/#es-dictionary (4.1.5)
via
https://fedidcg.github.io/FedCM/#fetch-the-accounts-list (5.2, "Convert" link)

spec/index.bs
```diff
@@ -484,6 +475,11 @@ failure.
     1. If |config| is failure, return failure.
     1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
         and |globalObject|.
+    1. If |accountsList|'s size is 0 or if the fetch threw an exception:
+        1. Stop processing the algorithm and wait indefinitely.
```
npm1, line 479: One issue I see with this is that we do not allow new FedCM calls when there is a pending call. So this effectively means that the FedCM API cannot be invoked again in this page load. Other than that, this might be ok though it seems to require an implementation change on the Chromium side, right? Also, would we need to note somewhere that this would also be the behavior when the user closes/dismisses the user agent UI?

spec/index.bs
```diff
@@ -538,11 +534,17 @@ To <dfn>parse url</dfn> given a {{USVString}} |stringUrl|, a |globalObject|, and
 <div algorithm>
 To <dfn>fetch request</dfn> given a [=/request=] |request|, |globalObject|, and an algorithm
 |processResponseConsumeBody|, run the following steps:
+  1. Set |response| to null.
```
npm1, line 537:
```suggestion
  1. Let |response| be null.
```
I think it is clearer to use this when announcing a new variable.

spec/index.bs
```diff
@@ -538,11 +534,17 @@ To <dfn>parse url</dfn> given a {{USVString}} |stringUrl|, a |globalObject|, and
...
   1. [=Queue a global task=] on the [=network task source=] given |globalObject| to:
-      1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to
-          |processResponseConsumeBody|.
+      1. Set |response| to the result of [=fetch=]ing |request| with <a
```
npm1, line 539: I don't think this is right. Fetch does not return the response, it seems to return a fetch controller.

spec/index.bs
RByers: Good point, and IMHO one that gets at the heart of why I think it's good to move the spec in this direction. Imagine an implementation which chooses to suppress or show very quiet UI when FedCM is invoked without a user gesture, but will show noisier UI when invoked in response to eg. a "login with X" button click. Such an implementation needs a way to cancel the first request and show the second (possibly with <60 seconds between them), right?

cbiesinger: We do have a way to cancel the UI: https://w3c.github.io/webappsec-credential-management/#dom-credentialrequestoptions-signal

RByers: Ah, of course. So is it OK to leave aborting current requests up to the page, or should we auto-abort when a new request comes in? In particular, thinking about multi-IDP scenarios with two un-cooperating SDKs on the same page. Presumably the one with a user gesture should take priority over one without (regardless of whether it's before or after), right?

cbiesinger: That's a good question, hmm

Currently the spec requires us to fail with NotAllowedErr (step 5 in https://w3c.github.io/webappsec-credential-management/#algorithm-store) and payments also fails a second request if the first one is not done yet

spec/index.bs

spec/index.bs

spec/index.bs
npm1: So for the hypothetical scenario where the page uses a FedCM call during the page load and then there is a click which triggers another FedCM call, we would have two options:
1. Remove the UI from the first FedCM call (possibly reject the promise, or just leave it unresolved) and initiate the second call. This has the advantage that there is no need to manually abort the first FedCM call. 
2. Force the developer to abort the first FedCM call manually, otherwise reject the second call (this is the current behavior). This has the advantage that we don't have to guess which FedCM calls should take priority, but does seem to require a bit more work from the RP. It also assumes the IDP does provide an abort method to cancel the call.

spec/index.bs
```diff
@@ -538,11 +534,17 @@ To <dfn>parse url</dfn> given a {{USVString}} |stringUrl|, a |globalObject|, and
...
-      1. [=Fetch=] |request| with <a spec=fetch>processResponseConsumeBody</a> set to
-          |processResponseConsumeBody|.
+      1. [=Fetch=]ing |request| with <a spec=fetch>processResponseConsumeBody</a> set to:
+          1. Set |response| to the |response| parameter.
+
+    Note: We queue a task since the [=fetch=] needs to be run within a task, not [=in parallel=].
+
+  1. Wait for |response| to be set.
+  1. If |response| is a [=network error=], throw a new "{{NetworkError}}" {{DOMException}}.
+  1. Call |processResponseConsumeBody| with response.
```
npm1, line 546: response here is a variable.

spec/index.bs
cbiesinger: I have updated the PR to not have the indefinite timeout. Please take another look. Also interested in thoughts from others (@martinthomson / @cboozar ?)

spec/index.bs
```diff
@@ -471,13 +462,14 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
         algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
 
     1. If |credential| is failure, [=queue a global task=] on the [=DOM manipulation task source=]
-        to throw a new "{{NetworkError}}" {{DOMException}}.
+        to throw a new "{{NetworkError}}" {{DOMException}}. If |throwImmediately| is true, this
```
npm1, line 465: Where does `throwImmediately` come from? This is the first mention and it asks its value...

spec/index.bs
```diff
@@ -471,13 +462,14 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
+        MUST happen after a random delay between 0 and 60 seconds.
 </div>
 
 <div algorithm>
 To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
 |provider| and a |globalObject|, run the following steps. This returns an {{IdentityCredential}} or
-failure.
+failure with an optional boolean |throwImmediately|, defaulting to false.
```
npm1, line 472: So `throwImmediately` is part of the return here? Then perhaps we can return a tuple/pair here `(credential, throwImmediately)`?

spec/index.bs
```diff
@@ -501,17 +493,19 @@ failure.
             with |account|, |accountState|, |config|, |provider|, and |globalObject|.
         1. Otherwise, show a dialog to request user permission to sign in via |account|, and set the
             result in |permission|.
-        1. If |permission|, [=sign-in=] with |accountState|.
+        1. If |permission|, [=sign-in=] with |accountState|; otherwise, return failure with
```
npm1, line 496: Return (failure, true) etc.

spec/index.bs
```diff
@@ -501,17 +493,19 @@ failure.
...
-        1. If |account| is failure, return failure.
+        1. If |account| is failure, return failure with |throwImmediately| set to true.
         1. Let |accountState| be the result of running the [=compute account state=] algorithm
             given |provider| and |account|.
         1. If |accountState|'s {{AccountState/registration state}} is {{unregistered}}:
             1. Let |permission| be the result of running the [=request permission to sign-up=]
                 algorithm with |account|, |accountState|, |config|, |provider|, and |globalObject|.
-            1. If |permission|, [=sign-in=] with |accountState|.
+            1. If |permission|, [=sign-in=] with |accountState|; otherwise, return failure with
+                |throwImmediately| set to true.
```
npm1, line 508: Seems this is just used for the failure cases so maybe actually this should return an IdentityCredential or a boolean (for failure cases), or IdentityCredential or a pair (failure, boolean)?

spec/index.bs
cbiesinger: Restructured as you suggested.

cbiesinger:
Please take another look!

spec/index.bs
```diff
@@ -470,21 +461,25 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
         {{Credential/[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}}
         algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
 
-    1. If |credential| is failure, [=queue a global task=] on the [=DOM manipulation task source=]
-        to throw a new "{{NetworkError}}" {{DOMException}}.
+    1. If |credential| is a pair:
+        1. Let |throwImmediately| be the value of the second element of the pair.
+        1. [=Queue a global task=] on the [=DOM manipulation task source=]
+            to throw a new "{{NetworkError}}" {{DOMException}}. If |throwImmediately| is true, this
+            MUST happen after a random delay between 0 and 60 seconds.
```
npm1, line 468: A couple of comments:
* I think it is a bit clearer to add the wait before the queue a global task step, otherwise. this sounds like this delay could happen in main thread! Something like `1. If |throwImmediately| is true, wait ... 1. Queue a global task`
* Do we need to hardcode the 60 seconds here? In fact, does the delay here need to be mandatory? I can imagine user agent UI such that it only attempts fetching once the user has interacted enough with the dialog, and may thus not need this delay. Perhaps making this wait optional and explaining the purpose is a better approach?

spec/index.bs
```diff
@@ -470,21 +461,25 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
+    1. Otherwise, return |credential|.
 </div>
 
 <div algorithm>
 To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
 |provider| and a |globalObject|, run the following steps. This returns an {{IdentityCredential}} or
-failure.
+a pair (failure, bool), where the bool indicates whether to skip delaying the exception throw.
```
npm1, line 475:
```suggestion
a pair (failure, bool), where the bool indicates whether to skip delaying the exception thrown.
```

spec/index.bs
```diff
@@ -501,21 +496,21 @@ failure.
...
         1. If |accountState|'s {{AccountState/registration state}} is {{unregistered}}:
             1. Let |permission| be the result of running the [=request permission to sign-up=]
                 algorithm with |account|, |accountState|, |config|, |provider|, and |globalObject|.
-            1. If |permission|, [=sign-in=] with |accountState|.
+            1. If |permission|, [=sign-in=] with |accountState|; otherwise, return (failure, true).
         1. Otherwise, [=sign-in=] with |accountState|.
     1. Wait until the [=user agent=]'s dialog is closed.
     1. If |accountState|'s {{AccountState/registration state}} is {{unregistered}} then return
-        failure.
+        (failure, false).
```
npm1, line 513: I'm having a hard time following: now that you added the return (failure, true) above, when can this step happen?

spec/index.bs

spec/index.bs
cbiesinger: Removed.

spec/index.bs
cbiesinger: Done, but let me know if you want this less prescriptive

spec/index.bs
```diff
@@ -470,21 +461,28 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
         {{Credential/[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}}
         algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
 
-    1. If |credential| is failure, [=queue a global task=] on the [=DOM manipulation task source=]
-        to throw a new "{{NetworkError}}" {{DOMException}}.
+    1. If |credential| is a pair:
+        1. Let |throwImmediately| be the value of the second element of the pair.
+        1. If |throwImmediately| is false, the user agent SHOULD wait a random
+            amount of time between 0 and 60 seconds before the next step, unless
+            it has implemented another way to prevent exposing to the RP whether
```
npm1, line 468: Maybe we can be a bit clearer here: the wait is to prevent the caller from distinguishing between the user dismissing the user agent UI and there being no UI to show because the FedCM flow failed at some point and no accounts were successfully fetched. We need to ask other user agents what they think about this step.

npm1 (commented):
Looks good! We should get someone from Mozilla to take a look before merging though...

cbiesinger:
@martinthomson, @bvandersloot-mozilla and @cboozar -- I'd love to hear your thoughts on this PR

spec/index.bs
```diff
@@ -470,21 +461,36 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
         {{Credential/[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)}}
         algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
 
-    1. If |credential| is failure, [=queue a global task=] on the [=DOM manipulation task source=]
-        to throw a new "{{NetworkError}}" {{DOMException}}.
+    1. If |credential| is a pair:
+        1. Let |throwImmediately| be the value of the second element of the pair.
+        1. If |throwImmediately| is false, the user agent SHOULD wait a random
+            amount of time between 0 and 60 seconds before the next step, unless
```
npm1, line 467: Probably no need to specify the 60 seconds here?

spec/index.bs
npm1: Yep thanks!

npm1 (approved):

spec/index.bs

npm1:
Thanks for the review! Merging.

PR #454, opened 2023-04-12
Rework sign-in status API explainer background

RByers:
This is my first attempt in trying to reduce Chrome implementation details in FedCM explainers and specs.

### DIFF
--- a/proposals/idp-sign-in-status-api.md
+++ b/proposals/idp-sign-in-status-api.md
@@ -8,7 +8,7 @@ information to IDPs or RPs which could enable cross-site tracking of the user.
 Because similar consent dialogs can lead to permission blindness and user
 frustration, browsers rely on various implementation-defined heuristics to
 balance user consent fatigue with privacy risk (eg. in the implementation of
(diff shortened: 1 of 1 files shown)


proposals/idp-sign-in-status-api.md
```diff
@@ -2,32 +2,34 @@
...
+a cross-origin URL that passes some information in a query parameter, browsers
+generally rely on the fact that the navigation is visible to the user, rather
+than requesting consent to share information with the target origin. FedCM aims
+to enable such optimizations without specifying the precise behavior which
+browsers are likely to evolve independently over time. Such notification in
+FedCM contexts presents a problem when the user is currently not signed into a
+given IDP as resulting UI is likely to be of low value to the user, eg. "You are
+not signed into IDP X and so no sign-in UI can be shown".
 
 To solve this problem (and to slightly optimize network traffic), we propose
```
samuelgoto, line 25:
```suggestion
To add more signals that user agents could use to address this problem, we propose
```

Maybe the spec can be less opinionated that this is necessarily the only/good solution to solving this problem?

(also, I'd remove the optimization to network traffic)

PR #455, opened 2023-04-19
Correctly handle fetch errors

cbiesinger:
The extract and convert steps can throw exceptions, but fetch's
processResponseConsumeBody does not really handle that. Instead,
set the respective return objects to failure and update the call
site where necessary.

This was split out of PR #453 with major modifications.

***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/455.html" title="Last updated on Apr 19, 2023, 6:29 PM UTC (8ec2e3a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/455/9ce095f...cbiesinger:8ec2e3a.html" title="Last updated on Apr 19, 2023, 6:29 PM UTC (8ec2e3a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -484,6 +484,7 @@ failure.
     1. If |config| is failure, return failure.
     1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
         and |globalObject|.
+    1. If |accountsList| is failure, return failure.
     1. For each |account| in |accountsList|:
         1. If |account|["{{IdentityProviderAccount/picture}}"] is present,
             [=fetch the account picture=] with |account| and |globalObject|.
@@ -592,8 +593,9 @@ To <dfn>fetch the config file</dfn> given an {{IdentityProviderConfig}} |provide
             |responseBody|.
         1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderWellKnown}},
             |discovery|.
-        1. If the [=list/size=] of |discovery|["{{IdentityProviderWellKnown/provider_urls}}"]
-            is greater than 1, set |config| to failure.
+        1. If one of the previous two steps threw an exception, or if the
+            [=list/size=] of |discovery|["{{IdentityProviderWellKnown/provider_urls}}"] is
+            greater than 1, set |config| to failure.
 
             Issue: [relax](https://github.com/fedidcg/FedCM/issues/333) the size of the
             provider_urls array.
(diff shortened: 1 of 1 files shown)


cbiesinger:
@martinthomson, @bvandersloot-mozilla and @cboozar -- FYI, but I don't expect this to be controversial.

cbiesinger:
This text is incorrect, please hold off on reviewing until I have fixed it

cbiesinger:
OK this should be ready for review now.

spec/index.bs
```diff
@@ -592,8 +593,9 @@ To <dfn>fetch the config file</dfn> given an {{IdentityProviderConfig}} |provide
...
         1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderWellKnown}},
             |discovery|.
-        1. If the [=list/size=] of |discovery|["{{IdentityProviderWellKnown/provider_urls}}"]
-            is greater than 1, set |config| to failure.
+        1. If one of the previous two steps threw an exception, or if the
+            [=list/size=] of |discovery|["{{IdentityProviderWellKnown/provider_urls}}"] is
+            greater than 1, set |config| to failure.
 
             Issue: [relax](https://github.com/fedidcg/FedCM/issues/333) the size of the
             provider_urls array.
```
npm1, line 601: I think the next step is exclusive of this one so perhaps change it to start with "Otherwise, if ..."

spec/index.bs
```diff
@@ -806,6 +811,8 @@ To <dfn>fetch an identity assertion</dfn> given an [=AccountState=] |accountStat
         1. Let |json| be the result of [=extract the JSON fetch response=] from |response| and
             |responseBody|.
         1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderToken}}, |token|.
+        1. If one of the previous two steps threw an exception, set |credential| to failure
+            and return.
```
npm1, line 815: We need to modify the initial description of the methods to say that they may return the object, or failure. For instance this one says it returns a credential (but now it can return failure too)

spec/index.bs
```diff
@@ -915,6 +922,7 @@ an {{IdentityProviderConfig}} |provider|, run the following steps. This returns
             |responseBody|.
         1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderClientMetadata}},
             and store the result in |metadata|.
+        1. If one of the previous two steps threw an exception, set |metadata| to null.
```
npm1, line 925: |metadata| starts set as null, and we then wait for it to be set. It's a bit unclear how this happens since it can be set to null? Perhaps we can set to failure, return an object or failure, and handle failure appropriately in the caller?

npm1 (approved):
LGTM. Do you want to wait for someone else to review as well or merge this since it is not really changing anything, just making things more precise?

cbiesinger:
fine to merge IMO

(I personally do not have permission to merge stuff here)

PR #458, opened 2023-04-26
Allow auto-reauthentication in FedCM

npm1:
Fixes https://github.com/fedidcg/FedCM/issues/429

***
<a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/458.html" title="Last updated on May 3, 2023, 8:35 PM UTC (0c811a6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/458/c2277e6...0c811a6.html" title="Last updated on May 3, 2023, 8:35 PM UTC (0c811a6)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -527,13 +527,13 @@ following steps. This returns an {{IdentityCredential}} or failure.
         Note: The [=user agent=] may choose to show UI which does not initially require fetching the
         account pictures. In these cases, the [=user agent=] may delay these fetches until they are
         needed. Because errors from these fetches are ignored, they can happen in any order.
-    1. Let |registeredAccount|, |numRegistedAccounts| be null and 0, respectively.
+    1. Let |registeredAccount|, |numRegisteredAccounts| be null and 0, respectively.
     1. Let |account| be null.
     1. For each |acc| in |accountsList|:
         1. Let |accState| be the result of running the [=compute account state=] algorithm given
             |provider| and |acc|.
         1. If |accState| is {{registered}}, set |registeredAccount| to |acc| and increase
-            |numRegistedAccounts| by 1.
+            |numRegisteredAccounts| by 1.
     1. Let |permission| be false.
     1. If |mediation| is not "{{CredentialMediationRequirement/required}}", |requiresUserMediation|
         is false, and |numRegisteredAccounts| is equal to 1:


yi-gu:
Do we need to update the algorithm to take mediation requirements into account? Or can it be implied?

npm1:
Hmm I thought we might get the logic for free from credential management but this appears to not be the case. We currently have all our stuff in `collecting Credentials from the credential store` since it requires the user to tell the user agent what IDP and account they would like to use. Credential management would only work for free if we had some of the logic in `asking the user to choose a Credential`. But I don't think we can move this because otherwise having only 1 possible credential with mediation equal to "optional" or "silent" would mean that we sign in the user without them having provided any consent before. So we probably need to keep doing all our stuff in `collecting Credentials from the credential store` and our `asking the user to choose a Credential` is a no-op, but that does mean we need to specify the logic in the spec. I can take a stab at it in this PR.

npm1:
Ok, I took a stab at it, please take a look? I can loop in Mozilla folks once it looks good to you.

spec/index.bs
```diff
@@ -476,25 +476,43 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
 To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderConfig}}
-|provider| and a |globalObject|, run the following steps. This returns an {{IdentityCredential}} or
-failure.
+|provider|, a {{CredentialRequestOptions/mediation}} |mediation|, and a |globalObject|, run the
+following steps. This returns an {{IdentityCredential}} or failure.
     1. Assert: These steps are running [=in parallel=].
+    1. Let |requiresUserMediation| be |provider|'s {{IdentityProviderConfig/configURL}}'s [=/origin=]'s
+        [=requires user mediation=].
+    1. If |requiresUserMediation| is true and |mediation| is
+        "{{CredentialMediationRequirement/silent}}", return failure.
```
yi-gu, line 485: Should this be `conditional`?

yi-gu: Ah it's `silent` indeed. But we may add one for `conditional`.
Also, shall we avoid using `requiresUserMediation` given that the function has been deprecated/replaced by `preventSilentAccess`? https://w3c.github.io/webappsec-credential-management/#dom-credentialscontainer-preventsilentaccess

npm1: We do not need one for conditional since it is handled in step 4 of https://w3c.github.io/webappsec-credential-management/#algorithm-request.

`requires user mediation` (not `requiresUserMediation`) is not deprecated and is used in step 12.3.2 in the CredMan algorithm, so I used it for consistency (it is equivalent to using `prevent silent access flag` per the definition).

yi-gu: SG

npm1:
@bvandersloot-mozilla  @cboozar @martinthomson - requesting Mozilla review! As y'all had requested, we're not filing a standards position issue. So consider this it :) Please take a look.

spec/index.bs
```diff
@@ -476,25 +476,43 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
+            with |acc| and |globalObject|.
 
         Note: The [=user agent=] may choose to show UI which does not initially require fetching the
         account pictures. In these cases, the [=user agent=] may delay these fetches until they are
         needed.
-    1. If |accountsList|'s size is 1:
-        1. Let |account| be |accountsList|[0].
-        1. Let |accountState| be the result of running the [=compute account state=] algorithm
+    1. Let |account|, |accountState| be initially set to null.
+    1. Let |registeredAccount|, |numRegistedAccounts| be null and 0, respectively.
```
cbiesinger, line 500:
```suggestion
    1. Let |registeredAccount|, |numRegisteredAccounts| be null and 0, respectively.
```

spec/index.bs
```diff
@@ -476,25 +476,43 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
+    1. For each |acc| in |accountsList|:
+        1. Let |accState| be the result of running the [=compute account state=] algorithm given
+            |provider| and |acc|.
+        1. If |accState| is {{registered}}, set |registeredAccount| to |acc| and increase
+            |numRegistedAccounts| by 1.
+    1. If |mediation| is not "{{CredentialMediationRequirement/required}}", |requiresUserMediation|
+        is false, and |numRegistedAccounts| is equal to 1:
```
cbiesinger, line 507:
```suggestion
        is false, and |numRegisteredAccounts| is equal to 1:
```

spec/index.bs
```diff
@@ -476,25 +476,43 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
-    1. If |accountsList|'s size is 1:
-        1. Let |account| be |accountsList|[0].
-        1. Let |accountState| be the result of running the [=compute account state=] algorithm
+    1. Let |account|, |accountState| be initially set to null.
+    1. Let |registeredAccount|, |numRegistedAccounts| be null and 0, respectively.
+    1. For each |acc| in |accountsList|:
+        1. Let |accState| be the result of running the [=compute account state=] algorithm given
+            |provider| and |acc|.
+        1. If |accState| is {{registered}}, set |registeredAccount| to |acc| and increase
+            |numRegistedAccounts| by 1.
```
cbiesinger, line 505:
```suggestion
            |numRegisteredAccounts| by 1.
```

cbiesinger (approved):

cboozar:
I think these changes make sense, not sure why some of the variables needed shortening but readable either way

npm1:
Thanks for the review, Cameron! Merging then. Re variable shortening, perhaps you mean `acc`, the reason is that I wanted it to be different from `account` and I'm bad at coming up with names. Happy to get suggestions on the naming.

PR #459, opened 2023-04-26
Let user agents delay fetching the endpoints

cbiesinger:
***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/459.html" title="Last updated on May 9, 2023, 5:40 PM UTC (315d6ea)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/459/c8609ec...cbiesinger:315d6ea.html" title="Last updated on May 9, 2023, 5:40 PM UTC (315d6ea)">Diff</a>


cbiesinger:
@martinthomson, @bvandersloot-mozilla and @cboozar -- any thoughts?

cbiesinger:
also @rbyers

spec/index.bs
```diff
@@ -463,6 +463,11 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
         method to throw an error. If there was no such timer, the developer could easily infer
         whether the user has an account with the [=IDP=] or not, or whether the user closed the UI without granting permission to share the [=IDP=] account information with the [=RP=].
     1. Let |provider| be |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"][0].
+    1. The user agent can decide when to continue with the next steps.
```
npm1, line 466: Presumably this should go before setting |provider|, since you are saying this is to select the |provider| used. Also, perhaps it can be a bit clearer about what the wait is about instead of having a note: `The user agent MAY show a UI to for the user to select the identity provider to be used. If it does this, the user agent SHOULD wait until such UI is closed. If the user did not provide a selection, [=queue a global task=] on the [=DOM manipulation task source=] to throw a new "{{NetworkError}}" {{DOMException}}.` This is awkward because we currently do not support multiple IDPs, so this would at the moment be a single IDP chooser. It's still fine, but will make much more sense once we add multi IDP support.

cbiesinger: Hm, I don't like that wording for two reasons:
- It is still very prescriptive on why the user agent may wait. In particular it does not allow a UA to choose the more quiet UI of just showing a symbol in the URL bar the user may click
- I don't understand the part about "SHOULD wait"? why not MUST?

I did move this step earlier as suggested.

npm1: Ahh yes I missed that example in the note, then this looks good. Hopefully Mozilla enjoys the added freedom this provides to user agents in the spec! :)

npm1 (approved):

martinthomson (commented):
This is fine, at least in isolation, but there is a bigger question to ask here.

For this flow to succeed, it is going to show a selector in some cases and not others.  The involvement of a user is critical at the point that the RP-IdP connection is first made, and less so thereafter.

So an initial request will likely always involve prompting.  My (strong) preference is that this both require and consume engagement if the prompt is shown in that case.

For the re-login prompt, my (again strong) preference is that a prompt NOT be shown.  A prompt only really makes sense if the RP wants to ask for a different login to the one provided recently, or it is asking for something new of the user with respect to how they interact with the IdP (we're not talking Authorization yet, so maybe set this aside).

This seems to be trying to split the difference without changing how the API is invoked, which might be fine, but it seems like including an API hook would help to split the different scenarios.

Then, the site can remember if a user previously logged in with an IdP and it can invoke the correct version.  If it gets it wrong and chooses relogin when there is no existing relationship, it fails.  If it attempts to establish a new relationship when there is one, then the browser shows UX, but can present any previously-selected option in a way that allows the user to choose that same option again if they prefer that.

If your goal is to allow sites to pop this dialog without user interaction, that's gross, but it's your browser, I guess.  If you want to do that, we might be inclined to do the button thing.  That has no bearing on the new/re-up split though.

npm1:
> This is fine, at least in isolation, but there is a bigger question to ask here.

Just to be clear: are you ok merging this as an improvement, and working on the other points separately?

> For this flow to succeed, it is going to show a selector in some cases and not others. The involvement of a user is critical at the point that the RP-IdP connection is first made, and less so thereafter.
> 
> So an initial request will likely always involve prompting. My (strong) preference is that this both require and consume engagement if the prompt is shown in that case.

I think we have discussed this in the past. We don't seem to agree on this point (whether FedCM should be gated on user gesture), but I thought we agreed that user agents could diverge on this at least for now?
 
> For the re-login prompt, my (again strong) preference is that a prompt NOT be shown. A prompt only really makes sense if the RP wants to ask for a different login to the one provided recently, or it is asking for something new of the user with respect to how they interact with the IdP (we're not talking Authorization yet, so maybe set this aside).

Agreed, though this is probably best left to the mediation parameter. Shameless plug: PTAL at https://github.com/fedidcg/FedCM/pull/458? :)

martinthomson:
I'm OK with merging, but requesting that we open an issue to track the bigger issue.

npm1:
> I'm OK with merging, but requesting that we open an issue to track the bigger issue.

Filed https://github.com/fedidcg/FedCM/issues/469 for you :)

PR #460, opened 2023-04-26
Fix assertion in DiscoverFromExternalSource algorithm

RByers:
Make `providers` required in the `identity` options dictionary so that the assert is true.
Replace the `providers` size assert with an explicit check and throw, matching chromium code.

***
<a href="https://pr-preview.s3.amazonaws.com/RByers/FedCM/pull/460.html" title="Last updated on Apr 26, 2023, 8:59 PM UTC (5345414)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/460/a2437e0...RByers:5345414.html" title="Last updated on Apr 26, 2023, 8:59 PM UTC (5345414)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -378,7 +378,7 @@ pre-registered with (i.e. it has a `clientId`).
 
 <xmp class=idl>
 dictionary IdentityCredentialRequestOptions {
-  sequence<IdentityProviderConfig> providers;
+  required sequence<IdentityProviderConfig> providers;
 };
 </xmp>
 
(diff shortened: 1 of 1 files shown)


cbiesinger:
Looks good except that I am not sure if we need to create the exception in a task

npm1:
I think this change makes sense but it actually does not match Chromium implementation (`required` is missing from the IDL file). I guess the only difference is that the implementation currently will just silently return a trivial promise instead of throwing an error.

spec/index.bs
```diff
@@ -450,7 +450,7 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
 
     1. Assert: These steps are running [=in parallel=].
     1. Assert: |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"] [=map/exists=].
-    1. Assert: |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"] [=list/size=] is 1.
+    1. If |options|["{{CredentialRequestOptions/identity}}"]["{{IdentityCredentialRequestOptions/providers}}"] [=list/size=] is not 1, throw a "{{TypeError}}" {{DOMException}}.
```
npm1, line 453: You cannot throw an error while in parallel, need to do it in a task (see below for an example).

npm1:
Filed https://bugs.chromium.org/p/chromium/issues/detail?id=1440522 by the way.

npm1:
This is still relevant. Are you planning on addressing the comments or should I start a new PR for this?

npm1:
I see spec now has 'required' which we added elsewhere already. This also asserts the size, which we would get rid of when we add multi IDP support. So closing this

PR #462, opened 2023-04-28
Editorial: some improvements to FedCM spec

npm1:
I had a chat with @jyasskin and he gave some ideas to improve the FedCM spec. While we didn't finish going over the spec, I think this is a good start.


spec/index.bs
```diff
@@ -178,85 +181,87 @@ API to be used for tracking purposes, while preserving the functionality of
...
-      |                               |                                 |
-     -+-                             -+-                               -+-
+    Relying                             User                            Identity
+    Party                               Agent                           Provider
+
+      |                                   |                                 |
+      | "navigator.credentials.get({    " | "(1) The RP invokes the API."   |
+      | "  identity: {                  " |                                 |
+      | "    providers: [{              " |                                 |
+      | "      configURL: "config.json"," |                                 |
```
samuelgoto, line 191:
```suggestion
      | "      configURL: 'config.json'," |                                 |
```

spec/index.bs
```diff
@@ -178,85 +181,87 @@ API to be used for tracking purposes, while preserving the functionality of
...
+      | "      clientId: "client_id",   " |                                 |
```
samuelgoto, line 192:
```suggestion
      | "      clientId: clientId,   " |                                 |
```

spec/index.bs
```diff
@@ -178,85 +181,87 @@ API to be used for tracking purposes, while preserving the functionality of
...
+      | "      nonce: "nonce",          " |                                 |
```
samuelgoto, line 193:
```suggestion
      | "      nonce: nonce,          " |                                 |
```

spec/index.bs
```diff
@@ -178,85 +181,87 @@ API to be used for tracking purposes, while preserving the functionality of
...
+      |                                   |                                 |
+      | "(4) The browser proceeds to    " | "GET /accounts"                 |
+      | "fetch the list of accounts     " | "Cookie: name=value"            |
+      | "that the user is logged in     " |-------------------------------->|
+      | "to.                            " |<--------------------------------| "{"
+      |                                   |                                 |   "accounts: [ ... ]"
+      |                                   |                                 | "}"
+      |                                   |                                 |
+      |                                   |                                 |
+      | "(5) The browser then           " | "GET /client_metadata"           |
```
samuelgoto, line 234:
```suggestion
      | "(5) The browser then           " | "GET /client_metadata"          |
```

spec/index.bs
```diff
@@ -294,49 +299,48 @@ as if the resource was loaded as a same-origin request, e.g.
...
+the |rp| via the |idp| |account|.
+
+If a user clears browsing data for an |origin| (cookies, localStorage, etc.), the user agent MUST
+[=list/remove=] all triples with an [=/origin=] matching the |origin| from <a>registered accounts set</a>.
 
 <div algorithm>
 To <dfn>compute account state</dfn> given an {{IdentityProviderConfig}} |provider|, an
 {{IdentityProviderAccount}} |account|, and a |globalObject|, run the following steps. This returns
-an [=AccountState=].
+"registered" or "unregistered".
```
samuelgoto, line 321:
```suggestion
<dfn>registered</dfn> or <dfn>unregistered</dfn>.
```

Wouldn't it be good to keep the ability to linkify these, e.g. with {{registered}}?

spec/index.bs
```diff
@@ -294,49 +299,48 @@ as if the resource was loaded as a same-origin request, e.g.
...
-    1. Let |accountState| be [=state machine map=][|triple|].
-    1. Return |accountState|.
+    1. If [=registered accounts set=] [=list/contains=] |triple| , return "registered".
+    1. Otherwise, return "unregistered".
 </div>
 
 <div algorithm>
-To <dfn>sign-in</dfn> the user with a given an [=AccountState=] |accountState|:
-    1. Assert that |accountState|'s {{AccountState/registration state}} is {{registered}}
+To <dfn>register an account</dfn> given an {{IdentityProviderConfig}} |provider|, an
```
samuelgoto, line 333: Something that is a bit confusing about this is that the "registered" vs "unregistered" state is purely at the "browser level", but the user can already be "registered" with the IdP and "unregistered" in the browser. Maybe we should try to use different words so that these two different concepts don't collide?

Maybe "accepted"? "permitted"? "acknowledged"? Language that's closer to "accepting a browser permission" rather than "an account lifecycle"?

spec/index.bs
```diff
@@ -363,8 +367,23 @@ by the [[CM]] API.
...
 <!-- ============================================================ -->
 
-This specification starts by introducing an extension to the
-{{CredentialRequestOptions}} object:
+This section defines the dictionaries passed into the JavaScript call:
+
+<div class=example>
+```js
+const credential = await navigator.credentials.get({
+  identity: { // IdentityCredentialRequestOptions
```
samuelgoto, line 375: Ah, this is really useful indeed! Thanks for putting it together!

spec/index.bs
```diff
@@ -491,60 +519,41 @@ failure.
...
+        needed. Because errors from these fetches are ignored, they can happen in any order.
+    1. Let |permission| be false.
     1. If |accountsList|'s size is 1:
         1. Let |account| be |accountsList|[0].
         1. Let |accountState| be the result of running the [=compute account state=] algorithm
             given |provider|, |account|, and |globalObject|.
-        1. If |accountState|'s {{AccountState/registration state}} is {{unregistered}},
-            let |permission| be the result of running [=request permission to sign-up=] algorithm
-            with |account|, |accountState|, |config|, |provider|, and |globalObject|.
+        1. If |accountState| is "unregistered", set |permission| to the result of running
```
samuelgoto, line 528:
```suggestion
        1. If |accountState| is {{unregistered}}, set |permission| to the result of running
```

Can we keep the <dfn>?

spec/index.bs
```diff
@@ -560,7 +569,6 @@ or failure.
     1. Set |rootUrl|'s [=url/scheme=] to |configUrl|'s [=url/scheme=].
     1. Set |rootUrl|'s [=url/host=] to |configUrl|'s [=url/host=]'s [=host/registrable domain=].
     1. Set |rootUrl|'s [=url/path=] to the <a>list</a> «".well-known", "web-identity"».
-    1. If |rootUrl| is not a [=potentially trustworthy URL=], return failure.
```
samuelgoto, line 563: Why are we removing this?

spec/index.bs
```diff
@@ -785,7 +814,7 @@ To <dfn>fetch an identity assertion</dfn> given an [=AccountState=] |accountStat
 
         :  [=request/url=]
         :: |idTokenUrl|
-        :  [=request/mode=]
+        :  [=request/method=]
```
samuelgoto, line 817: ah, nice catch.

spec/index.bs
```diff
@@ -936,14 +949,70 @@ dictionary IdentityProviderClientMetadata {
 };
 </xmp>
 
+<!-- ============================================================ -->
+## Helper algorithms ## {#helper-algorithms}
```
samuelgoto, line 953:
```suggestion
### Helper algorithms ### {#helper-algorithms}
```

indent++?

samuelgoto (commented):
Thanks for putting this together!

Took an initial pass at this PR.

spec/index.bs
npm1: Because we check earlier `If |configUrl| is not a [=potentially trustworthy URL=], return failure.` so this check cannot possibly fail.

spec/index.bs
npm1: I think Jeffrey suggested in parallel to IdentityCredential, so "##", but I don't mind "###" myself.

npm1:
Addressed feedback, ready for you to take another look.

spec/index.bs
samuelgoto: Either works -- I thought this was just a typo, but if this is deliberate, I'm fine with either.

spec/index.bs
```diff
@@ -294,49 +299,50 @@ as if the resource was loaded as a same-origin request, e.g.
 regardless of the
 [SameSite](https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#name-the-samesite-attribute-2)
 value (which is used when a resource loaded as a third-party, not first-party). This makes it easy
-for an [=IDP=] to adopt the FedCM API without introducing security issues on the API, since the
+for an [=IDP=] to adopt the FedCM API. It doesn't introduce security issues on the API because the
 [=RP=] cannot inspect the results from the fetches in any way.
 
 <!-- ============================================================ -->
-## The State Machine ## {#browser-api-state-machine}
+## The registered accounts set ## {#browser-registered-accounts-set}
```
samuelgoto, line 306:
```suggestion
## The permission storage ## {#browser-permission-storage}
```
Maybe we should call this a permission store instead if we use "permitted" and "non-permitted" for values? WDYT?

It matches more the language that we use in implementation at least, and it feels like it would have a different implication than "registered accounts".

WDYT?

samuelgoto: Maybe we could add language that connects to the privacy threat model, e.g. "once a permission has been granted for the user to login in to the Rp with the IdP, the browser allows features like x, y, z".

Also, instead of "register an account" we could call this "grant permission" and instead of "compute account state", something like "check permission".

spec/index.bs
```diff
@@ -491,60 +521,41 @@ failure.
...
+        needed. Because errors from these fetches are ignored, they can happen in any order.
+    1. Let |permission| be false.
     1. If |accountsList|'s size is 1:
         1. Let |account| be |accountsList|[0].
         1. Let |accountState| be the result of running the [=compute account state=] algorithm
             given |provider|, |account|, and |globalObject|.
-        1. If |accountState|'s {{AccountState/registration state}} is {{unregistered}},
-            let |permission| be the result of running [=request permission to sign-up=] algorithm
-            with |account|, |accountState|, |config|, |provider|, and |globalObject|.
+        1. If |accountState| is <a for="compute account state">not permitted</a>, set |permission|
```
samuelgoto, line 530: {{not permitted}} doesn't work anymore? didn't you <dfn> earlier?

spec/index.bs
samuelgoto: Got it.

spec/index.bs
npm1: Actually I no longer like "permission" because it is already a thing that's not the same as what we have, ie https://www.w3.org/TR/permissions. It's different for instance in that you should not be able to query our 'permissions'.  I actually now think registered vs unregistered is still best compared with your other suggestions. Do you have any other ideas for naming?

spec/index.bs
npm1: Does not seem to work, maybe because of the scoped dfn (as suggested by Jeffrey)

npm1:
Updated the naming to 'connected', if that works for you?

spec/index.bs
```diff
@@ -760,32 +782,41 @@ To <dfn>fetch the account picture</dfn> given an {{IdentityProviderAccount}} |ac
...
+provided to the [=RP=].
+
 <div algorithm>
-To <dfn>fetch an identity assertion</dfn> given an [=AccountState=] |accountState|, a {{USVString}}
+To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
     |accountId|, an {{IdentityProviderConfig}} |provider|, an {{IdentityProviderAPIConfig}}
     |config|, and |globalObject|, run the following steps. This returns an {{IdentityCredential}}
     or failure.
-    1. Assert |accountState|'s {{AccountState/registration state}} is {{registered}}.
     1. Let |idTokenUrl| be the result of [=computing the manifest URL=] given |provider|,
```
asr-enid, line 809: Given we agreed that the token passed back from the id assertion endpoint is an opaque string it does not seem to make sense to use idTokenUrl here

spec/index.bs
```diff
@@ -294,49 +299,50 @@ as if the resource was loaded as a same-origin request, e.g.
...
-    1. Return |accountState|.
+    1. If [=connected accounts set=] [=list/contains=] |triple| , return
+        [=compute account state/connected=].
+    1. Otherwise, return [=compute account state/disconnected=].
 </div>
 
 <div algorithm>
-To <dfn>sign-in</dfn> the user with a given an [=AccountState=] |accountState|:
-    1. Assert that |accountState|'s {{AccountState/registration state}} is {{registered}}
+To <dfn>register an account</dfn> given an {{IdentityProviderConfig}} |provider|, an
```
samuelgoto, line 335:
```suggestion
To <dfn>create a connection between the RP and the IdP account</dfn> given an {{IdentityProviderConfig}} |provider|, an
```

spec/index.bs
```diff
@@ -294,49 +299,50 @@ as if the resource was loaded as a same-origin request, e.g.
...
-</dl>
+an account identifier. It represents the set of triples such that the user has used FedCM to login to
+the |rp| via the |idp| |account|.
 
-<div algorithm>
+If a user clears browsing data for an |origin| (cookies, localStorage, etc.), the user agent MUST
+[=list/remove=] all triples with an [=/origin=] matching the |origin| from <a>connected accounts set</a>.
+
+<div algorithm="compute account state">
 To <dfn>compute account state</dfn> given an {{IdentityProviderConfig}} |provider|, an
```
samuelgoto, line 319:
```suggestion
To <dfn>compute the connection status</dfn> given an {{IdentityProviderConfig}} |provider|, an
```

spec/index.bs
```diff
@@ -294,49 +299,50 @@ as if the resource was loaded as a same-origin request, e.g.
...
 {{IdentityProviderAccount}} |account|, and a |globalObject|, run the following steps. This returns
```
samuelgoto, line 320:
```suggestion
{{IdentityProviderAccount}} |account|, and a [=RP=]'s |globalObject|, run the following steps. This returns
```

samuelgoto (approved):
LGTM

Some small suggestions to try to use more the new language here.

I tried my best to find if there were any normative changes here and didn't find anything obvious. I believe the intent here is to make this fully editorial - without any normative changes -, and if so, feel free to merge this when you resolve this last round of comments.

Thanks for working with Jeffrey and putting this together!

spec/index.bs
npm1: I'll write |globalObject| (the [=RP=]'s) instead as otherwise this breaks from the usual way of passing parameters with either just variable name or type and name.

PR #464, opened 2023-04-28
Add support to show only top-level origin when calling from an iframe

npm1:
This PR fixes the client metadata fetch by passing the client ID in the query, and fixes https://github.com/fedidcg/FedCM/issues/449 by adding the proposed logic.

***
<a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/464.html" title="Last updated on May 3, 2023, 8:15 PM UTC (9382f86)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/464/c2277e6...9382f86.html" title="Last updated on May 3, 2023, 8:15 PM UTC (9382f86)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -22,6 +22,10 @@ Test Suite: https://github.com/web-platform-tests/wpt/blob/master/credential-man
 </pre>
 
 <pre class=anchors>
+spec: html; urlPrefix: https://html.spec.whatwg.org/multipage
+    type: dfn; for: window
+        text: navigable; urlPrefix: nav-history-apis.html#window-navigable
+
 spec: ecma262; urlPrefix: https://tc39.github.io/ecma262/
     type: dfn
         text: internal method; url: sec-ordinary-object-internal-methods-and-internal-slots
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -22,6 +22,10 @@ Test Suite: https://github.com/web-platform-tests/wpt/blob/master/credential-man
 </pre>
 
 <pre class=anchors>
+spec: html; urlPrefix: https://html.spec.whatwg.org/multipage
```
npm1, line 25: This is now exported from HTML (filed a bug about it), but hasnt reached bikeshed yet.

npm1 (commented):
Adding Yi for the algorithm logic sanity check, Jeffrey for help in making sure the navigable stuff is being handled properly

spec/index.bs
```diff
@@ -1231,6 +1254,8 @@ The {{IdentityProviderClientMetadata}} object's members have the following seman
     ::  A link to the [=RP=]'s Privacy Policy.
     :   <dfn>terms_of_service_url</dfn>
     ::  A link to the [=RP=]'s Terms of Service.
+    :   <dfn>client_matches_top_frame_origin</dfn>
+    ::  Whether the FedCM iframe caller matches the top-level embedder's origin.
```
yi-gu, line 1258: s/caller/client

spec/index.bs
```diff
@@ -880,6 +893,11 @@ To <dfn>request permission to sign-up</dfn> the user with a given an {{IdentityP
         is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
         |account|["{{IdentityProviderAccount/approved_clients}}"], then display the
         |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
+    1. If |topFrameOrigin| is not null, |metadata| is not failure, and
+        |metadata|["{{IdentityProviderClientMetadata/client_matches_top_frame_origin}}"] is defined
+        and set to true, the user agent MAY choose to only show the |topFrameOrigin| in its UI.
```
yi-gu, line 898: Once UA acquires such information, they may choose to show EITHER top frame OR iframe because they match each other. We may not need to suggest using "top frame" for UAs. WDYT?

yi-gu, line 898: Similarly, whether to show "topFrameOrigin" or "topFrameDomain" is UA's choice.

spec/index.bs
```diff
@@ -880,6 +893,11 @@ To <dfn>request permission to sign-up</dfn> the user with a given an {{IdentityP
...
+    1. Otherwise, if |topOrigin| is not null, the user agent SHOULD show both |globalObject|'s
```
yi-gu, line 899: Does this algorithm cover the case where top frame origin == iframe origin? Or the API is called from the top frame so there's no iframe at all? In those cases the UA should probably just show ONE frame instead of TWO on the UI regardless of whether there's a `client_matches_top_frame_origin` returned from the client metadata fetch response.

spec/index.bs
npm1: Sure, either would work I think.

npm1:
Addressed comments, let me know what you think

yi-gu:
Thanks! Algorithm LGTM.

npm1:
@martinthomson @cboozar @bvandersloot-mozilla please take a look. Per your request, not filing an official Mozilla standards position but you can consider this to be it. As discussed in the call, we would like to ship this on Chrome 115, which branches in May 23 so we'll send I2S perhaps a week or so before then.

cboozar:
Makes sense to me, though I want to understand why the urgency for 115, who is consuming this ATM?

npm1:
I am closing this as we're not moving forward with this proposal.

PR #465, opened 2023-04-28
Add WebDriver commands

cbiesinger:


spec/index.bs
```diff
@@ -1358,6 +1370,229 @@ occur when there is access to the [=current settings object=]. The [=internal me
...
 by {{CredentialsContainer}}'s <a abstract-op>Request a `Credential`</a> abstract operation.
 
+<!-- ============================================================ -->
+# User Agent Automation # {#automation}
+<!-- ============================================================ -->
+
+For the purposes of user agent automation and website testing, this document
+defines the below [[WebDriver2]] [=extension commands=].
+
+## <dfn export>WebDriver extension commands</dfn> ## {#webdriver-commands}
```
npm1, line 1380: Do you plan to add another section other than this? Otherwise seems unneeded

spec/index.bs
```diff
@@ -1358,6 +1370,229 @@ occur when there is access to the [=current settings object=]. The [=internal me
...
+
+These WebDriver [=extension commands=] allow automated interactions with any
```
npm1, line 1382: Remove this paragraph? Seems to say the same as previous

spec/index.bs
```diff
@@ -1358,6 +1370,229 @@ occur when there is access to the [=current settings object=]. The [=internal me
...
+active FedCM dialogs.
+
```
npm1, line 1384: Perhaps we add a subsection for each remote end step. Right now it's unclear what you're defining in each case unless you look at the POST closely. It could be in a title.

spec/index.bs
```diff
@@ -1358,6 +1370,229 @@ occur when there is access to the [=current settings object=]. The [=internal me
...
+                <td>POST</td>
+                <td>`/session/{session id}/fedcm/canceldialog`</td>
+            </tr>
+        </tbody>
+    </table>
+</figure>
+
+The [=remote end steps=] are:
+
+1. If |parameters| is not a JSON [[ECMASCRIPT#sec-json-object|Object]], return a
```
npm1, line 1404: Why do you care about the object type if its never used? Even where it is used, you invoke [getting a property](https://w3c.github.io/webdriver/#dfn-getting-properties) and I'm not sure this is needed?

npm1:
That was an initial review, it mostly looks good. I'll take a closer look after the comments are addressed!

spec/index.bs
cbiesinger: Removed.

spec/index.bs

spec/index.bs
cbiesinger: Removed here, thanks. Can you clarify why getting a property is not needed / what I should do instead?

cbiesinger (commented):
PTAL!

cbiesinger:
@martinthomson @cboozar @bvandersloot-mozilla -- could you comment whether you are fine with this proposal?

cbiesinger:
(I know Ben said as much in https://github.com/fedidcg/FedCM/pull/395#issuecomment-1384466751 but I'd like to reconfirm now that this is a spec PR)

spec/index.bs
```diff
@@ -1358,6 +1370,231 @@ occur when there is access to the [=current settings object=]. The [=internal me
...
+
+1. Let |accounts| be the list of accounts that the user can or could choose
+    from in the current flow.
+
+1. Let |list| be an empty list.
+
+1. For each |account| in |accounts|:
+    1. Let |accountState| be the result of running the [=compute account state=]
+        algorithm given |account|
+    1. Append an object to |list| with the following properties:
```
npm1, line 1483: [=list/Append=]. Also what kind of object is it?

spec/index.bs
```diff
@@ -1358,6 +1370,231 @@ occur when there is access to the [=current settings object=]. The [=internal me
...
+        1. `name` set to the account's {{IdentityProviderAccount/name}}
+        1. `givenName` set to the account's {{IdentityProviderAccount/given_name}},
+            if present
+        1. `pictureUrl` set to the account's {{IdentityProviderAccount/picture}},
+            if present.
+        1. `idpConfigUrl` set to the {{IdentityProviderConfig/configURL}} of the
+            IDP this account belongs to
+        1. `loginState` to `"SignUp"` if |accountState|'s {{AccountState/registration
+            state}} is {{unregistered}} and `"SignIn"` otherwise
+        1. `termsOfServiceUrl` to the {{IdentityProviderClientMetadata/terms_of_service_url}}
```
npm1, line 1495: The way it is phrased, I'm unsure what happens if the loginState is SignIn? This property is not set I guess? If it is SignUp and there is no TOS, is this also not set?

spec/index.bs
```diff
@@ -1358,6 +1370,231 @@ occur when there is access to the [=current settings object=]. The [=internal me
...
+            <tr>
+                <td>GET</td>
+                <td>`/session/{session id}/fedcm/gettitle`</td>
+            </tr>
+        </tbody>
+    </table>
+</figure>
+
+Note: This command lets automation verify that the
+    [https://github.com/fedidcg/FedCM/blob/main/proposals/context-api.md](context API)
```
npm1, line 1522: Linking not working (in preview at least), use `<a href>` perhaps.

spec/index.bs
```diff
@@ -1358,6 +1370,229 @@ occur when there is access to the [=current settings object=]. The [=internal me
...
+    [=WebDriver error=] with [=WebDriver error code=] [=invalid argument=].
+
+1. If no FedCM dialog is currently open, return a [=WebDriver error=] with
+    [=WebDriver error code=] [=no such alert=].
+
+1. Let |accountIndex| be the result of [=getting a property=] named `"accountIndex"`
+    from |parameters|.
+
+1. If |accountIndex| is [=undefined=] or is not in the range from 0 to the
+    number of accounts that the user can choose from in the current flow,
```
mknowles36, line 1441: I'm assuming the range is inclusive at the beginning and exclusive at the end (i.e., zero up to the number of accounts minus one)?  If so, perhaps make that explicit by replacing "...in the range from 0 to..." with "...in the range [0..num-accounts), from 0 to..." or something similar?

spec/index.bs
```diff
@@ -29,6 +29,17 @@ spec: ecma262; urlPrefix: https://tc39.github.io/ecma262/
 spec: credential-management-1; urlPrefix: https://w3c.github.io/webappsec-credential-management/
     type: dfn
         text: same-origin with its ancestors; url: same-origin-with-its-ancestors
+spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
```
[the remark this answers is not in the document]
cbiesinger: Filed https://github.com/w3c/webdriver/issues/1736 and in the process realized that several of these are already exported, so I removed those from here.

spec/index.bs
```diff
@@ -454,7 +465,8 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
 
           Issue: Support choosing accounts from multiple [=IDP=]s, as described [here](https://github.com/fedidcg/FedCM/issues/319).
     1. Run {{WindowOrWorkerGlobalScope/setTimeout()}} passing a [=task=] which throws a
-        {{NetworkError}}, after a timeout of 60 seconds.
+        {{NetworkError}}, after a timeout of 60 seconds, unless this <dfn>promise rejection
```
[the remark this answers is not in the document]
cbiesinger: That's an unfortunate aspect of https://github.com/fedidcg/FedCM/pull/453 not having landed yet. Once either PR is pushed I'll update the other one.

spec/index.bs
cbiesinger: Done

spec/index.bs
cbiesinger: Done

spec/index.bs
cbiesinger: Done

spec/index.bs
cbiesinger: Correct. Tried to clarify.

spec/index.bs
cbiesinger: Done

spec/index.bs
cbiesinger: Hmm places like https://w3c.github.io/webdriver/#dfn-validate-capabilities have the same wording as I have?

spec/index.bs
cboozar: I still don't understand how the behavior is limited in terms of interaction to make sure it cannot be exploited. Or is that a question for later?

cbiesinger: I'm not sure I understand the concern. Webdriver APIs are not exposed to the web (and also not enabled by default). See, for example, https://chromedriver.chromium.org/security-considerations

There are lots of ways webdriver can be exploited if that weren't the case since the APIs are very powerful.

This is only meant to be used by local automation/testing tools

cbiesinger:
I have added a getdialogtype command to this PR, to support the recently added re-authentication support.

spec/index.bs
```diff
@@ -1456,6 +1464,260 @@ occur when there is access to the [=current settings object=]. The [=internal me
...
+    1. Let |accountState| be the result of running the [=compute account state=]
+        algorithm given |account|
+    1. [=list/Append=] a [=dictionary=] to |list| with the following properties:
+        1. `accountId` set to the account's {{IdentityProviderAccount/id}}
+        1. `email` set to the account's {{IdentityProviderAccount/email}}
+        1. `name` set to the account's {{IdentityProviderAccount/name}}
+        1. `givenName` set to the account's {{IdentityProviderAccount/given_name}},
+            if present
+        1. `pictureUrl` set to the account's {{IdentityProviderAccount/picture}},
+            if present.
```
npm1, line 1584: nit: be consistent about whether items in a list include . at the end or not.

spec/index.bs
```diff
@@ -492,10 +498,12 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
         |options|["{{CredentialRequestOptions/mediation}}"], and |globalObject|.
     1. If |credential| is a pair:
         1. Let |throwImmediately| be the value of the second element of the pair.
-        1. If |throwImmediately| is false, the user agent SHOULD wait a random
-            amount of time before the next step, unless it has implemented
-            another way to prevent exposing to the RP whether the user has
-            an account logged in to the RP.
+        1. If |throwImmediately| is false, and this <dfn>promise rejection
```
npm1, line 501: This is a long sentence, perhaps we can split it into conditions. Something like...

The user agent SHOULD wait... if the following conditions hold:
* throwImmediately is false
* user agent automation
* no other way to prevent exposing the RP

npm1 (approved):

spec/index.bs

spec/index.bs

samuelgoto:
Thanks for putting this together and thanks @bvandersloot-mozilla for the review!

Merged!

PR #468, opened 2023-05-05
More editorial changes

npm1:


spec/index.bs
```diff
@@ -358,9 +358,18 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
 </pre>
 
 <dl>
+    :   <b>{{Credential/id}}</b>
+    ::  The {{Credential/id}}'s attribute getter returns the empty string.
+    :   <b>{{Credential/type}}</b>
+    ::  The {{Credential/type}}'s attribute getter is defined by the {{Credential}} interface.
```
cbiesinger, line 364: It feels weird to describe this here. Surely, it is expected that Credential::type is defined by Credential's spec.

spec/index.bs
```diff
@@ -358,9 +358,18 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
 </pre>
 
 <dl>
+    :   <b>{{Credential/id}}</b>
+    ::  The {{Credential/id}}'s attribute getter returns the empty string.
```
cbiesinger, line 362: this feels a little weird for the same reason as below, but since Credential leaves this up to each specific type this is probably the expected way to specify this

spec/index.bs
```diff
@@ -728,9 +742,13 @@ To <dfn>fetch the accounts list</dfn> given an {{IdentityProviderAPIConfig}} |co
         :  [=request/mode=]
         :: "no-cors"
 
-        Issue: The credentialed fetch in this algorithm can lead to a timing attack that leaks the
+        <p class=issue>: The credentialed fetch in this algorithm can lead to a timing attack that leaks the
```
cbiesinger, line 745: Why did you make this change? I thought bikeshed should do this automatically.

spec/index.bs
npm1: Yea ok I think we can remove this (I asked Jeffrey and will do whatever he suggests though)

spec/index.bs
npm1: Yep this seems like a 'pure virtual' of sorts.

spec/index.bs
npm1: Oh yea you're right, I'll change these to Issue since it's nicer.

npm1:
Ready for further review, thanks!

cbiesinger (approved):

PR #470, opened 2023-05-12
Add loginHint, RP context, and getUserInfo

npm1:
Fixes https://github.com/fedidcg/FedCM/issues/426, https://github.com/fedidcg/FedCM/issues/456, and https://github.com/fedidcg/FedCM/issues/382

***
<a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/470.html" title="Last updated on Jun 8, 2023, 7:23 PM UTC (e0024bf)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/470/a2b9636...e0024bf.html" title="Last updated on Jun 8, 2023, 7:23 PM UTC (e0024bf)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1139,6 +1139,9 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
   };
 </pre>
 
+Issue: [Decide](https://github.com/fedidcg/FedCM/issues/476) whether {{IdentityProvider}} is the
+correct location for the {{IdentityProvider/getUserInfo()}} method.
+
 An {{IdentityUserInfo}} represents user account information from a user. This information is exposed
 to the [=IDP=] once the user has already used the FedCM API to login in the [=RP=]. That is, it is
 exposed when there exists an account |account| such that the [=connected accounts set=] [=list/contains=]
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -453,7 +453,8 @@ dictionary IdentityProviderConfig {
     :   <b>{{IdentityProviderConfig/loginHint}}</b>
     ::  A string representing the login hint corresponding to an account which the RP wants the user
         agent to show to the user. If provided, the user agent will not show accounts which do not
-        match this login hint value.
+        match this login hint value. It generally matches some attribute from the desired
+        {{IdentityProviderAccount}}.
 </dl>
 
 <!-- ============================================================ -->
@@ -1129,7 +1130,7 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
   [Exposed=Window, SecureContext] dictionary IdentityUserInfo {
     USVString email;
     USVString name;
-    USVString given_name;
+    USVString givenName;
     USVString picture;
   };
 
@@ -1139,10 +1140,31 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
 </pre>
 
 An {{IdentityUserInfo}} represents user account information from a user. This information is exposed
-to the [=IDP=] once the user has already used the FedCM API to login in the [=RP=]. The information
-matches what is received from the <a>accounts list endpoint</a>. The [=IDP=] can obtain this
-information by invoking the {{IdentityProvider/getUserInfo()}} static method from an iframe matching
-the [=/origin=] of its {{IdentityProviderConfig/configURL}}.
+to the [=IDP=] once the user has already used the FedCM API to login in the [=RP=]. That is, it is
+exposed when there exists an account |account| such that the the [=connected accounts set=] [=list/contains=]
+the triple ([=RP=], [=IDP=], |account|). The information matches what is received from the
+<a>accounts list endpoint</a>. The [=IDP=] can obtain this information by invoking the
+{{IdentityProvider/getUserInfo()}} static method from an iframe matching the [=/origin=] of its
+{{IdentityProviderConfig/configURL}}.
+
+<div class="example">
+```js
+const user_info = await IdentityProvider.getUserInfo({
+    configUrl: "https://idp.example/fedcm.json",
+    clientId: "client1234"
+});
+
+if (user_info.length > 0) {
+  // It's up to the IDP regarding how to display the returned accounts.
+  // Accounts are sorted based on RP registration status.
+  const name = user_info[0].name;
+  const given_name = user_info[0].given_name;
+  const display_name = given_name ? given_name : name;
+  const picture = user_info[0].picture;
+  const email = user_info[0].email;
+}
+```
+</div>
 
 <div algorithm="getUserInfo">
 When invoking the {{IdentityProvider/getUserInfo()}} method given an {{IdentityProviderConfig}}
@@ -1152,6 +1174,11 @@ When invoking the {{IdentityProvider/getUserInfo()}} method given an {{IdentityP
     1. Let |document| be |globalObject|'s [=associated Document=].
     1. If |document| is not [=allowed to use=] the [=identity-credentials-get=]
         [=policy-controlled feature=], throw a "{{NotAllowedError}}" {{DOMException}}.
+    1. If there does not exist an account |account| such that [=compute the connection status=] of
+        |provider|, |account|, and |globalObject| returns
+        [=compute the connection status/connected=], then throw a new "{{NetworkError}}"
+        {{DOMException}}. This check can be performed by iterating over the
+        [=connected accounts set=] or by keeping a separate data structure to make this lookup fast.
     1. Let |configUrl| be the result of running [=parse url=] with |provider|'s
         {{IdentityProviderConfig/configURL}} and |globalObject|.
     1. If |configUrl| is failure, throw an "{{InvalidStateError}}" {{DOMException}}.
@@ -1194,7 +1221,7 @@ When invoking the {{IdentityProvider/getUserInfo()}} method given an {{IdentityP
                 :: |account|["{{IdentityProviderAccount/email}}"]   
                 :  {{IdentityUserInfo/name}}
                 :: |account|["{{IdentityProviderAccount/name}}"]   
-                :  {{IdentityUserInfo/given_name}}
+                :  {{IdentityUserInfo/givenName}}
                 :: |account|["{{IdentityProviderAccount/given_name}}"]   
                 :  {{IdentityUserInfo/picture}}
                 :: |account|["{{IdentityProviderAccount/picture}}"]
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -567,7 +567,7 @@ the exception thrown.
         and |globalObject|.
     1. If |provider|'s {{IdentityProviderConfig/loginHint}} is not empty:
         1. For every |account| in |accountList|, remove |account| from |accountList| if |account|'s
-            {{IdentityProviderAccount/hints}} does not [=list/contain=] |provider|'s
+            {{IdentityProviderAccount/login_hints}} does not [=list/contain=] |provider|'s
             {{IdentityProviderConfig/loginHint}}.
     1. If |accountsList| is failure, return (failure, false).
     1. For each |acc| in |accountsList|:
@@ -838,7 +838,7 @@ dictionary IdentityProviderAccount {
   USVString given_name;
   USVString picture;
   sequence<USVString> approved_clients;
-  sequence<DOMString> hints;
+  sequence<DOMString> login_hints;
 };
 dictionary IdentityProviderAccountList {
   sequence<IdentityProviderAccount> accounts;
@@ -1127,7 +1127,7 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
 {{IdentityProvider}} interface:
 
 <pre class="idl">
-  [Exposed=Window, SecureContext] dictionary IdentityUserInfo {
+  dictionary IdentityUserInfo {
     USVString email;
     USVString name;
     USVString givenName;
@@ -1141,7 +1141,7 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
 
 An {{IdentityUserInfo}} represents user account information from a user. This information is exposed
 to the [=IDP=] once the user has already used the FedCM API to login in the [=RP=]. That is, it is
-exposed when there exists an account |account| such that the the [=connected accounts set=] [=list/contains=]
+exposed when there exists an account |account| such that the [=connected accounts set=] [=list/contains=]
 the triple ([=RP=], [=IDP=], |account|). The information matches what is received from the
 <a>accounts list endpoint</a>. The [=IDP=] can obtain this information by invoking the
 {{IdentityProvider/getUserInfo()}} static method from an iframe matching the [=/origin=] of its
@@ -1149,19 +1149,18 @@ the triple ([=RP=], [=IDP=], |account|). The information matches what is receive
 
 <div class="example">
 ```js
-const user_info = await IdentityProvider.getUserInfo({
+const userInfo = await IdentityProvider.getUserInfo({
     configUrl: "https://idp.example/fedcm.json",
     clientId: "client1234"
 });
 
-if (user_info.length > 0) {
+if (userInfo.length > 0) {
   // It's up to the IDP regarding how to display the returned accounts.
-  // Accounts are sorted based on RP registration status.
-  const name = user_info[0].name;
-  const given_name = user_info[0].given_name;
-  const display_name = given_name ? given_name : name;
-  const picture = user_info[0].picture;
-  const email = user_info[0].email;
+  const name = userInfo[0].name;
+  const givenName = userInfo[0].givenName;
+  const displayName = givenName ? givenName : name;
+  const picture = userInfo[0].picture;
+  const email = userInfo[0].email;
 }
 ```
 </div>
@@ -1186,8 +1185,6 @@ When invoking the {{IdentityProvider/getUserInfo()}} method given an {{IdentityP
         throw an "{{InvalidStateError}}" {{DOMException}}.
     1. Run a [[!CSP]] check with a [[CSP#directive-connect-src|connect-src]] directive on the URL
         passed as |configUrl|. If it fails, throw a new "{{NetworkError}}" {{DOMException}}.
-    1. If there exists a pending {{IdentityProvider/getUserInfo()}} call for this |document|, throw
-        a new "{{AbortError}}" {{DOMException}}.
     1. If |globalObject|'s [=Window/navigable=] is a [=/top-level traversable=], throw a new
         "{{NetworkError}}" {{DOMException}}.
     1. If the user has disabled the FedCM API on the |globalObject|'s [=Window/navigable=]'s
@@ -1206,7 +1203,8 @@ When invoking the {{IdentityProvider/getUserInfo()}} method given an {{IdentityP
                 [=list/contain=] |provider|'s {{IdentityProviderConfig/clientId}}, continue.
             
                 Note: this allows the [=IDP=] to override whether an account is a returning account.
-                This could be useful for instance in cases where the user revokes the account.
+                This could be useful for instance in cases where the user has revoked the account
+                out of band.
             
             1. [=Compute the connection status=] of |provider|, |account|, and |globalObject|. If the
                 result is [=compute the connection status/connected=], set |hasReturningAccount| to
@@ -1443,7 +1441,7 @@ Every {{IdentityProviderAccount}} is expected to have members with the following
     ::  A list of [=RP=]s (that gets matched against the requesting {{IdentityProviderConfig/clientId}}) this account is already registered with.
         Used in the [=request permission to sign-up=] to allow the [=IDP=] to control whether to show
         the Privacy Policy and the Terms of Service.
-    :   <dfn>hints</dfn>
+    :   <dfn>login_hints</dfn>
     ::  A list of strings which correspond to all of the login hints which match with this account.
         An [=RP=] can use the {{IdentityProviderConfig/loginHint}} to request that only an account
         matching a given value is shown to the user.
@@ -1461,7 +1459,7 @@ For example:
    "email": "john_doe@idp.example",
    "picture": "https://idp.example/profile/123",
    "approved_clients": ["123", "456", "789"],
-   "hints": ["john_doe"]
+   "login_hints": ["john_doe"]
   }, {
    "id": "5678",
    "given_name": "Johnny",
@@ -1469,7 +1467,7 @@ For example:
    "email": "johnny@idp.example",
    "picture": "https://idp.example/profile/456",
    "approved_clients": ["abc", "def", "ghi"],
-   "hints": ["email=johhny@idp.example", "id=5678"]
+   "login_hints": ["email=johhny@idp.example", "id=5678"]
   }]
 }
 ```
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -433,6 +444,10 @@ dictionary IdentityProviderConfig {
     ::  A random number of the choice of the [=RP=]. It is generally used to associate a client
         session with a {{IdentityProviderToken/token}} and to mitigate replay attacks. Therefore, this value should have
         sufficient entropy such that it would be hard to guess.
+    :   <b>{{IdentityProviderConfig/loginHint}}</b>
+    ::  A string representing the login hint corresponding to an account which the RP wants the user
```
yi-gu, line 448: Should we call out some examples for the "string"? e.g. it should be one of the fields in the IdentityProviderAccount

spec/index.bs
```diff
@@ -525,20 +540,25 @@ agent UI, and creates the {{IdentityCredential}} that is then returned to the [=
...
     1. If |requiresUserMediation| is true and |mediation| is
         "{{CredentialMediationRequirement/silent}}", return failure.
     1. Let |config| be the result of running [=fetch the config file=] with |provider| and
         |globalObject|.
     1. If |config| is failure, return (failure, false).
     1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
         and |globalObject|.
+    1. If |provider|'s {{IdentityProviderConfig/loginHint}} is not empty:
+        1. For every |account| in |accountList|, remove |account| from |accountList| if |account|'s
+            {{IdentityProviderAccount/hints}} does not [=list/contain=] |provider|'s
```
samuelgoto, line 560: Not really worth the trouble, but it is a bit awkward to me that the JS API is called `loginHint` and the JSON is `hints`. This seems right to me for reasons that I can't quite put my finger on, but just wanted to raise in case other people also feel that they are not parallel to each other.

spec/index.bs
```diff
@@ -805,6 +826,7 @@ dictionary IdentityProviderAccount {
   USVString given_name;
   USVString picture;
   sequence<USVString> approved_clients;
+  sequence<DOMString> hints;
```
samuelgoto, line 829: I think it would be awkward to call this`loginHints`, but just wanted to raise that it is also awkward that this is called `hints` and the JS API is called `loginHint`.

Maybe you are envisioning that `hints`are going to be used for things beyond `login`?

spec/index.bs
```diff
@@ -1082,6 +1105,89 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
     1. Return |json|.
 </div>
 
+<!-- ============================================================ -->
+## The IdentityProvider Interface ## {#browser-api-identity-provider-interface}
+<!-- ============================================================ -->
+
+This specification introduces the {{IdentityUserInfo}} dictionary as well as the
```
samuelgoto, line 1112:
```suggestion
This section introduces the {{IdentityUserInfo}} dictionary as well as the
```

spec/index.bs
```diff
@@ -1082,6 +1105,89 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+{{IdentityProvider}} interface:
+
+<pre class="idl">
+  [Exposed=Window, SecureContext] dictionary IdentityUserInfo {
+    USVString email;
+    USVString name;
+    USVString given_name;
```
samuelgoto, line 1119: shouldn't this take the same style guide as JS objects? that is, shouldn't this be givenName instead?

```suggestion
    USVString givenName;
```

spec/index.bs
```diff
@@ -1082,6 +1105,89 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+    USVString picture;
+  };
+
+  [Exposed=Window, SecureContext] interface IdentityProvider {
+      static Promise&lt;sequence&lt;IdentityUserInfo&gt;?&gt; getUserInfo(IdentityProviderConfig config);
+  };
+</pre>
+
+An {{IdentityUserInfo}} represents user account information from a user. This information is exposed
+to the [=IDP=] once the user has already used the FedCM API to login in the [=RP=]. The information
```
samuelgoto, line 1129:
```suggestion
to the [=IDP=] once the user has already used the FedCM API to login in the [=RP=]. That is, when [=compute the connection status=] is `connected`. The information
```

spec/index.bs
```diff
@@ -1082,6 +1105,89 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+matches what is received from the <a>accounts list endpoint</a>. The [=IDP=] can obtain this
```
samuelgoto, line 1130:
```suggestion
gathered from the <a>accounts list endpoint</a>. The [=IDP=] can obtain this
```

spec/index.bs
```diff
@@ -1082,6 +1105,89 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+information by invoking the {{IdentityProvider/getUserInfo()}} static method from an iframe matching
+the [=/origin=] of its {{IdentityProviderConfig/configURL}}.
+
+<div algorithm="getUserInfo">
+When invoking the {{IdentityProvider/getUserInfo()}} method given an {{IdentityProviderConfig}}
+|provider|, perform the following steps:
+
+    1. Let |globalObject| be the [=current global object=].
```
samuelgoto, line 1138: don't you need to check for the Rp/IdP connection status somewhere in this algorithm here?

spec/index.bs
```diff
@@ -1082,6 +1105,89 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+    USVString given_name;
+    USVString picture;
+  };
+
+  [Exposed=Window, SecureContext] interface IdentityProvider {
+      static Promise&lt;sequence&lt;IdentityUserInfo&gt;?&gt; getUserInfo(IdentityProviderConfig config);
+  };
+</pre>
+
+An {{IdentityUserInfo}} represents user account information from a user. This information is exposed
```
samuelgoto, line 1128: Maybe add an non normative example here to give readers a sense of how this is going to be used?

spec/index.bs
npm1: It's not required so not should, but I added a sentence.

spec/index.bs
npm1: Yea, I don't have a strong opinion, it could be `login_hints`

spec/index.bs
npm1: It would be `login_hints` but other than that, I don't recall why this ended up as `hints` instead of `login_hints`.

spec/index.bs
npm1: There are already two mentions of "This specification introduces". It is not really wrong, so I'll keep it.

spec/index.bs
npm1: Ah yes you're right, since this is exposed to JS.

spec/index.bs
npm1: That's handwavy since the connection status depends on three things. It's also not a complete sentence. Adding my own phrasing of this

spec/index.bs
npm1: I don't think your suggestion is quite right..

spec/index.bs
npm1: It is... used further down. Though I think we should bail out earlier if there is no connection status: https://bugs.chromium.org/p/chromium/issues/detail?id=1445263. I updated the implementation to where I think the check ought to be.

npm1:
PTAL

spec/index.bs
yi-gu: Any reason why we want to use `hints`?

spec/index.bs
```diff
@@ -1082,6 +1106,115 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+to the [=IDP=] once the user has already used the FedCM API to login in the [=RP=]. That is, it is
+exposed when there exists an account |account| such that the the [=connected accounts set=] [=list/contains=]
```
yi-gu, line 1131: nit: remove redundent "the"

spec/index.bs
```diff
@@ -1082,6 +1106,115 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+the triple ([=RP=], [=IDP=], |account|). The information matches what is received from the
+<a>accounts list endpoint</a>. The [=IDP=] can obtain this information by invoking the
+{{IdentityProvider/getUserInfo()}} static method from an iframe matching the [=/origin=] of its
+{{IdentityProviderConfig/configURL}}.
+
+<div class="example">
```
yi-gu, line 1137: Should we add a comment to call out the origin of the html?

spec/index.bs
```diff
@@ -1082,6 +1106,115 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+const user_info = await IdentityProvider.getUserInfo({
+    configUrl: "https://idp.example/fedcm.json",
+    clientId: "client1234"
+});
+
+if (user_info.length > 0) {
+  // It's up to the IDP regarding how to display the returned accounts.
+  // Accounts are sorted based on RP registration status.
+  const name = user_info[0].name;
+  const given_name = user_info[0].given_name;
```
yi-gu, line 1148: s/given_name/givenName

spec/index.bs
```diff
@@ -1082,6 +1106,115 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+            {{DOMException}}.
+        1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
+            and |globalObject|.
+        1. Let |hasReturningAccount| be false.
+        1. For each |account| in |accountsList|:
+            1. If |account|["{{IdentityProviderAccount/approved_clients}}"] is not empty and it does not
+                [=list/contain=] |provider|'s {{IdentityProviderConfig/clientId}}, continue.
+            
+                Note: this allows the [=IDP=] to override whether an account is a returning account.
+                This could be useful for instance in cases where the user revokes the account.
```
yi-gu, line 1196: nit: s/revokes the account/has revoked the account out of band.

spec/index.bs
```diff
@@ -1082,6 +1106,115 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+        {{DOMException}}. This check can be performed by iterating over the
+        [=connected accounts set=] or by keeping a separate data structure to make this lookup fast.
+    1. Let |configUrl| be the result of running [=parse url=] with |provider|'s
+        {{IdentityProviderConfig/configURL}} and |globalObject|.
+    1. If |configUrl| is failure, throw an "{{InvalidStateError}}" {{DOMException}}.
+    1. If |document|'s [=Document/origin=] is not [=same origin=] as |configUrl|'s [=url/origin=],
+        throw an "{{InvalidStateError}}" {{DOMException}}.
+    1. Run a [[!CSP]] check with a [[CSP#directive-connect-src|connect-src]] directive on the URL
+        passed as |configUrl|. If it fails, throw a new "{{NetworkError}}" {{DOMException}}.
+    1. If there exists a pending {{IdentityProvider/getUserInfo()}} call for this |document|, throw
```
bvandersloot-mozilla, line 1176: Why?

spec/index.bs
```diff
@@ -1082,6 +1106,115 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+        a new "{{AbortError}}" {{DOMException}}.
+    1. If |globalObject|'s [=Window/navigable=] is a [=/top-level traversable=], throw a new
```
bvandersloot-mozilla, line 1178: Why is this restricted to iframes only?

spec/index.bs
```diff
@@ -1082,6 +1106,115 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
     1. Return |json|.
 </div>
 
+<!-- ============================================================ -->
```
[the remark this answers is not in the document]
bvandersloot-mozilla: This also feels like it is solving a very specific convenience-centered problem while having a more general name and putting itself in a general context.

If the IDP's iframe wants access to the account information and already has gone through a FedCM flow, why wouldn't they just use their token to request it themselves?

spec/index.bs
npm1: Changing to `login_hints`

spec/index.bs
npm1: Not sure what you mean, can you elaborate?

spec/index.bs
```diff
@@ -406,10 +406,20 @@ This specification introduces an extension to the {{CredentialRequestOptions}} o
 The {{IdentityCredentialRequestOptions}} contains a list of
 {{IdentityProviderConfig}}s that the [=RP=] supports and has
 pre-registered with (i.e. the [=IDP=] has given the [=RP=] a `clientId`).
+The {{IdentityCredentialRequestOptions}} also contains a {{IdentityCredentialRequestOptionsContext}}
+which the user agent can use to provide a more meaningful dialog to users.
 
 <xmp class=idl>
+enum IdentityCredentialRequestOptionsContext {
```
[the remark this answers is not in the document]
npm1: These are the ones we know of, but we can expect to get requests to add more over time

spec/index.bs
```diff
@@ -406,10 +406,20 @@ This specification introduces an extension to the {{CredentialRequestOptions}} o
...
+  "signin",
+  "signup",
+  "use",
+  "continue"
+};
+
 dictionary IdentityCredentialRequestOptions {
   sequence<IdentityProviderConfig> providers;
+  IdentityCredentialRequestOptionsContext context = "signin";
```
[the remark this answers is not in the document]
npm1: I don't think enums are nullable? But even if it was, the user agent needs to show some dialog, so it makes sense to pick a default context.

spec/index.bs
```diff
@@ -1082,6 +1106,115 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+<pre class="idl">
+  [Exposed=Window, SecureContext] dictionary IdentityUserInfo {
+    USVString email;
+    USVString name;
+    USVString givenName;
+    USVString picture;
+  };
+
+  [Exposed=Window, SecureContext] interface IdentityProvider {
+      static Promise&lt;sequence&lt;IdentityUserInfo&gt;?&gt; getUserInfo(IdentityProviderConfig config);
```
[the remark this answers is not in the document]
npm1: Not sure I follow your question. This call only provides one config. If there are multiple IDPs that support personalized buttons, they would each have to invoke this API separately (note it needs to be called in an IDP iframe).

spec/index.bs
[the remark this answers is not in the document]
npm1: We use two dictionaries because one is describing JSON values while the other is describing the dictionary returned in JavaScript. They do need to be different, both due to semantics as well as difference in naming conventions.

I don't think it makes sense to add this to IdentityCredential, as this API does not return a credential, and it also does not require the JavaScript to have a credential around. I'm not sure what problem you are trying to solve with this suggestion.

Regarding why the IDP iframe can't just use the token: the FedCM call is done in the RP, not the IDP iframe. As such, the token should not allow querying updated account information on its own: this is power the IDP should not grant to the RP automatically. Using the token would also require the IDP to store this token somewhere in cookies so it can perform this query. So there really is no way to do this in a reasonable way, in the absence of third-party cookies.

spec/index.bs
npm1: Probably not needed. It would certainly need to be tweaked for multi IDP support.

spec/index.bs
npm1: We can remove this condition but we of course need to keep the condition about the origin being the IDP's. If the caller is not an iframe, it is the top-level IDP page: why would it call this API? So this API really is only useful in IDP iframe (embedded within an RP). Hence the check. I don't have a strong opinion on keeping or removing this, wdyt?

spec/index.bs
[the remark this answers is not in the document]
bvandersloot-mozilla: Dictionary elements not declared `required` are optional. I think it may make sense to permit a generic option.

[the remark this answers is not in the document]
npm1: Correct, this value is optional. Doesn't "signin" feel like the generic option to you? Any idea on a better generic default value?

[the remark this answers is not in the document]
bvandersloot-mozilla: If this value is optional, the absence of a value would be a better default. That would presumably be treated as the generic option by user agents.

spec/index.bs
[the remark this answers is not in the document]
bvandersloot-mozilla: > We use two dictionaries because one is describing JSON values while the other is describing the dictionary returned in JavaScript. 

One benefit of WebIDL is that we can re-use a single definition for both.

> They do need to be different, both due to semantics as well as difference in naming conventions.

I'm not sure which semantics you mean here. And is there external pressure on the spec to keep separate naming conventions for field names for single values? If I'm looking for a given name and I have to remember whether this is the case where I have to use obj['given_name'] or  obj['givenName'], I would find that experience bad.

> I'm not sure what problem you are trying to solve with this suggestion.

Polluting the window namespace.

> Regarding why the IDP iframe can't just use the token: the FedCM call is done in the RP, not the IDP iframe. 

Sure, but presumably the RP can pass the token to the IDP iframe? 

> As such, the token should not allow querying updated account information on its own: this is power the IDP should not grant to the RP automatically

I thought there was a responsibility by the IDP to make the most recent identifiers available. Restricting updates to chrome UI rather than including RPs seems like a tradeoff.

spec/index.bs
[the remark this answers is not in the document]
bvandersloot-mozilla: 👍

spec/index.bs
asr-enid: For my understanding the FedCM API (as a means of a flow starting with a getCredential call from the RP) is mainly a precondition / facilitator here - the precondition is a "connected" state of a user with the RP and the information for the button is populated via the _account list endpoint_. In case the user chooses to click that button to sign-in this is happening outside of the FedCM API surface (say a classical OIDC based flow)?

spec/index.bs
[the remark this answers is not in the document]
npm1: Well it is just the text which changes: "Continue with idp.com on rp.com" vs "Use idp.com on rp.com". They are generic enough that I imagine there are cases where an RP could choose either, but we have both based on IDP demand.

spec/index.bs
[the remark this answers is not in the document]
npm1: Yea sure we could, although I think we'd need to change the dictionary parameter to be `IdentityCredentialRequestOptionsContext? context` because otherwise we could not pass null. @jyasskin what do you think is preferable here?

spec/index.bs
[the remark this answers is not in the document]
npm1: > One benefit of WebIDL is that we can re-use a single definition for both.
> 
> I'm not sure which semantics you mean here. And is there external pressure on the spec to keep separate naming conventions for field names for single values? If I'm looking for a given name and I have to remember whether this is the case where I have to use obj['given_name'] or obj['givenName'], I would find that experience bad.

I guess semantics are fairly similar so that would be fine. But the naming conventions are at https://www.w3.org/TR/design-principles/#casing-rules. I think they should be respected?

> Polluting the window namespace.

Your suggestion is to add the method to IdentityCredential instead of IdentityProvider. Either way the method is not polluting the namespace. What am I missing?

> Sure, but presumably the RP can pass the token to the IDP iframe?

This assumes the IDP wants to have control over the storage space in the RP. It would need to do so to store some token that it can then pass to its iframe later on. This seems undesirable: the IDP should not need to touch RP storage!

> I thought there was a responsibility by the IDP to make the most recent identifiers available. Restricting updates to chrome UI rather than including RPs seems like a tradeoff.

Sure, the IDP needs to make the most recent information available, whenever that information needs to be used (FedCM or some other form of federated login or authentication). That does not in any way imply that the IDP should permit the RP unrestricted queries into the accounts that the user has because the user once used FedCM on that RP.

spec/index.bs
npm1: Yep that's correct!

spec/index.bs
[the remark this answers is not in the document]
jyasskin: You usually don't want an options parameter that accepts both `null` and `undefined`, unless it's not confusing for the two to do different things. Do you need to be able to pass `null` explicitly, or does omitting the parameter work just as well?

I see that the only meaning of this option is "The user agent MAY use |options|'s context to customize the dialog." For most fields, there would be a specification of what the values do, and that specification would have to give a meaning to the "absent" option, which would tend to tell you what default to use. Do UAs want to diverge in how they treat an absent context?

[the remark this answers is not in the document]
npm1: Ah ok, so we could do `IdentityCredentialRequestOptionsContext context;` (so the default is undefined). The parameter is only used for the user agent to customize the title of their dialog shown to users, so not sure how we would specify the meaning of context when not passed / undefined. So I still have a preference to have it default to "signin" but I'm open to ideas on what to specify if the default is undefined.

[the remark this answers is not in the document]
jyasskin: It sounds like Chromium will treat an absent value the same as "signin". Will Firefox do the same, or something different? If Firefox will also treat an absent value like "signin", that should probably be the default value in the WebIDL. Otherwise, Ben's right in saying we should omit the default value.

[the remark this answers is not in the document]
npm1: That's correct, in Chrome our default value (the current shipped UI) says "sign in". Of course, Firefox can choose to not follow this and have different wording. Although ideally the browsers could be consistent about 'the rp context assumed when you pass no rp context'.

spec/index.bs
npm1: I'll resolve this but reopen if there's something that was not addressed.

spec/index.bs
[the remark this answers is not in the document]
bvandersloot-mozilla: We currently use sign in as our language as well, so that would likely be our default. We can have "signin" be the default then.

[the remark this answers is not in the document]
npm1: Ok, keeping this as is, and resolving. Thanks!

spec/index.bs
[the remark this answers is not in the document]
bvandersloot-mozilla: > I guess semantics are fairly similar so that would be fine. But the naming conventions are at https://www.w3.org/TR/design-principles/#casing-rules. I think they should be respected?

I hadn't seen that before. The difference between JSON keys and WebIDL attributes in convention is annoying, so perhaps that is out of scope for this.

> Your suggestion is to add the method to IdentityCredential instead of IdentityProvider. Either way the method is not polluting the namespace. What am I missing?

This patch creates the `window.IdentityProvider` interface. I don't think we need to for this change.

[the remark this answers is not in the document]
bvandersloot-mozilla: I hadn't been thinking about this in the context of agl's proposal... However if that is adopted, there is a very big breaking change happening that allows other things to move around. We could have a `partial interface IdentityCredential { FedCMCredentialUtilityFunctions fedcm }` that holds this function. 

Going through the existing experimental uses in Chrome[1]:
- this proposal
- sign-in status API that Sam[ proposed moving ](https://github.com/privacycg/is-logged-in/issues/53#issue-1664953653)to `navigator` 
- IDP registration (abandoned?)
- A new one to me! but that looks like it would be just as at home on IdentityCredential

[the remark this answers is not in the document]
bvandersloot-mozilla: Yeah, that doesn't quite make sense. That means that there isn't a great place for this to go. We could add an inline issue that we should evaluate where this goes long-term.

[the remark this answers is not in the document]
bvandersloot-mozilla: With an inline issue, I'd call this LGTM.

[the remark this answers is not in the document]
npm1: Added inline issue

PR #474, opened 2023-05-25
Add disclosureTextShown to the processing logic

npm1:
The `disclosure_text_shown` is supposed to be part of the ID assertion fetch. It is already somewhat part of the spec but was unfortunately not properly integrated into the processing model. This PR addresses that. The `disclosure_text_shown` solves the problem that the IDP wants assurance that the user has been presented the privacy policy and terms of service to use federated login, when they are new users. Since it is now the user agent that controls the UI, it can provide that assurance to the IDP via the ID assertion request.

***
<a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/474.html" title="Last updated on Jun 8, 2023, 1:50 PM UTC (d23fac5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/474/03a49d1...d23fac5.html" title="Last updated on Jun 8, 2023, 1:50 PM UTC (d23fac5)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1412,7 +1412,9 @@ It will also contain the following parameters in the request body `application/x
     :   <dfn>disclosure_text_shown</dfn>
     ::  Whether the user agent has explicitly shown to the user what specific information the
         [=IDP=] intends to share with the [=RP=] (e.g. "idp.example will share your name, email...
-        with rp.example"), used by the [=request permission to sign-up=] algorithm for new users.
+        with rp.example"), used by the [=request permission to sign-up=] algorithm for new users. It
+        is used as an assurance by the user agent to the [=IDP=] that it has indeed shown the terms
+        of service and privacy policy to the user in the cases where it is required to do so.
 </dl>
 
 For example:


yi-gu:
Thanks! LGTM

npm1:
This has a small normative addition. @bvandersloot-mozilla @cboozar or @martinthomson could one of you review this before I merge it? Thanks!

samuelgoto:
LGTM++

npm1:
Friendly ping

npm1:
> Does this mean that this variable is only for this one interaction? So if the user linked in the past and saw the disclosure text then- and then later reauthenticated there would be no "disclosureTextShown" on the later token request? If so, that is counterintuitive to me given the description.

That is correct. Why do you think it is counterintuitive?

bvandersloot-mozilla:
Before the algorithm was in place I assumed it was "has ever been shown the disclosure text".  I suppose that may be my own confusion. Basically, a "has opted-in on this browser" bit. That seems more useful to me, but I am not an IDP. That was a while ago that the field was added in the IDP section.

npm1:
> Before the algorithm was in place I assumed it was "has ever been shown the disclosure text". I suppose that may be my own confusion. Basically, a "has opted-in on this browser" bit. That seems more useful to me, but I am not an IDP. That was a while ago that the field was added in the IDP section.

Hmm ok. Well the only way to never show the disclosure text is if the IDP tells the user agent that this is a returning user via approved_clients, because otherwise the disclosure text will be shown if this user is considered to not have been seen before by the browser. So I think this would not be very useful to the IDP. So I do think the interesting one is "did you show the disclosure text in the UI in this particular instance" since this may or may not have happened depending on whether the browser sees this user as returning or not (assuming of course it does not show up in the approved_clients).

martinthomson (commented):
What problem does this pull request solve?

asr-enid:
> I believe this allows the IDP to receive confirmation about whether the privacy policy and terms of service are shown. 

The IDP must be aware both of the presented scope of information access (it would be good to provide a scope in any case btw.) and that that was shown in context of the tos/privacy policy, otherwise I can't proceed with minting a token from a legal perspective. Given FedCM as of now does not allow the IDP to shown its own UI to request a users approval.

martinthomson:
As a point for future reference, it is good practice to document why a change is being made.  Or rather, what problem it solves.  For a pull request, referencing an issue is acceptable, but the pull request comment here was empty.

FWIW, I think that the disclosure text feature is problematic, but if it already exists and this is just fixing what is effectively a spec bug, then we can take the question about whether disclosure notice stuff is something a browser should be taking responsibility for.

samuelgoto:
> then we can take the question about whether disclosure notice stuff is something a browser should be taking responsibility for.

Yes, I believe it should.

martinthomson:
I'm ok with landing technical fixes like this while the larger issue remains open.

PR #483, opened 2023-06-23
Returning accounts go first in getUserInfo

npm1:
This PR fixes the way an account is chosen as returning for getUserInfo: if approvedClients is available, that is considered the source of truth. Also, the returning list should list the returning accounts first, and then all the remaining accounts.

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1204,23 +1204,28 @@ When invoking the {{IdentityProvider/getUserInfo()}} method given an {{IdentityP
             {{DOMException}}.
         1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
             and |globalObject|.
-        1. Let |hasReturningAccount| be false.
-        1. For each |account| in |accountsList|:
-            1. If |account|["{{IdentityProviderAccount/approved_clients}}"] is not empty and it does not
-                [=list/contain=] |provider|'s {{IdentityProviderConfig/clientId}}, continue.
+        1. Let |isReturningAccount| be a new [=list=] of the same length as |accountsList|, with all
+            values initially set to false.
+        1. For each |i| from 0 to the length of |accountsList| minus 1:
+            1. Let |account| be |accountsList|[i].
+            1. If |account|["{{IdentityProviderAccount/approved_clients}}"] is not empty:
+                1. If |account|["{{IdentityProviderAccount/approved_clients}}"] [=list/contain=]
+                    |provider|'s {{IdentityProviderConfig/clientId}}, set |isReturningAccount|[i] to true.
             
                 Note: this allows the [=IDP=] to override whether an account is a returning account.
                 This could be useful for instance in cases where the user has revoked the account
                 out of band.
             
-            1. [=Compute the connection status=] of |provider|, |account|, and |globalObject|. If the
-                result is [=compute the connection status/connected=], set |hasReturningAccount| to
+            1. Otherwise, [=compute the connection status=] of |provider|, |account|, and |globalObject|. If the
+                result is [=compute the connection status/connected=], set |isReturningAccount|[i] to
                 true.
-        1. If |hasReturningAccount| is false, [=reject=] |promise| with a new "{{NetworkError}}"
-            {{DOMException}}.
+        1. If |isReturningAccount| does not [=list/contain=] true, [=reject=] |promise| with a new
+            "{{NetworkError}}" {{DOMException}}.
         1. Let |userInfoList| be a new [=list=].
-        1. For each |account| in |accountsList|:
-            1. [=list/Append=] an {{IdentityUserInfo}} to |userInfoList| with the following values:
+        1. Let |notReturningUserInfos| be a new [=list=].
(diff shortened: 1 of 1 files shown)


npm1:
@bvandersloot-mozilla please take a look

spec/index.bs
```diff
@@ -1204,23 +1204,28 @@ When invoking the {{IdentityProvider/getUserInfo()}} method given an {{IdentityP
...
-        1. For each |account| in |accountsList|:
-            1. If |account|["{{IdentityProviderAccount/approved_clients}}"] is not empty and it does not
-                [=list/contain=] |provider|'s {{IdentityProviderConfig/clientId}}, continue.
+        1. Let |isReturningAccount| be a new [=list=] of the same length as |accountsList|, with all
+            values initially set to false.
+        1. For each |i| from 0 to the length of |accountsList| minus 1:
+            1. Let |account| be |accountsList|[|i|].
+            1. If |account|["{{IdentityProviderAccount/approved_clients}}"] is not empty:
+                1. If |account|["{{IdentityProviderAccount/approved_clients}}"] [=list/contain=]
+                    |provider|'s {{IdentityProviderConfig/clientId}}, set |isReturningAccount|[|i|] to true.
```
cbiesinger, line 1213: This does not match what Chrome has implemented, is that intentional?

Chrome's behavior is:
- if approved_clients exists and client_id not in approved_clients, isReturningAccount=false
- else if compute connection status!=connected, isReturningAccount=false
- else isReturningAccount=true

npm1: Oh yes, I think you're right that this is wrong. We want the case where approved clients exists and says that the user is returning, but there is no connection status from the browser, to be considered as new user.

npm1:
Ping

cbiesinger (approved):
not sure if you were waiting for me here but lgtm

npm1:
I noticed this PR is still relevant. Rebased since it was pretty old but still want to merge this. Anyone have any further comments?

spec/index.bs
```diff
@@ -1635,23 +1635,27 @@ When invoking the {{IdentityProvider/getUserInfo()}} method given an {{IdentityP
             {{DOMException}}.
         1. Let |accountsList| be the result of [=fetch the accounts list=] with |config|, |provider|,
             and |globalObject|.
-        1. Let |hasReturningAccount| be false.
-        1. For each |account| in |accountsList|:
+        1. Let |isReturningAccount| be a new [=list=] of the same length as |accountsList|, with all
```
bvandersloot-mozilla, line 1638: Should this be on the accounts, rather than adding a new boolean?

spec/index.bs
```diff
@@ -1635,23 +1635,27 @@ When invoking the {{IdentityProvider/getUserInfo()}} method given an {{IdentityP
...
-            {{DOMException}}.
+        1. If |isReturningAccount| does not [=list/contain=] true, [=reject=] |promise| with a new
+            "{{NetworkError}}" {{DOMException}}.
         1. Let |userInfoList| be a new [=list=].
-        1. For each |account| in |accountsList|:
-            1. [=list/Append=] an {{IdentityUserInfo}} to |userInfoList| with the following values:
+        1. Let |notReturningUserInfos| be a new [=list=].
+        1. For each |i| from 0 to the length of |accountsList| minus 1:
+            1. Let |account| be |accountsList|[|i|].
+            1. Let |userInfo| be an {{IdentityUserInfo}} with the following values:
```
bvandersloot-mozilla, line 1658: Or at least in the IdentityUserInfo? that way we can expose it to the webpage

bvandersloot-mozilla (changes requested):
Computing a secret bit and using it to re-sort a list returned to the webpage is not a good pattern. If we are going to alter the output, it should at least be observable why. 
This is hard to work with. Is `accounts[0]` previously used? `accounts[n-1]`?

spec/index.bs
npm1: The accounts are dictionaries parsed from the IdP's response, so we cannot add properties to them.

spec/index.bs
npm1: This seems like a broader change (adding a webexposed member to the IdentityUserInfo. The goal of this PR is to fix a bug in the existing returned values.

npm1:
> Computing a secret bit and using it to re-sort a list returned to the webpage is not a good pattern. If we are going to alter the output, it should at least be observable why. This is hard to work with. Is `accounts[0]` previously used? `accounts[n-1]`?

The idea is that `accounts[0]` is previously used. @yi-gu do you remember why we do not expose whether an account is returning or not explicitly in this API? It's not clear to me.

bvandersloot-mozilla:
> The idea is that accounts[0] is previously used.
But you can't guarantee that.

npm1:
> > The idea is that accounts[0] is previously used.
> > But you can't guarantee that.

You can. `getUserInfo` only returns anything if there are returning accounts. And we surface returning accounts first, followed by other accounts. So `accounts[0]` is a returning account.

PR #486, opened 2023-07-06
Add IdentityProvider.close() to the explainer

cbiesinger:


proposals/idp-sign-in-status-api.md
```diff
@@ -60,11 +60,20 @@ And IdP can alternatively call the IdP Sign-in Status API via JS calls through t
 interface IdentityProvider {
   static void login();
   static void logout();
+
+  static void close();
 }
 ```
 
 To be called from the IDP's origin, and marks the current origin as signed in or signed out.
```
npm1, line 68: Maybe move to the previous paragraph?

npm1 (approved):

proposals/idp-sign-in-status-api.md
TallTed: Also note that the first word of the previous paragraph should be "An", rather than "And".

cbiesinger: Thanks, fixed both comments.

PR #498, opened 2023-08-30
Add API to show error messages from failed token fetches

npm1:
Adds the capability for the FedCM API to show error dialogs in certain scenarios after the user has chosen to perform federated login with an account. For this purpose:

* An `IdentityCredentialError` is created, which may contain an `error` which is a string that contains the specific error, and the user agent may use to customize the UI). It may also contain `url`, in case the IDP wants to show a url where the user could get more information.
* Shows error UI when there are network errors or when the IDP returns that an error has occurred.

Fixes https://github.com/fedidcg/FedCM/issues/488

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/pull/498.html" title="Last updated on Jun 20, 2025, 6:09 PM UTC (5abb21d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/498/a17bc50...5abb21d.html" title="Last updated on Jun 20, 2025, 6:09 PM UTC (5abb21d)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -609,38 +609,6 @@ dictionary DisconnectedAccount {
   required USVString account_id;
 };
 </xmp>
-## The IdentityCredentialError Interface ## {#browser-api-identity-credential-error-interface}
-<!-- ============================================================ -->
-
-This specification introduces a new type of {{DOMException}}, the {{IdentityCredentialError}}. It
-is used when the [=IDP=] cannot create an {{IdentityCredential}} after the user has requested to use
-a federated account.
-
-<pre class="idl">
-  [Exposed=Window, SecureContext]
-  interface IdentityCredentialError: DOMException {
-    constructor(optional DOMString message = "", IdentityCredentialErrorInit options);
-    readonly attribute DOMString code;
-    readonly attribute DOMString url;
-  };
-</pre>
-
-<div algorithm="IdentityCredentialError constructor">
-The {{IdentityCredentialError/constructor()}}, given a |realm|, a {{DOMString}} |message| and a
-    {{IdentityCredentialErrorInit}} |options| must run the following steps:
-    1. Invoke the {{DOMException}}'s {{DOMException/constructor()}}, passing |realm| and |message|.
-    1. Set <a>this</a>.{{IdentityCredentialError/code}} to |options|.{{IdentityCredentialError/code}}.
-    1. Set <a>this</a>.{{IdentityCredentialError/url}} to |options.{{IdentityCredentialError/url}}.
-</div>
-
-<dl>
-    :   <b>{{IdentityCredentialError/code}}</b>
-    ::  The {{IdentityCredentialError/code}}'s attribute getter returns the value it is set to.
-        It represents the type of error which resulted in an {{IdentityCredential}} not being created.
-    :   <b>{{IdentityCredentialError/url}}</b>
-    ::  The {{IdentityCredentialError/url}}'s attribute getter returns the value it is set to.
-        It represents a URL where the user can learn more information about the error.
-</dl>
 
 <!-- ============================================================ -->
 ## The IdentityCredentialError Interface ## {#browser-api-identity-credential-error-interface}
@@ -1579,13 +1547,6 @@ An extension may add the following steps after the [=fetch identity assertion/cr
 dictionary IdentityAssertionResponse {
   USVString token;
   USVString continue_on;
-};
-</xmp>
-
-<xmp class="idl">
-dictionary IdentityCredentialErrorInit {
-  DOMString code;
-  DOMString url;
 };
 </xmp>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -635,8 +635,8 @@ dictionary IdentityCredentialErrorInit {
 </pre>
 
 <div algorithm="IdentityCredentialError constructor">
-The {{IdentityCredentialError/constructor()}}, given a |realm|, a {{DOMString}} |message| and a
-    {{IdentityCredentialErrorInit}} |options| must run the following steps:
+Given a |realm|, a {{DOMString}} |message|, and an {{IdentityCredentialErrorInit}} |options|, the
+{{IdentityCredentialError/constructor()}} must run the following steps:
     1. Invoke the {{DOMException}}'s {{DOMException/constructor()}}, passing |realm| and |message|.
     1. Set <a>this</a>.{{IdentityCredentialError/error}} to |options|.{{IdentityCredentialError/error}}
         if it is present in |options|, or to "" otherwise.
@@ -826,7 +826,7 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
                 However, UAs may have different UI approaches here and prevent it in a different way.
         1. [=Queue a global task=] on the [=DOM manipulation task source=]
             to throw a new "{{NetworkError}}" {{DOMException}}.
-    1. Otherwise, return |credential| (this may throw since it may be an
+    1. Else, return |credential| (this may throw, since it may be an
         {{IdentityCredentialError}}).
 </div>
 
@@ -1036,8 +1036,8 @@ throwing the exception.
             the user agent MAY show some UI to the user indicating that they are being
             <dfn>auto-reauthenticated</dfn>.
         1. Set |isAutoSelected| to true.
-    1. Otherwise, if |mediation| is "{{CredentialMediationRequirement/silent}}", return (failure, false).
-    1. Otherwise, if |accountsList|'s size is 1:
+    1. Else if |mediation| is "{{CredentialMediationRequirement/silent}}", return (failure, false).
+    1. Else if |accountsList|'s size is 1:
         1. Set |account| to |accountsList|[0].
         1. Set |accountState| to the result of running the [=compute the connection status=] algorithm
             given |provider|, |account|, and |globalObject|.
@@ -1045,10 +1045,10 @@ throwing the exception.
             let |permission| be the result of running [=request permission to sign-up=] algorithm
             with |account|, |accountState|, |config|, |provider|, and |globalObject|. Also set
             |disclosureTextShown| to true.
-        1. Otherwise, show a dialog to request user permission to sign in via |account|, and set the
+        1. Else, show a dialog to request user permission to sign in via |account|, and set the
             result in |permission|. The user agent MAY use |options|'s
             {{IdentityCredentialRequestOptions/context}} to customize the dialog.
-    1. Otherwise:
+    1. Else:
         1. Set |account| to the result of running the [=select an account=] from the
             |accountsList|.
         1. If |account| is failure, return (failure, true).
@@ -1058,7 +1058,7 @@ throwing the exception.
             1. Let |permission| be the result of running the [=request permission to sign-up=]
                 algorithm with |account|, |config|, |provider|, and |globalObject|.
             1. Set |disclosureTextShown| to true.
-        1. Otherwise, set |permission| to true.
+        1. Else, set |permission| to true.
     1. Wait until the [=user agent=]'s dialogs requesting for user choice or permission to be
         closed, if any are created in the previous steps.
     1. Assert: |account| is not null.
@@ -1080,8 +1080,8 @@ throwing the exception.
             in its UI so the user can click on it to learn more information about the error message.
         1. Wait until the UI is acknowledged or dismissed by the user.
         
-        Note: this wait means the [=user agent=] rejects the call only after the UI has been
-            closed by the user.
+          Note: this wait means the [=user agent=] rejects the call only after the UI has been
+              closed by the user.
     1. Return |credential|.
 </div>
 
@@ -1500,7 +1500,7 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
         1. Otherwise, i.e. if |json|[|error|] [=map/exists=]:
             1. If |json|[|error|] is not an [=ordered map=], set |credential| to a new
                 {{IdentityCredentialError}} given |globalObject|'s [=global object/realm=] and
-                "IdentityCredentialError" and return.
+                "IdentityCredentialError", and return.
             1. [=converted to an IDL value|Convert=] |json| to an {{IdentityCredentialErrorInit}},
                 |errorInit|.
             1. If |errorInit|.{{IdentityCredentialErrorInit/url}} is present:
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -611,6 +611,40 @@ dictionary DisconnectedAccount {
 </xmp>
 
 <!-- ============================================================ -->
+## The IdentityCredentialError Interface ## {#browser-api-identity-credential-error-interface}
+<!-- ============================================================ -->
+
+This specification introduces a new type of {{DOMException}}, the {{IdentityCredentialError}}. It
+is used when the [=IDP=] cannot create an {{IdentityCredential}} after the user has requested to use
+a federated account.
+
+<pre class="idl">
+  [Exposed=Window, SecureContext]
+  interface IdentityCredentialError: DOMException {
+    constructor(optional DOMString message = "", IdentityCredentialErrorInit options);
+    readonly attribute DOMString code;
+    readonly attribute DOMString url;
+  };
+</pre>
+
+<div algorithm="IdentityCredentialError constructor">
+The {{IdentityCredentialError/constructor()}}, given a |realm|, a {{DOMString}} |message| and a
+    {{IdentityCredentialErrorInit}} |options| must run the following steps:
+    1. Invoke the {{DOMException}}'s {{DOMException/constructor()}}, passing |realm| and |message|.
+    1. Set <a>this</a>.{{IdentityCredentialError/code}} to |options|.{{IdentityCredentialError/code}}.
+    1. Set <a>this</a>.{{IdentityCredentialError/url}} to |options.{{IdentityCredentialError/url}}.
+</div>
+
+<dl>
+    :   <b>{{IdentityCredentialError/code}}</b>
+    ::  The {{IdentityCredentialError/code}}'s attribute getter returns the value it is set to.
+        It represents the type of error which resulted in an {{IdentityCredential}} not being created.
+    :   <b>{{IdentityCredentialError/url}}</b>
+    ::  The {{IdentityCredentialError/url}}'s attribute getter returns the value it is set to.
+        It represents a URL where the user can learn more information about the error.
+</dl>
+
+<!-- ============================================================ -->
 ### The CredentialRequestOptions ### {#browser-api-credential-request-options}
 <!-- ============================================================ -->
 
@@ -745,7 +779,7 @@ requests.
 When the {{IdentityCredential}}'s
 <dfn for="IdentityCredential" method>\[[DiscoverFromExternalSource]](origin, options, sameOriginWithAncestors)</dfn>
 algorithm is invoked, the user agent MUST execute the following steps. This returns an
-{{IdentityCredential}} (or throws an error to the caller).
+{{IdentityCredential}} or throws an error to the caller.
 
     1. Assert: These steps are running [=in parallel=].
     1. If
@@ -765,10 +799,10 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
         algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
 
     1. If |credential| is a pair:
-        1. Let |throwImmediately| be the value of the second element of the pair.
+        1. Let |delayException| be the value of the second element of the pair.
         1. The user agent SHOULD wait a random amount of time
             before the next step if all of the following conditions hold:
-            * |throwImmediately| is false
+            * |delayException| is true
             * The <dfn>promise rejection delay</dfn> was not disabled by
                 [=setdelayenabled|user agent automation=]
             * The user agent has not implemented another way to prevent
@@ -783,7 +817,8 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
                 However, UAs may have different UI approaches here and prevent it in a different way.
         1. [=Queue a global task=] on the [=DOM manipulation task source=]
             to throw a new "{{NetworkError}}" {{DOMException}}.
-    1. Otherwise, return |credential|.
+    1. Otherwise, return |credential| (this may throw since it may be an
+        {{IdentityCredentialError}}).
 </div>
 
 <!-- ============================================================ -->
@@ -791,14 +826,15 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
 <!-- ============================================================ -->
 
 The <a>create an IdentityCredential</a> algorithm invokes the various FedCM fetches, shows the user
-agent UI, and creates the {{IdentityCredential}} that is then returned to the [=RP=].
+agent UI, and creates the {{IdentityCredential}} or {{IdentityCredentialError}} which is then
+returned to the [=RP=].
 
 <div algorithm="create an IdentityCredential">
-To <dfn>create an IdentityCredential</dfn> given a [=sequence=] of  {{IdentityProviderRequestOptions}}
-|providerList|, a {{CredentialRequestOptions}} |options|, and a
-|globalObject|, run the following steps. This returns an {{IdentityCredential}}
-or a pair (failure, bool), where the bool indicates whether to skip delaying
-the exception thrown.
+To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOptions}}
+|provider|, a {{CredentialRequestOptions}} |options|, and a
+|globalObject|, run the following steps. This returns an {{IdentityCredential}}, an
+{{IdentityCredentialError}}, or a pair (failure, bool), where the bool indicates whether to delay
+throwing the exception.
     1. Assert: These steps are running [=in parallel=].
     1. Let |mode| be |options|'s {{IdentityCredentialRequestOptions/mode}}.
     1. If |mode| is {{IdentityCredentialRequestOptionsMode/active}}:
@@ -942,26 +978,101 @@ the exception thrown.
 
             * If the [=show an IDP login dialog=] algorithm was triggered:
 
-                1. Let |result| be the result of that algorithm.
-                1. If |result| is failure, return (failure, true). The user
-                    agent MAY show a dialog to the user before or after
-                    returning failure indicating this failure.
-                1. Otherwise, go back to the [=fetch accounts step=] to get an updated
-                    value of |providerMap| for this [=IDP=].
-        1. Otherwise, |value| is a [=list=] of accounts. [=list/Extend=] |allAccounts| with |value|.
-    1. Also include a UI affordance to close the dialog. If the user closes this dialog, return (failure,
-        true).
-    1. <dfn for="create identity credential">Show accounts</dfn> step: if |allAccounts| is not [=list/empty=], also add UI to present the account options to the user.
-        If the user selects an account, perform the following steps:
-            1. Set |selectedAccount| to the chosen {{IdentityProviderAccount}}.
-            1. Close the dialog.
-    1. If the [=user agent=] created any dialogs requesting user choice or permission in the previous
-        steps, wait until they are closed.
+                * If the user closes the dialog, return (failure, true).
+
+                * If the [=show an IDP login dialog=] algorithm was triggered:
+
+                    1. Let |result| be the result of that algorithm.
+                    1. If |result| is failure, return (failure, true). The user
+                        agent MAY show a dialog to the user before or after
+                        returning failure indicating this failure.
+                    1. Otherwise, go back to the [=fetch accounts step=].
+
+    1. Assert: |accountsList| is not failure and the size of |accountsList| is not 0.
+    1. [=Set the login status=] for the [=/origin=] of the
+        {{IdentityProviderConfig/configURL}} to [=logged-in=].
+    1. If |provider|'s {{IdentityProviderRequestOptions/loginHint}} is not empty:
+        1. For every |account| in |accountList|, remove |account| from |accountList| if |account|'s
+            {{IdentityProviderAccount/login_hints}} does not [=list/contain=] |provider|'s
+            {{IdentityProviderRequestOptions/loginHint}}.
+        1. If |accountList| is now empty, go to the [=mismatch dialog step=].
+    1. If |provider|'s {{IdentityProviderRequestOptions/domainHint}} is not empty:
+        1. For every |account| in |accountList|:
+            1. If {{IdentityProviderRequestOptions/domainHint}} is "any":
+                1. If |account|'s {{IdentityProviderAccount/domain_hints}} is empty, remove
+                    |account| from |accountList|.
+            1. Otherwise, remove |account| from |accountList| if |account|'s
+                {{IdentityProviderAccount/domain_hints}} does not [=list/contain=] |provider|'s
+                {{IdentityProviderRequestOptions/domainHint}}.
+        1. If |accountList| is now empty, go to the [=mismatch dialog step=].
+    1. For each |acc| in |accountsList|:
+        1. If |acc|["{{IdentityProviderAccount/picture}}"] is present, [=fetch the account picture=]
+            with |acc| and |globalObject|.
+
+        Note: The [=user agent=] may choose to show UI which does not initially require fetching the
+        account pictures. In these cases, the [=user agent=] may delay these fetches until they are
+        needed. Because errors from these fetches are ignored, they can happen in any order.
+    1. Let |registeredAccount|, |numRegisteredAccounts| be null and 0, respectively.
+    1. Let |account| be null.
+    1. For each |acc| in |accountsList|:
+        1. Let |accState| be the result of running the [=compute the connection status=] algorithm given
+            |provider| and |acc|.
+        1. If |accState| is [=compute the connection status/connected=], set |registeredAccount| to
+            |acc| and increase |numRegisteredAccounts| by 1.
+    1. Let |permission|, |disclosureTextShown|, and |isAutoSelected| be set to false.
+    1. If |mediation| is not "{{CredentialMediationRequirement/required}}", |requiresUserMediation|
+        is false, and |numRegisteredAccounts| is equal to 1:
+        1. Set |account| to |registeredAccount| and |accountState| to the result of running
+            [=compute the connection status=] algorithm given |provider| and |account|. When doing this,
+            the user agent MAY show some UI to the user indicating that they are being
+            <dfn>auto-reauthenticated</dfn>.
+        1. Set |isAutoSelected| to true.
+    1. Otherwise, if |mediation| is "{{CredentialMediationRequirement/silent}}", return (failure, false).
+    1. Otherwise, if |accountsList|'s size is 1:
+        1. Set |account| to |accountsList|[0].
+        1. Set |accountState| to the result of running the [=compute the connection status=] algorithm
+            given |provider|, |account|, and |globalObject|.
+        1. If |accountState| is [=compute the connection status/disconnected=],
+            let |permission| be the result of running [=request permission to sign-up=] algorithm
+            with |account|, |accountState|, |config|, |provider|, and |globalObject|. Also set
+            |disclosureTextShown| to true.
+        1. Otherwise, show a dialog to request user permission to sign in via |account|, and set the
+            result in |permission|. The user agent MAY use |options|'s
+            {{IdentityCredentialRequestOptions/context}} to customize the dialog.
+    1. Otherwise:
+        1. Set |account| to the result of running the [=select an account=] from the
+            |accountsList|.
+        1. If |account| is failure, return (failure, true).
+        1. Set |accountState| to the result of running the [=compute the connection status=] algorithm
+            given |provider| and |account|.
+        1. If |accountState| is [=compute the connection status/disconnected=]:
+            1. Let |permission| be the result of running the [=request permission to sign-up=]
+                algorithm with |account|, |config|, |provider|, and |globalObject|.
+            1. Set |disclosureTextShown| to true.
+        1. Otherwise, set |permission| to true.
+    1. Wait until the [=user agent=]'s dialogs requesting for user choice or permission to be
+        closed, if any are created in the previous steps.
+    1. Assert: |account| is not null.
     1. If |permission| is false, then return (failure, true).
     1. Assert: |selectedAccount| is not null.
     1. Let |credential| be the result of running the [=fetch an identity assertion=] algorithm with
         |selectedAccount|'s {{IdentityProviderAccount/id}}, |permissionRequested|, |isAutoSelected|,
         |provider|, |config|, and |globalObject|.
+    1. If |credential| is an {{IdentityCredentialError}}:
+        1. The [=user agent=] MUST show UI to the user indicating that the identity assertion fetch
+            has failed.
+        1. The [=user agent=] MAY use the {{IdentityCredentialError/code}} in its UI to provide a
+            more specific error message to the user. For instance, if the value is one of
+            "invalid_request", "unauthorized_client", "access_denied", "server_error", and
+            "temporarily_unavailable", the [=user agent=] may note the reason in a user-friendly
+            manner. See [here](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1) for
+            more details about these.
+        1. The [=user agent=] MAY use the {{IdentityCredentialError/url}} to display a hyperlink
+            in its UI so the user can click on it to learn more information about the error message.
+        1. Wait until the UI is acknowledged or dismissed by the user.
+        
+        Note: this wait means the [=user agent=] rejects the call only after the UI has been
+            dismissed by the user.
     1. Return |credential|.
 </div>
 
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -541,22 +575,24 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
                 However, UAs may have different UI approaches here and prevent it in a different way.
         1. [=Queue a global task=] on the [=DOM manipulation task source=]
             to throw a new "{{NetworkError}}" {{DOMException}}.
-    1. Otherwise, return |credential|.
+    1. Otherwise, return |credential| (this may throw since it may be an
```
yi-gu, line 578: Since we only throw `IdentityCredentialError` post user permission, should we note that `delayException` should be false?

spec/index.bs
```diff
@@ -624,6 +661,21 @@ the exception thrown.
     1. Let |credential| be the result of running the [=fetch an identity assertion=] algorithm with
         |account|'s {{IdentityProviderAccount/id}}, |disclosureTextShown|, |provider|, |config|, and
         |globalObject|.
+    1. If |credential| is an {{IdentityCredentialError}}:
+        1. The [=user agent=] MUST show UI to the user indicating that the identity assertion fetch
```
yi-gu, line 665: Should this be MAY instead of MUST given that the Error API is optional for UA?

spec/index.bs
```diff
@@ -624,6 +661,21 @@ the exception thrown.
...
+            has failed.
+        1. The [=user agent=] MAY use the {{IdentityCredentialError/code}} in its UI to provide a
+            more specific error message to the user. For instance, if the value is one of
+            "invalid_request", "unauthorized_client", "access_denied", "server_error", and
+            "temporarily_unavailable", the [=user agent=] may note the reason in a user-friendly
+            manner. See [here](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1) for
+            more details about these.
+        1. The [=user agent=] MAY use the {{IdentityCredentialError/url}} to display a hyperlink
```
yi-gu, line 673: Should we use more generic terms for "hyperlink" to allow UA to use different UI affordance to proceed to the url?

spec/index.bs
```diff
@@ -624,6 +661,21 @@ the exception thrown.
...
+            in its UI so the user can click on it to learn more information about the error message.
+        1. Wait until the UI is acknowledged or dismissed by the user.
+        
+        Note: this wait means the [=user agent=] rejects the call only after the UI has been
+            dismissed by the user.
```
yi-gu, line 678: s/dismissed/closed to cover the "acknowledged" part.

spec/index.bs
```diff
@@ -893,16 +945,20 @@ To <dfn>fetch the account picture</dfn> given an {{IdentityProviderAccount}} |ac
...
 <div algorithm>
 To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
     |accountId|, a boolean |disclosureTextShown|, an {{IdentityProviderConfig}} |provider|, an
     {{IdentityProviderAPIConfig}} |config|, and |globalObject|, run the following steps. This
-    returns an {{IdentityCredential}} or failure.
+    returns an {{IdentityCredential}} or an {{IdentityCredentialError}}.
     1. Let |tokenUrl| be the result of [=computing the manifest URL=] given |provider|,
         |config|["{{IdentityProviderAPIConfig/id_assertion_endpoint}}"], and |globalObject|.
-    1. If |tokenUrl| is failure, return failure.
+    1. If |tokenUrl| is failure:
```
yi-gu, line 957: How would `tokenUrl` be failure? I think we should have validated all the urls after fetching the manifest from configURL.

spec/index.bs
```diff
@@ -961,6 +1028,13 @@ dictionary IdentityProviderToken {
 };
 </xmp>
 
+<xmp class="idl">
+dictionary IdentityCredentialErrorInit {
```
yi-gu, line 1032: Why do we need "Init" in the idl?

spec/index.bs
npm1: No, this is handled separately as we do not return a pair in this case.

spec/index.bs
npm1: Isn't the whole point of this API to ensure error UI is shown? That is what Chrome intends to ship, we should only make this optional if other vendors feel they do not want to show these errors.

spec/index.bs
npm1: Hyperlink is generic and not tied to a specific UI.

spec/index.bs
npm1: It seems that is not the case right now. Filed https://github.com/fedidcg/FedCM/issues/499 but I will not fix that in this PR.

spec/index.bs
npm1: Because it should not be the same as the exception name?

npm1:
Ready for another look @yi-gu

yi-gu (approved):
Thanks. LGTM!

npm1:
Hey @bvandersloot-mozilla @martinthomson @cboozar requesting Mozilla to take a look.

spec/index.bs
```diff
@@ -945,12 +1003,23 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
             |responseBody|.
-        1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderToken}}, |token|.
-        1. If one of the previous two steps threw an exception, set |credential| to failure
-            and return.
-        1. Let |credential| be a new {{IdentityCredential}} given |globalObject|'s
-            <a for="global object">realm</a>.
-        1. Set |credential|'s {{IdentityCredential/token}} to |token|.
+        1. If |json|[|token|] and |json|[|error|] both [=map/exist=] or both do not [=map/exist=],
+            set |credential| to a new {{IdentityCredentialError}} given |globalObject|'s
+            <a for="global object">realm</a> and "IdentityCredentialError", and return.
```
cbiesinger, line 1008: If both exist, should error's code be used for the IdentityCredentialError constructor?

spec/index.bs
```diff
@@ -945,12 +1003,23 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
+        1. If |json|[|token|] [=map/exists=]:
+            1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderToken}}, |token|.
+            1. If one of the previous steps threw an exception, set |credential| to a new
+                {{IdentityCredentialError}} given |globalObject|'s <a for="global object">realm</a>
```
cbiesinger, line 1012: why does this use an IdentityCredentialError...

spec/index.bs
```diff
@@ -945,12 +1003,23 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
+                and "IdentityCredentialError", and return.
+            1. Let |credential| be a new {{IdentityCredential}} given |globalObject|'s
+                <a for="global object">realm</a>.
+            1. Set |credential|'s {{IdentityCredential/token}} to |token|.
+        1. Otherwise, i.e. if |json|[|error|] [=map/exists=]:
+            1. If |json|[|error|] is not an [=ordered map=], set |credential| to failure and return.
```
cbiesinger, line 1018: ...but this sets credential to failure?

spec/index.bs
```diff
@@ -386,7 +386,43 @@ The main entrypoint in this specification is through the entrypoints exposed
...
+<!-- ============================================================ -->
+
+This specification introduces a new type of {{DOMException}}, the {{IdentityCredentialError}}. It
+is used when the [=user agent=] does not receive an identity assertion after the user has requested
+to use a federated account.
+
+<pre class="idl">
+  [Exposed=Window, SecureContext]
+  interface IdentityCredentialError: DOMException {
+    constructor(optional DOMString message = "", optional IdentityCredentialErrorInit options = {});
```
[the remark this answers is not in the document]
npm1: Moving up

spec/index.bs
```diff
@@ -386,7 +386,43 @@ The main entrypoint in this specification is through the entrypoints exposed
...
+    readonly attribute DOMString code;
```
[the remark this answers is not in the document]
npm1: I want to allow the IDP to use their own codes even if they do not mean anything to the user agent and they result in the generic error UI.

spec/index.bs
npm1: Hmm maybe. I thought of this one as developer error: should token be trusted or error?

spec/index.bs
npm1: We now only allow returning IdentityCredential or this kind of error

spec/index.bs
npm1: Missed this one

cbiesinger (approved):
Thanks, lgtm

spec/index.bs
```diff
@@ -1102,7 +1180,10 @@ When <dfn>computing the manifest URL</dfn> given an {{IdentityProviderConfig}} |
         allowed.
 
     1. If |manifestUrl| is failure, return failure.
-    1. If |manifestUrl| is not [=same origin=] with |configUrl|, return failure.
+    1. If |requireSameOrigin| and |manifestUrl| is not [=same origin=] with |configUrl|, return
+        failure.
+    1. If |requireSameOrigin| is false and |manifestUrl|'s <a spec=url>domain</a> is not equal to
```
cbiesinger, line 1185: Shouldn't we use the registrable domain here? https://url.spec.whatwg.org/#host-registrable-domain

spec/index.bs
```diff
@@ -386,7 +386,50 @@ The main entrypoint in this specification is through the entrypoints exposed
...
+  interface IdentityCredentialError : DOMException {
+    constructor(optional DOMString message = "", optional IdentityCredentialErrorInit options = {});
+    readonly attribute DOMString code;
+    readonly attribute DOMString url;
+  };
+</pre>
+
+<div algorithm="IdentityCredentialError constructor">
+The {{IdentityCredentialError/constructor()}}, given a |realm|, a {{DOMString}} |message| and a
+    {{IdentityCredentialErrorInit}} |options| must run the following steps:
```
TallTed, line 414:
```suggestion
Given a |realm|, a {{DOMString}} |message|, and an {{IdentityCredentialErrorInit}}
|options|, the {{IdentityCredentialError/constructor()}} must run the following steps:
```

spec/index.bs
```diff
@@ -386,7 +386,50 @@ The main entrypoint in this specification is through the entrypoints exposed
...
+    1. Invoke the {{DOMException}}'s {{DOMException/constructor()}}, passing |realm| and |message|.
+    1. Set <a>this</a>.{{IdentityCredentialError/code}} to |options|.{{IdentityCredentialError/code}}
+        if it is present in |options|, or to "" otherwise.
+    1. Set <a>this</a>.{{IdentityCredentialError/url}} to |options|.{{IdentityCredentialError/url}}
+        if it is present in |options|, or to "" otherwise.
```
TallTed, line 419:
```suggestion
    1. If |options|.{{IdentityCredentialError/code}} is present in |options|,
       set <a>this</a>.{{IdentityCredentialError/code}} to it; otherwise, set
       <a>this</a>.{{IdentityCredentialError/code}} to "".
    1. If |options|.{{IdentityCredentialError/url}} is present in |options|,
       set <a>this</a>.{{IdentityCredentialError/url}} to it; otherwise, set
       <a>this</a>.{{IdentityCredentialError/url}} to "".
```

spec/index.bs
```diff
@@ -386,7 +386,50 @@ The main entrypoint in this specification is through the entrypoints exposed
...
+</div>
+
+<dl>
+    :   <b>{{IdentityCredentialError/code}}</b>
+    ::  The {{IdentityCredentialError/code}}'s attribute getter returns the value it is set to.
```
TallTed, line 424:
```suggestion
    ::  The {{IdentityCredentialError/code}}'s attribute getter returns the value to which it is set.
```

spec/index.bs
```diff
@@ -386,7 +386,50 @@ The main entrypoint in this specification is through the entrypoints exposed
...
+        It represents the type of error which resulted in an {{IdentityCredential}} not being created.
+    :   <b>{{IdentityCredentialError/url}}</b>
+    ::  The {{IdentityCredentialError/url}}'s attribute getter returns the value it is set to.
```
TallTed, line 427:
```suggestion
    ::  The {{IdentityCredentialError/url}}'s attribute getter returns the value to which it is set.
```

spec/index.bs
```diff
@@ -945,12 +1010,30 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
+            [=global object/realm=] and "IdentityCredentialError", and return.
+        1. If |json|[|token|] [=map/exists=]:
+            1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderToken}}, |token|.
+            1. If one of the previous steps threw an exception, set |credential| to a new
+                {{IdentityCredentialError}} given |globalObject|'s [=global object/realm=]
+                and "IdentityCredentialError", and return.
+            1. Let |credential| be a new {{IdentityCredential}} given |globalObject|'s
+                [=global object/realm=].
+            1. Set |credential|'s {{IdentityCredential/token}} to |token|.
+        1. Otherwise, i.e. if |json|[|error|] [=map/exists=]:
```
TallTed, line 1024:
```suggestion
        1. Otherwise (i.e., if |json|[|error|] [=map/exists=]):
```

TallTed (changes requested):

spec/index.bs
```diff
@@ -554,6 +554,49 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
...
+  DOMString code;
+  DOMString url;
+};
+</xmp>
+
+<pre class="idl">
+  [Exposed=Window, SecureContext]
+  interface IdentityCredentialError : DOMException {
+    constructor(optional DOMString message = "", optional IdentityCredentialErrorInit options = {});
+    readonly attribute DOMString code;
```
bvandersloot-mozilla, line 576: This overwrites a deprecated value that should be `0`. Another name would be good.

npm1: Ah, good catch! We missed that... here are some other naming options, any preference?

- errorCode
- serverError{Code} (eg with or without Code)
- idpError{Code}
- credentialError{Code}

npm1: - error

samuelgoto:
@npm1 just checking: is there anything else we need to act on before we merge this?

npm1:
> @npm1 just checking: is there anything else we need to act on before we merge this?

Yes, I need to rebase and change the `code` to whatever we agree on (awaiting @bvandersloot-mozilla's confirmation about whether `error` is acceptable)

spec/index.bs
npm1: Ping, I'd like to merge this sooner rather than later especially if we want to make a naming change here.

bvandersloot-mozilla: error is good. credentialErrorCode may be better

bvandersloot-mozilla (changes requested):
I don't think we need to make this web-observable and doing so risks leaking information to the RP about shared browser state.

npm1:
> I don't think we need to make this web-observable and doing so risks leaking information to the RP about shared browser state.

What is 'this'? And can you elaborate on the leak you are thinking about?

bvandersloot-mozilla:
"this" is the `IdentityCredentialError` itself, particularly the `code` variable which contains an unspecified string (presumably one of "invalid_request", "unauthorized_client", "access_denied", "server_error", and "temporarily_unavailable"). If I understand this patch correctly, that ends up returned from `navigator.credentials.get` and reveals internal information about the credential discovery to the RP.

bvandersloot-mozilla:
Ah, I didn't realize this is happening on the id assertion endpoint only. In that case, this is reasonable, but we should probably enumerate the possible values for `code` and specify when they are returned.

npm1:
> Ah, I didn't realize this is happening on the id assertion endpoint only. In that case, this is reasonable, but we should probably enumerate the possible values for `code` and specify when they are returned.

I was hoping to leave the values for `error` (previously `code`) unspecified similar to how we do not specify the format/return of `token`. That way the IdP can send the information needed for the RP to process the error, and the user agent may choose to show generic UI or show more specific UI if it chooses to do so based on the error string. What do you think? I did mention the strings Chrome uses to show more specific UI in case it is of interest for an implementor to match.

spec/index.bs
```diff
@@ -851,22 +894,24 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
                 However, UAs may have different UI approaches here and prevent it in a different way.
         1. [=Queue a global task=] on the [=DOM manipulation task source=]
             to throw a new "{{NetworkError}}" {{DOMException}}.
-    1. Otherwise, return |credential|.
+    1. Otherwise, return |credential| (this may throw since it may be an
```
TallTed, line 897:
```suggestion
    1. Otherwise, return |credential| (this may throw, since it may be an
```

spec/index.bs
```diff
@@ -975,7 +1020,7 @@ the exception thrown.
             the user agent MAY show some UI to the user indicating that they are being
             <dfn>auto-reauthenticated</dfn>.
         1. Set |isAutoSelected| to true.
-    1. Otherwise, if |mediation| is "{{CredentialMediationRequirement/silent}}", return (failure, true).
+    1. Otherwise, if |mediation| is "{{CredentialMediationRequirement/silent}}", return (failure, false).
     1. Otherwise, if |accountsList|'s size is 1:
```
TallTed, line 1024: Two "otherwise" at the same level is confusing.

spec/index.bs
```diff
@@ -1005,6 +1050,21 @@ the exception thrown.
...
+            "invalid_request", "unauthorized_client", "access_denied", "server_error", and
+            "temporarily_unavailable", the [=user agent=] may note the reason in a user-friendly
+            manner. See [here](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1) for
+            more details about these.
+        1. The [=user agent=] MAY use the {{IdentityCredentialError/url}} to display a hyperlink
+            in its UI so the user can click on it to learn more information about the error message.
+        1. Wait until the UI is acknowledged or dismissed by the user.
+        
+        Note: this wait means the [=user agent=] rejects the call only after the UI has been
+            closed by the user.
```
TallTed, line 1067: Nest the note about the wait in the step with the wait, not in its superior.
```suggestion
        1. Wait until the UI is acknowledged or dismissed by the user.  
           Note: this wait means the [=user agent=] rejects the call only after the UI has been
           closed by the user.
```

spec/index.bs
```diff
@@ -1341,14 +1406,31 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
+                {{IdentityCredentialError}} given |globalObject|'s [=global object/realm=]
+                and "IdentityCredentialError", and return.
+            1. Let |credential| be a new {{IdentityCredential}} given |globalObject|'s
+                [=global object/realm=].
+            1. Set |credential|'s {{IdentityCredential/token}} to |token|.
+            1. Set |credential|'s {{IdentityCredential/isAutoSelected}} to |isAutoSelected|.
+        1. Otherwise, i.e. if |json|[|error|] [=map/exists=]:
+            1. If |json|[|error|] is not an [=ordered map=], set |credential| to a new
+                {{IdentityCredentialError}} given |globalObject|'s [=global object/realm=] and
+                "IdentityCredentialError" and return.
```
TallTed, line 1424:
```suggestion
                "IdentityCredentialError", and return.
```

TallTed (changes requested):
Small tweaks for clarity

spec/index.bs
npm1: Changed to else if, which is arguably less confusing.

TallTed: I think this will need revisiting. I don't think the global replace of "otherwise" with "else" resolves my concern, since there are still two of them at the same nesting level.

I think the if/then/else nesting should be more carefully considered: definitely with some nest demotions; probably with some nest promotions; and possibly making some `ol` into `ul` and/or vice versa.

npm1: I have updated the PR to use a switch style, let me know if you think that is clearer. That said, note that having conditional statements like this is accepted, as this is an algorithm, not pure prose. https://infra.spec.whatwg.org/#algorithm-conditional-statements

TallTed: Ugh. I see that [one of the examples in the linked section](https://infra.spec.whatwg.org/#algorithm-conditional-statements) looks like what was being done here. I maintain that it is confusing, and has strong potential to lead to problems in implementation and interop, which seem worth avoiding.

Happily, the rendered "switch" style seems likely to reduce if not eliminate those potential problems.

npm1:
Rebased this PR in case people want to take another look ahead of TPAC discussion

spec/index.bs
```diff
@@ -648,6 +648,8 @@ Given a |realm|, a {{DOMString}} |message|, and an {{IdentityCredentialErrorInit
     :   <b>{{IdentityCredentialError/error}}</b>
     ::  The {{IdentityCredentialError/error}}'s attribute getter returns the value it is set to.
         It represents the type of error which resulted in an {{IdentityCredential}} not being created.
+        An [=IDP=] MUST not expose sensitive user information in this field, since it is exposed to
```
```suggestion
        An [=IDP=] MUST NOT expose sensitive user information in this field, since it is exposed to
```

spec/index.bs
```diff
@@ -611,7 +611,52 @@ dictionary DisconnectedAccount {
...
+        if it is present in |options|, or to "" otherwise.
+    1. Set <a>this</a>.{{IdentityCredentialError/url}} to |options|.{{IdentityCredentialError/url}}
+        if it is present in |options|, or to "" otherwise.
+</div>
+
+<dl>
+    :   <b>{{IdentityCredentialError/error}}</b>
+    ::  The {{IdentityCredentialError/error}}'s attribute getter returns the value it is set to.
+        It represents the type of error which resulted in an {{IdentityCredential}} not being created.
+        An [=IDP=] MUST not expose sensitive user information in this field, since it is exposed to
```
cbiesinger:
```suggestion
        An [=IDP=] MUST NOT expose sensitive user information in this field, since it is exposed to
```

spec/index.bs
```diff
@@ -765,10 +810,10 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
         algorithm. See <a href="https://github.com/w3c/webappsec-credential-management/issues/210">issue</a>.
 
     1. If |credential| is a pair:
-        1. Let |throwImmediately| be the value of the second element of the pair.
+        1. Let |delayException| be the value of the second element of the pair.
```
cbiesinger: Why did you invert the meaning of this parameter? It doesn't look like you changed the callers?

spec/index.bs
```diff
@@ -987,78 +989,21 @@ throwing the exception.
...
+                1. Let |result| be the result of that algorithm.
+                1. If |result| is failure, return (failure, true). The user
+                    agent MAY show a dialog to the user before or after
+                    returning failure indicating this failure.
+                1. Otherwise, go back to the [=fetch accounts step=] to get an updated
+                    value of |providerMap| for this [=IDP=].
+        1. Otherwise, |value| is a [=list=] of accounts. [=list/Extend=] |allAccounts| with |value|.
+    1. Also include a UI affordance to close the dialog. If the user closes this dialog, return (failure,
+        true).
+    1. <dfn for="create identity credential">Show accounts</dfn> step: if |allAccounts| is not [=list/empty=], also add UI to present the account options to the user.
```
```suggestion
    1. <dfn for="create identity credential">Show accounts</dfn> step: if |allAccounts|
       is not [=list/empty=], also add UI to present the account options to the user.
```

spec/index.bs
```diff
@@ -1462,31 +1427,45 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
         set to the following steps given a <a spec=fetch for=/>response</a> |response| and |responseBody|:
         1. Let |json| be the result of [=extract the JSON fetch response=] from |response| and
             |responseBody|.
-        1. If |json|[|token|] and |json|[|error|] both [=map/exist=] or both do not [=map/exist=],
-            set |credential| to a new {{IdentityCredentialError}} given |globalObject|'s
+        1. [=converted to an IDL value|Convert=] |json| to an {{IdentityAssertionResponse}}, |token|.
+        1. If one of the previous two steps threw an exception, set |credential| to a new {{IdentityCredentialError}} given |globalObject|'s
```
```suggestion
        1. If one of the previous two steps threw an exception, set |credential| to a new
           {{IdentityCredentialError}} given |globalObject|'s
```

TallTed (changes requested):

spec/index.bs
```diff
@@ -962,6 +1009,21 @@ the exception thrown.
     1. Let |credential| be the result of running the [=fetch an identity assertion=] algorithm with
         |selectedAccount|'s {{IdentityProviderAccount/id}}, |permissionRequested|, |isAutoSelected|,
         |provider|, |config|, and |globalObject|.
+    1. If |credential| is an {{IdentityCredentialError}}:
+        1. The [=user agent=] MUST show UI to the user indicating that the identity assertion fetch
+            has failed.
+        1. The [=user agent=] MAY use the {{IdentityCredentialError/error}} in its UI to provide a
+            more specific error message to the user. For instance, if the value is one of
+            "invalid_request", "unauthorized_client", "access_denied", "server_error", and
+            "temporarily_unavailable", the [=user agent=] may note the reason in a user-friendly
```
cbiesinger, line 1032: hmm I don't love the way this is specified. Have you considered something like `The user agent SHOULD support displaying a user-readable message for "invalid_request", "unauthorized_client", "access_denied", "server_error", and "temporarily_unavailable"`?

spec/index.bs
```diff
@@ -611,7 +611,52 @@ dictionary DisconnectedAccount {
...
+  [Exposed=Window, SecureContext]
+  interface IdentityCredentialError : DOMException {
+    constructor(optional DOMString message = "", optional IdentityCredentialErrorInit options = {});
+    readonly attribute DOMString error;
+    readonly attribute DOMString url;
+  };
+</pre>
+
+<div algorithm="IdentityCredentialError constructor">
+Given a |realm|, a {{DOMString}} |message|, and an {{IdentityCredentialErrorInit}} |options|, the
```
cbiesinger, line 651: This PR assumes that |options| is optional, but this algorithm does not say so, it probably should? Alternatively the callers should specify an empty dict.

spec/index.bs
```diff
@@ -1308,16 +1370,21 @@ To <dfn>fetch the account picture</dfn> given an {{IdentityProviderAccount}} |ac
...
+    and |globalObject|, run the following steps. This returns an {{IdentityCredential}} or an
+    {{IdentityCredentialError}}.
     1. Let |tokenUrl| be the result of [=computing the manifest URL=] given |provider|,
-        |config|["{{IdentityProviderAPIConfig/id_assertion_endpoint}}"], and |globalObject|.
-    1. If |tokenUrl| is failure, return failure.
+        |config|["{{IdentityProviderAPIConfig/id_assertion_endpoint}}"], true, and |globalObject|.
+    1. If |tokenUrl| is failure:
+        1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to
+            let |error| be a new {{IdentityCredentialError}} given |globalObject|'s
+            [=global object/realm=].
```
cbiesinger: this needs to specify the message argument

spec/index.bs
```diff
@@ -1593,7 +1669,10 @@ When <dfn>computing the manifest URL</dfn> given an {{IdentityProviderRequestOpt
         allowed.
 
     1. If |manifestUrl| is failure, return failure.
-    1. If |manifestUrl| is not [=same origin=] with |configUrl|, return failure.
+    1. If |requireSameOrigin| and |manifestUrl| is not [=same origin=] with |configUrl|, return
+        failure.
+    1. If |requireSameOrigin| is false and |manifestUrl|'s [=url/host=]'s [=host/registrable domain=]
```
cbiesinger: seems like it should be called requireSameSite, not requireSameOrigin

spec/index.bs
```diff
@@ -1361,14 +1428,20 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
-            and return.
-        1. If neither {{IdentityAssertionResponse/token}} nor
-            {{IdentityAssertionResponse/continue_on}} was specified, set |credential| to failure
-            and return.
-        1. If {{IdentityAssertionResponse/token}} was specified, let |tokenString|
+        1. If one of the previous two steps threw an exception, set |credential| to a new {{IdentityCredentialError}} given |globalObject|'s
+            [=global object/realm=] and "IdentityCredentialError", and return.
+        1. If {{IdentityAssertionResponse/error}} was specified:
+            1. If |errorInit|.{{IdentityCredentialErrorInit/url}} is present:
+                1. Let |errorUrl| be the result of [=computing the manifest URL=] given |provider|,
```
cbiesinger: this does not sound right. shouldn't the error URL be relative to the id assertion endpoint, not the config URL?

spec/index.bs
npm1: The fact they are optional is in the idl. When invoked, default value is passed if needed. Example: https://webidl.spec.whatwg.org/#dom-domexception-domexception

spec/index.bs
npm1: Ah this used to be a bug in the spec but it is fixed so undoing this

spec/index.bs
npm1: That would make it normatively required to support these. These are arbitrary enough that I do not want to make this required. I'm only listing these so other user agents know the ones we currently show custom messages for.

spec/index.bs
npm1: Oh looks like that is how we implemented. That is surprising to me given all other urls were relative to the config url.

spec/index.bs
npm1: ? No, it was same origin, we only check same site when it was false. But I removed this

spec/index.bs
npm1: It doesnt?

npm1: Hmm actually maybe it does, the fact they are optional in the IDL might not mean they are optional in here. Not sure

spec/index.bs
cbiesinger: That's why I said SHOULD and not MUST. or even MAY...

spec/index.bs
cbiesinger: Even if it really is optional, which I am also unsure about, why are you specifying "IdentityCredentialError" everywhere else but using a default "" here?

spec/index.bs
cbiesinger: Usually relative URLs are supposed to be relative to the file they are in, in this case that's the ID assertion endpoint.

The other URLs are in the config file so they are relative to that.

Hmm, I see that we don't support relative URLs for continue_on in the spec :(

spec/index.bs
```diff
@@ -611,7 +611,52 @@ dictionary DisconnectedAccount {
...
 <!-- ============================================================ -->
-### The CredentialRequestOptions ### {#browser-api-credential-request-options}
+## The IdentityCredentialError Interface ## {#browser-api-identity-credential-error-interface}
+<!-- ============================================================ -->
+
+This specification introduces a new type of {{DOMException}}, the {{IdentityCredentialError}}. It
+is used when the [=user agent=] does not receive an identity assertion after the user has requested
+to use a federated account.
+
+<xmp class="idl">
```
cbiesinger: why is this using xmp but the interface right below is using pre?

spec/index.bs
```diff
@@ -783,22 +828,24 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
 
 <div algorithm="create an IdentityCredential">
 To <dfn>create an IdentityCredential</dfn> given a [=sequence=] of  {{IdentityProviderRequestOptions}}
 |providerList|, a {{CredentialRequestOptions}} |options|, and a
-|globalObject|, run the following steps. This returns an {{IdentityCredential}}
-or a pair (failure, bool), where the bool indicates whether to skip delaying
-the exception thrown.
+|globalObject|, run the following steps. This returns an {{IdentityCredential}}, an
+{{IdentityCredentialError}}, or a pair (failure, bool), where the bool indicates whether to delay
+throwing the exception.
```
cbiesinger: is the `delay throwing` change correct? (even if it is, seems unrelated?)

spec/index.bs
```diff
@@ -1361,14 +1429,27 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
+                1. If |errorUrl|'s [=url/host=]'s [=host/registrable domain=]
+                    is not equal to |tokenUrl|'s, set |errorUrl| to failure.
+                1. If |errorUrl| is not a [=potentially trustworthy URL=], set set |errorUrl| to
+                    failure.
+                1. If |errorUrl| is failure, set |errorInit|.{{IdentityCredentialErrorInit/url}}
+                    to "". Otherwise, set |errorInit|.{{IdentityCredentialErrorInit/url}} to
+                    |errorUrl|.
+            1. Let |credential| be a new {{IdentityCredentialError}} given |globalObject|'s
+                [=global object/realm=], "IdentityCredentialError", and |errorInit|.         
+        1. Otherwise, if {{IdentityAssertionResponse/token}} was specified, let |tokenString|
```
cbiesinger, line 1471: hmm we should mention somewhere that this should only happen for successful HTTP responses (200 OK).... though this is a pre-existing bug

cbiesinger (approved):

spec/index.bs
npm1: Agh, nope

spec/index.bs
npm1: Ok switched to should.

spec/index.bs
npm1: I'm told I should not use the constructor so used a create method instead

spec/index.bs
npm1: Fixed I think. Side note: I think our implementation validates the endpoints at the beginning, not at this point in time. But this is unrelated to this PR so will fix later

spec/index.bs
cbiesinger: this does not seem resolved?

spec/index.bs
```diff
@@ -783,22 +841,24 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
 ### Create an IdentityCredential ### {#create-identity-credential}
 <!-- ============================================================ -->
 
 The <a>create an IdentityCredential</a> algorithm invokes the various FedCM fetches, shows the user
-agent UI, and creates the {{IdentityCredential}} that is then returned to the [=RP=].
+agent UI, and creates the {{IdentityCredential}} or {{IdentityCredentialError}} which is then
+returned to the [=RP=].
 
 <div algorithm="create an IdentityCredential">
-To <dfn>create an IdentityCredential</dfn> given a [=sequence=] of  {{IdentityProviderRequestOptions}}
```
cbiesinger, line 797: Why are you removing the "sequence" part? Does that not break multi-IDP?

cbiesinger (approved):

spec/index.bs
npm1: Bad rebase...

spec/index.bs
npm1: I mean a response is only received in case of a successful fetch. Not sure what you mean?

spec/index.bs
npm1: Sorry I had started a change and resolved but lost my progress midway. Done now

spec/index.bs
cbiesinger: don't you get a response also for 404? I guess I'm not certain how the fetch API handles that

spec/index.bs
```diff
@@ -783,22 +841,23 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
...
-|providerList|, a {{CredentialRequestOptions}} |options|, and a
-|globalObject|, run the following steps. This returns an {{IdentityCredential}}
-or a pair (failure, bool), where the bool indicates whether to skip delaying
-the exception thrown.
+To <dfn>create an IdentityCredential</dfn> given a [=sequence=] of
+{{IdentityProviderRequestOptions}} |providerList|, a {{CredentialRequestOptions}} |options|, and a
+|globalObject|, run the following steps. This returns an {{IdentityCredential}} or a pair (failure,
+bool), where the bool indicates whether to throw the exception immediately.
```
cbiesinger: As far as I can tell, the new text is identical to the old text except for line wrapping. But it should probably say that it can also return an IdentityCredentialError?

cbiesinger (approved):

spec/index.bs
npm1: `responseBody` will only be converted to JSON without throwing if it fetched. Added a step to make it clearer

cbiesinger: But 404 responses do have a response body.

Maybe more importantly, we do want to process "error" in 404 responses as well

npm1: Ah yea. Hmm well it seems we DO allow http errors in our implementation, so we do parse the body in those cases, for ID assertion in particular. Fixed I think?

spec/index.bs
```diff
@@ -1461,6 +1461,8 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
                     |errorUrl|.
             1. Set |credential| to the result of [=create an IdentityCredentialError=] with
                 |errorInit|, and |globalObject|.
+        1. Otherwise, if |response|'s [=response/status=] is an [=ok status=], set |credential| to
```
cbiesinger: is this missing a "not"?

npm1: Oops yea

cbiesinger (approved):
lgtm. we should probably add status checks to our other fetches...

spec/index.bs
```diff
@@ -1358,17 +1440,33 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
+                1. If |errorUrl|'s [=url/host=]'s [=host/registrable domain=]
+                    is not equal to |tokenUrl|'s, set |errorUrl| to failure.
+                1. If |errorUrl| is not a [=potentially trustworthy URL=], set set |errorUrl| to
+                    failure.
+                1. If |errorUrl| is failure, set |errorInit|.{{IdentityCredentialErrorInit/url}}
+                    to "". Otherwise, set |errorInit|.{{IdentityCredentialErrorInit/url}} to
+                    |errorUrl|.
+            1. Set |credential| to the result of [=create an IdentityCredentialError=] with
+                |errorInit|, and |globalObject|.
+        1. Otherwise, if |response|'s [=response/status=] is not an [=ok status=], set |credential|
```
yi-gu: In Chrome we may update the error even if the IdP didn't specify an `error` in the response. e.g. we try to map 500/503 to the enum values to give users more context: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/webid/idp_network_request_manager.cc;l=895?q=content%2Fbrowser%2Fwebid%2Fidp_network_request_manager.cc

spec/index.bs
```diff
@@ -2234,9 +2335,11 @@ Every {{IdentityAssertionResponse}} is expected to have members with the followi
     ::  The resulting token.
     :   <dfn>continue_on</dfn>
     ::  A URL that the user agent will open in a popup to finish the authentication process.
+    :   <dfn>error</dfn>
+    ::  A dictionary containing the error that occurred when the ID assertion was fetched.
 </dl>
 
-Only one of `token` and `continue_on` should be specified.
+Only one of `token`, `continue_on`, and `error` should be specified.
```
yi-gu: I know it's specified in the algorithm above, but do you think it's worth calling out which will be respected if multiple are specified?

spec/index.bs
```diff
@@ -1461,8 +1461,13 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
                 |errorInit|, and |globalObject|.
-        1. Otherwise, if |response|'s [=response/status=] is not an [=ok status=], set |credential|
-            to the result of [=create an IdentityCredentialError=] with {} and |globalObject|.
+        1. Otherwise, if |response|'s [=response/status=] is not an [=ok status=]:
+            1. Set |credential| to the result of [=create an IdentityCredentialError=] with {} and
+                |globalObject|.
+            1. The user agent MAY set |credential|'s {{IdentityCredentialError/error}} based on
+                |response|'s [=response/status=]. For example, it could set it to "server_error" if
+                the [=response/status=] is 500 and it could set it to "temporarily_unavailable" if
+                the [=response/status=] is 503.
```
```suggestion
            1. The user agent MAY set |credential|'s {{IdentityCredentialError/error}}
               based on |response|'s [=response/status=]. For example, if the
               [=response/status=] is `500`, it could be set to "server_error",
               and if the [=response/status=] is `503`, it could be set to
               "temporarily_unavailable".
```

spec/index.bs
```diff
@@ -2339,7 +2344,9 @@ Every {{IdentityAssertionResponse}} is expected to have members with the followi
     ::  A dictionary containing the error that occurred when the ID assertion was fetched.
 </dl>
 
-Only one of `token`, `continue_on`, and `error` should be specified.
+Only one of `token`, `continue_on`, and `error` should be specified. When multiple are specified,
```
```suggestion
Only one of `token`, `continue_on`, or `error` should be specified. When multiple are specified,
```

spec/index.bs
```diff
@@ -2339,7 +2344,9 @@ Every {{IdentityAssertionResponse}} is expected to have members with the followi
...
+the order of processing is [`error`, `token`, `continue_on`], so the first present will be used in
+the response.
```
```suggestion
the order of processing is [`error`, `token`, `continue_on`], so the first encountered
will be used in the response.
```

TallTed (changes requested):

bvandersloot-mozilla (approved):

hlflanagan:
Discussed in [FedID CG/WG call, 24 June 2025](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-06-24-FedCM-notes.md)

PR #507, opened 2023-10-20
Change various "sign in" to "login"

cbiesinger:
Most notably, the signin_url field is renamed to login_url to match the spec.

### DIFF
--- a/proposals/idp-sign-in-status-api.md
+++ b/proposals/idp-sign-in-status-api.md
@@ -1,4 +1,4 @@
-# Explainer for IDP sign-in status API
+# Explainer for IDP login status API (née IDP sign-in status API)
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


proposals/idp-sign-in-status-api.md
```diff
@@ -1,4 +1,4 @@
-# Explainer for IDP sign-in status API
+# Explainer for IDP login status API (née IDP sign-in status API)
```
samuelgoto, line 1:
```suggestion
# Explainer for IDP Login Status API (née IDP Sign-in Status API)
```

samuelgoto (approved):

samuelgoto:
LGTM

Do we still actually need this explainer now that the spec is patched?

cbiesinger:
I think the explainer is still useful as a more concise description of the feature. (Note how the blink launch process needs explainers even when the spec PR has landed...)

Anyway, your suggestion is committed, so this is ready for merging.

proposals/idp-sign-in-status-api.md
```diff
@@ -1,4 +1,4 @@
-# Explainer for IDP sign-in status API
+# Explainer for IDP Login Status API (née IDP Sign-in Status API)
```
npm1, line 1: née?

cbiesinger: meaning 2 in https://www.merriam-webster.com/dictionary/n%C3%A9e

Should I use "fka" instead? I didn't want to use an abbreviation...

npm1: Oh ok I didn't know that word. I suppose if you think it won't be confusing...

TallTed: `Née` will often, but not always, be understood.

Spelled out `formerly known as` (rather than `fka`) might be better.

cbiesinger: Alright, see https://github.com/fedidcg/FedCM/pull/508

PR #512, opened 2023-11-09
Add domainHint to the spec

npm1:
Fixes https://github.com/fedidcg/FedCM/issues/427. Adds a `domainHint` parameter that can be used in FedCM to filter accounts based on the corporate domains the belong to. This behaves similarly to login hint, except with the additional behavior that "*" matches any domain hint. In addition, the `domainHint` and `loginHint` are passed as query parameters in the IDP login URL.

***
<a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/512.html" title="Last updated on Nov 29, 2023, 8:29 PM UTC (f1252ad)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/512/75a6d4a...f1252ad.html" title="Last updated on Nov 29, 2023, 8:29 PM UTC (f1252ad)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -145,7 +145,6 @@ could be implemented.
             configURL: "https://idp.example/manifest.json",
             clientId: "123",
             nonce: nonce,
-            domainHint: "any"
           }]
         }
       });
@@ -775,8 +774,9 @@ the exception thrown.
         1. If |accountList| is now empty, go to the [=mismatch dialog step=].
     1. If |provider|'s {{IdentityProviderConfig/domainHint}} is not empty:
         1. For every |account| in |accountList|:
-            1. If {{IdentityProviderConfig/domainHint}} is "any", remove |account| from
-                |accountList| if |account|'s {{IdentityProviderAccount/domain_hints}}  is not empty.
+            1. If {{IdentityProviderConfig/domainHint}} is "any":
+                1. If |account|'s {{IdentityProviderAccount/domain_hints}} is empty, remove
+                    |account| from |accountList|.
             1. Otherwise, remove |account| from |accountList| if |account|'s
                 {{IdentityProviderAccount/domain_hints}} does not [=list/contain=] |provider|'s
                 {{IdentityProviderConfig/domainHint}}.
@@ -1345,8 +1345,8 @@ success or failure.
         |loginUrl| to the result of running [=url parser=] with
         |config|.{{IdentityProviderAPIConfig/login_url}}.
     1. Wait until |loginUrl| is not null.
-    1. Assert: |loginUrl| is not failure (the [=user agent=] has checked
-        |config|.{{IdentityProviderAPIConfig/login_url}} to be a valid URL previously).
+    1. Assert: |loginUrl| is not failure (the [=user agent=] has previously checked that
+        |config|.{{IdentityProviderAPIConfig/login_url}} is a valid URL).
     1. Let |queryList| be a new [=list=].
     1. If |provider|'s {{IdentityProviderConfig/loginHint}} is not empty, [=list/append=]
         ("login_hint", {{IdentityProviderConfig/loginHint}}) to |queryList|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1345,7 +1345,7 @@ success or failure.
         ("login_hint", {{IdentityProviderConfig/loginHint}}) to |queryList|.
     1. If |provider|'s {{IdentityProviderConfig/domainHint}} is not empty, [=list/append=]
         ("domain_hint", {{IdentityProviderConfig/domainHint}}) to |queryList|.
-    1. If |queryList| is not [=list/empty=], let |queryParameters| be the result of the [=urlencoded serializer=] with |queryList|. Append |queryParameters| to |urlString|.
+    1. If |queryList| is not [=list/empty=], let |queryParameters| be the result of the [=urlencoded serializer=] with |queryList|. Append "?" and |queryParameters| to |urlString|.
     1. Let |loginUrl| be null.
     1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to set
         |loginUrl| to the result of running [=url parser=] with |urlString|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -574,6 +574,7 @@ dictionary IdentityProviderConfig {
   required USVString clientId;
   USVString nonce;
   DOMString loginHint;
+  DOMString domainHint;
 };
 </xmp>
 
@@ -591,6 +592,11 @@ dictionary IdentityProviderConfig {
         agent to show to the user. If provided, the user agent will not show accounts which do not
         match this login hint value. It generally matches some attribute from the desired
         {{IdentityProviderAccount}}.
+    :   <b>{{IdentityProviderConfig/domainHint}}</b>
+    ::  A string representing the domain hint corresponding to a domain which the [=RP=] is
+        interested in, or "*" if the [=RP=] wants any account associated with at least one domain
+        hint. If provided, the user agent will not show accounts which do not match the domain hint
+        value.
 </dl>
 
 <!-- ============================================================ -->
@@ -705,7 +711,7 @@ the exception thrown.
                 1. Let |config| be the result of running [=fetch the config file=]
                     with |provider| and |globalObject|.
                 1. If |config| is failure, return (failure, true).
-                1. [=Show an IDP login dialog=] with |config|.
+                1. [=Show an IDP login dialog=] with |config| and |provider|.
                 1. If that algorithm returns failure, return (failure, true).
 
         Issue: We should perhaps provide a way to let the [=RP=] request that
@@ -735,8 +741,8 @@ the exception thrown.
         1. <dfn>Mismatch dialog step</dfn>: If |loginStatus| is [=logged-in=], show a
             dialog to the user. The contents of this dialog are defined by the user
             agent. This dialog SHOULD provide an affordance for the user to trigger
-            the [=show an IDP login dialog=] algorithm with |config|; this dialog
-            is the <dfn>confirm IDP login dialog</dfn>.
+            the [=show an IDP login dialog=] algorithm with |config| and |provider|;
+            this dialog is the <dfn>confirm IDP login dialog</dfn>.
 
             Note: This situation happens when the browser expects the user
                 to be signed in, but the accounts fetch indicated that the user
@@ -765,6 +771,14 @@ the exception thrown.
         1. For every |account| in |accountList|, remove |account| from |accountList| if |account|'s
             {{IdentityProviderAccount/login_hints}} does not [=list/contain=] |provider|'s
             {{IdentityProviderConfig/loginHint}}.
+        1. If |accountList| is now empty, go to the [=mismatch dialog step=].
+    1. If |provider|'s {{IdentityProviderConfig/domainHint}} is not empty:
+        1. For every |account| in |accountList|:
+            1. If {{IdentityProviderConfig/domainHint}} is "*", remove |account| from |accountList|
+                if |account|'s {{IdentityProviderAccount/domain_hints}}  is not empty.
+            1. Otherwise, remove |account| from |accountList| if |account|'s
+                {{IdentityProviderAccount/domain_hints}} does not [=list/contain=] |provider|'s
+                {{IdentityProviderConfig/domainHint}}.
         1. If |accountList| is now empty, go to the [=mismatch dialog step=].
     1. For each |acc| in |accountsList|:
         1. If |acc|["{{IdentityProviderAccount/picture}}"] is present, [=fetch the account picture=]
@@ -1039,6 +1053,7 @@ dictionary IdentityProviderAccount {
   USVString picture;
   sequence<USVString> approved_clients;
   sequence<DOMString> login_hints;
+  sequence<DOMString> domain_hints;
 };
 dictionary IdentityProviderAccountList {
   sequence<IdentityProviderAccount> accounts;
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1316,10 +1331,24 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+{{IdentityProviderConfig}} |provider|, and a |globalObject|, run the following steps. This returns
+success or failure.
+    1. Assert: these steps are running [=in parallel=].
+    1. Let |urlString| be |config|.{{IdentityProviderAPIConfig/login_url}}.
+    1. Let |queryList| be a new [=list=].
+    1. If |provider|'s {{IdentityProviderConfig/loginHint}} is not empty, [=list/append=]
+        ("login_hint", {{IdentityProviderConfig/loginHint}}) to |queryList|.
+    1. If |provider|'s {{IdentityProviderConfig/domainHint}} is not empty, [=list/append=]
+        ("domain_hint", {{IdentityProviderConfig/domainHint}}) to |queryList|.
+    1. If |queryList| is not [=list/empty=], let |queryParameters| be the result of the [=urlencoded serializer=] with |queryList|. Append "?" and |queryParameters| to |urlString|.
```
cbiesinger, line 1344: please linebreak this

spec/index.bs
```diff
@@ -1316,10 +1331,24 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+    1. Let |loginUrl| be null.
+    1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to set
```
cbiesinger, line 1346: This is only done for the upcoming assert? it doesn't otherwise seem needed since you already appended the query string to urlString?

spec/index.bs
cbiesinger, line 1344: Should we somehow handle the case where urlString already contained a query string?

spec/index.bs
```diff
@@ -1693,15 +1726,17 @@ For example:
    "email": "john_doe@idp.example",
    "picture": "https://idp.example/profile/123",
    "approved_clients": ["123", "456", "789"],
-   "login_hints": ["john_doe"]
+   "login_hints": ["john_doe"],
+   "domain_hints": ["idp.example"],
```
cbiesinger, line 1730: perhaps use "*" in one of the examples

spec/index.bs
TallTed:
```suggestion
    1. If |queryList| is not [=list/empty=], let |queryParameters| be 
       the result of the [=urlencoded serializer=] with |queryList|. 
       Append "?" and |queryParameters| to |urlString|.
```

npm1: Hmm yea... I could not find a nice easy way. PTAL

spec/index.bs
npm1: I was told in the past that url parser can only run in a task. Not sure if that has changed, but if/when it does we should change all url parser calls to be invoked from in parallel.

spec/index.bs
npm1: This is the accounts, so not the right place. I added it in a JS example

spec/index.bs
cbiesinger: Oh sorry, I had misread this algorithm

cbiesinger (approved):
lgtm

spec/index.bs
cbiesinger: Looks good

spec/index.bs
```diff
@@ -762,6 +769,14 @@ the exception thrown.
             {{IdentityProviderAccount/login_hints}} does not [=list/contain=] |provider|'s
             {{IdentityProviderConfig/loginHint}}.
         1. If |accountList| is now empty, go to the [=mismatch dialog step=].
+    1. If |provider|'s {{IdentityProviderConfig/domainHint}} is not empty:
+        1. For every |account| in |accountList|:
+            1. If {{IdentityProviderConfig/domainHint}} is "any", remove |account| from
+                |accountList| if |account|'s {{IdentityProviderAccount/domain_hints}}  is not empty.
```
samuelgoto, line 775:
```suggestion
                |accountList| if |account|'s {{IdentityProviderAccount/domain_hints}}  is empty.
```

When "any" is passed, aren't you suppose to remove the accounts that have **empty** domain_hints, as opposed to **non-empty**?

That is, if "any" is passed, then:

domain_hint = [] <- should be removed
domain_hint = ["foobar"] <- should not be removed

right?

Either way, there are too many double-negatives here, can you try to simplify this statement?

spec/index.bs
```diff
@@ -144,7 +144,8 @@ could be implemented.
           providers: [{
             configURL: "https://idp.example/manifest.json",
             clientId: "123",
-            nonce: nonce
+            nonce: nonce,
+            domainHint: "any"
```
samuelgoto, line 148:
```suggestion
```

domainHint is a pretty narrow/specific use case, I'd avoid using it in the main example.

spec/index.bs
```diff
@@ -762,6 +769,14 @@ the exception thrown.
             {{IdentityProviderAccount/login_hints}} does not [=list/contain=] |provider|'s
             {{IdentityProviderConfig/loginHint}}.
         1. If |accountList| is now empty, go to the [=mismatch dialog step=].
+    1. If |provider|'s {{IdentityProviderConfig/domainHint}} is not empty:
+        1. For every |account| in |accountList|:
+            1. If {{IdentityProviderConfig/domainHint}} is "any", remove |account| from
+                |accountList| if |account|'s {{IdentityProviderAccount/domain_hints}}  is not empty.
+            1. Otherwise, remove |account| from |accountList| if |account|'s
+                {{IdentityProviderAccount/domain_hints}} does not [=list/contain=] |provider|'s
+                {{IdentityProviderConfig/domainHint}}.
```
TallTed, line 778: This includes the [fix suggested](https://github.com/fedidcg/FedCM/pull/512/files#r1409658311) by @samuelgoto.

It is not clear to me what "otherwise" refers to, so it is not clear how to rephrase it explicitly (which would be better). Here's my best guess:
```suggestion
            1. If {{IdentityProviderConfig/domainHint}} is "any" and |account|'s
               {{IdentityProviderAccount/domain_hints}} is not empty, remove
               |account| from |accountList|.
            1. If {{IdentityProviderConfig/domainHint}} is not "any" and
               |account|'s {{IdentityProviderAccount/domain_hints}} does not
               [=list/contain=] |provider|'s {{IdentityProviderConfig/domainHint}},
               remove |account| from |accountList|.
```

spec/index.bs
```diff
@@ -1316,10 +1332,27 @@ and a |responseBody|, run the following steps. This returns an [=ordered map=].
...
+{{IdentityProviderConfig}} |provider|, and a |globalObject|, run the following steps. This returns
+success or failure.
+    1. Assert: these steps are running [=in parallel=].
+    1. Let |loginUrl| be null.
+    1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to set
+        |loginUrl| to the result of running [=url parser=] with
         |config|.{{IdentityProviderAPIConfig/login_url}}.
+    1. Wait until |loginUrl| is not null.
+    1. Assert: |loginUrl| is not failure (the [=user agent=] has checked
+        |config|.{{IdentityProviderAPIConfig/login_url}} to be a valid URL previously).
```
TallTed, line 1345:
```suggestion
    1. Assert: |loginUrl| is not failure (the [=user agent=] has previously checked
        that |config|.{{IdentityProviderAPIConfig/login_url}} is a valid URL).
```

spec/index.bs
npm1: Yea needs to be removed if empty. Tried to simplify

spec/index.bs
npm1: I split the previous step into two so hopefully the "otherwise" is clear now?

samuelgoto (approved):

samuelgoto:
LGTM

@bvandersloot-mozilla WDYT?

bvandersloot-mozilla:
It will be confusing for a newcomer to distinguish domain_hint and login_hint and which they want/need. 

Thinking a bit more, couldn't we just make the Since they aren't used semantically as domain names, couldn't this be achieved by making `IdentityProviderConfig.login_hints` a sequence and failing when all members are not present in an account? The all case would require some finagling by the IDP, sure, but the reduced API complexity may well be worth it.


My concerns of adding more and more IDP-logic features into this API continue to rise, and this is an example of why. Nothing is wrong with this per-se but as it grows it is harder to get right, where we could defer to the experience of existing deployments of identity plaftorms.

npm1:
> It will be confusing for a newcomer to distinguish domain_hint and login_hint and which they want/need.

It is possible for small IdPs yea, but I think this can be prevented through proper developer documentation.

> Thinking a bit more, couldn't we just make the Since they aren't used semantically as domain names, couldn't this be achieved by making `IdentityProviderConfig.login_hints` a sequence and failing when all members are not present in an account? The all case would require some finagling by the IDP, sure, but the reduced API complexity may well be worth it.

Interesting idea! That could work, athough semantically it might be hard for IdPs. They'd need to add two different things to the same array and also figure out which corresponds to which when parsing the `login_url` which includes those fields. I actually think having the semantic separation makes it easier, not harder, for IdPs. This is why this field exists in Google Sign-In and also apparently in Microsoft[1]. Priority of constituents would imply that we should prioritize the IdP needs, WDYT?

[1] https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oapx/86fb452d-e34a-494e-ac61-e526e263b6d8

> My concerns of adding more and more IDP-logic features into this API continue to rise, and this is an example of why. Nothing is wrong with this per-se but as it grows it is harder to get right, where we could defer to the experience of existing deployments of identity plaftorms.

That is a valid concern regarding amount of features, although this one in particular is before user agrees to use federation, so in some sense is required to provide a good UX for cases where only certain accounts are valid.

bvandersloot-mozilla:
Given that this exists in Microsoft identity too, I'm more apt to include it. 

Is it possible that domainHint could be something more specific than a String? Or automatically populated? It feels weird to leave it so vague and effectively duplicative. I'm just not familiar with its typical use!

npm1:
> Given that this exists in Microsoft identity too, I'm more apt to include it.

Great!

> Is it possible that domainHint could be something more specific than a String? Or automatically populated? It feels weird to leave it so vague and effectively duplicative. I'm just not familiar with its typical use!

Hm well initially we were hoping to use something like trying to match the suffix of the email or something like that, which would make it automatic. However, we heard that this does not work in all cases: some domains go by multiple names so at least one would not really match anything about the user email. Given the requirement for this kind of flexibility, it's not possible to compute the domain hints automatically.

bvandersloot-mozilla:
Looking more closely at Microsoft's use case, they don't have a second type of hint. So I would expect that "domain_hint" for Microsoft could just be a "login_hint". Which brings me back to my original thought: is there any semantic difference between a domain hint and login hint? 

Priority of constituencies would guide us to building the easiest API to understand for all consumers, not just reproducing the conventions of one consumer into the API as an additional feature. It's my expectation that an identity dev would see an array of login_hints as an obvious place to stuff a domain_hint.

npm1:
> Looking more closely at Microsoft's use case, they don't have a second type of hint. So I would expect that "domain_hint" for Microsoft could just be a "login_hint". Which brings me back to my original thought: is there any semantic difference between a domain hint and login hint?

There is a semantic difference for sure. `login_hint` answers the question "which is an identifier for the user I want"? whereas `domain_hint` answers the question "which corporation/server must this account belong to?". The former usually targets a single user whereas the latter targets a set of users. Because both are essentially account filters, they look similar in the spec, but the hopefully that clarifies the semantic difference? In fact, you can see in the Microsoft example that they also have [login_hint](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oapx/a8622e66-2285-43c0-bbb9-abfcecdaed86) so the semantic separation between these is there in that provider as well.

> Priority of constituencies would guide us to building the easiest API to understand for all consumers, not just reproducing the conventions of one consumer into the API as an additional feature. It's my expectation that an identity dev would see an array of login_hints as an obvious place to stuff a domain_hint.

I disagree with this assessment. We have at least two examples of developers separating these, and due to the semantic difference explained above, I do believe an IdP implementing one or the other would not want to combine them.

bvandersloot-mozilla:
> In fact, you can see in the Microsoft example that they also have [login_hint](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-oapx/a8622e66-2285-43c0-bbb9-abfcecdaed86) so the semantic separation between these is there in that provider as well.

Ah, I missed that!

> I disagree with this assessment. We have at least two examples of developers separating these, and due to the semantic difference explained above, I do believe an IdP implementing one or the other would not want to combine them.

I'm not sure why they wouldn't, since they function identically. But you have done more outreach to IdPs than I have, so maybe there is insight there I don't get. 

Given this, I suppose it makes sense, but isn't my preferred way of going about it. Are there more `_hint`s lurking around the corner?

npm1:
> I'm not sure why they wouldn't, since they function identically. But you have done more outreach to IdPs than I have, so maybe there is insight there I don't get.

What I mean is that IDPs that support both would probably not combine them because having them separate is easier to reason about. Plus if you combine them then you'd have to figure out how to untangle them from the query parameters in URLs. Whereas by not combining them you do not need to guess which is is login and which one is domain hint.
 
> Given this, I suppose it makes sense, but isn't my preferred way of going about it. Are there more `_hint`s lurking around the corner?

No more hints lurking to the best of my knowledge :)

Let me know if this PR makes sense to merge from your perspective.

bvandersloot-mozilla:
This change in and of itself is reasonable, but I do still concerns over increased IDP logic in the API instead of rearchitecting to use an IDP navigatable as the way to handle several similar extensions.

samuelgoto:
> This change in and of itself is reasonable, but I do still concerns over increased IDP logic in the API instead of rearchitecting to use an IDP navigatable as the way to handle several similar extensions.

I tend to agree with both of you: I think that @npm1 is correct in thinking that these are semantically different enough that it is worth keeping them separate, but I also agree with you @bvandersloot-mozilla that we have to try to look at the API as holistically as possible (and yeah, I am not personally aware of any further `_hints`).

Some things are easier seen after the fact, as more IdPs use these extensions and tell us what works and what doesn't work. To the best of my knowledge, I think this is a reasonable approximation, but we are open to be told otherwise (and potentially incur in paying a price of backwards incompatibility).

> This change in and of itself is reasonable

Either way, I'm planning to merge this PR since it seems like we have arrived at something that seems reasonable as a way to move forward, and let ourselves be open to revisiting this as we gather more deployment experience.

LMK if you feel strongly otherwise, but if not, I'll merge this momentarily to unblock forward progress.

PR #513, opened 2023-11-15
Do not allow consider an account connected if IDP says so 

npm1:
In this PR, we fix the case where an account has been observed to be connected in the past by the user agent, but the IdP explicitly says that this account is not connected via approved_clients. This could happen if a user has revoked the federated account, and so the IdP information should be respected. In other words, if `approved_clients` is present and implies that the account is not connected, the user agent should consider it to be disconnected.

***
<a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/513.html" title="Last updated on Nov 15, 2023, 9:03 PM UTC (0fd7cc1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/513/5b07ad1...0fd7cc1.html" title="Last updated on Nov 15, 2023, 9:03 PM UTC (0fd7cc1)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -460,6 +460,10 @@ To <dfn>compute the connection status</dfn> given an {{IdentityProviderConfig}} 
 {{IdentityProviderAccount}} |account|, and a |globalObject|, run the following steps. This returns
 <dfn for="compute the connection status">connected</dfn> or
 <dfn for="compute the connection status">disconnected</dfn>.
+    1. If |account| [=map/contains=] {{IdentityProviderAccount/approved_clients}} and
+        |account|'s {{IdentityProviderAccount/approved_clients}} does not [=list/contain=]
+        |provider|'s {{IdentityProviderConfig/clientId}}, return
+        [=compute the connection status/disconnected=].
     1. Let |configUrl| be the result of running [=parse url=] with |provider|'s
         {{IdentityProviderConfig/configURL}} and |globalObject|.
     1. Let |idpOrigin| be the [=url/origin=] corresponding to |configUrl|.


spec/index.bs
```diff
@@ -460,6 +460,10 @@ To <dfn>compute the connection status</dfn> given an {{IdentityProviderConfig}}
 {{IdentityProviderAccount}} |account|, and a |globalObject|, run the following steps. This returns
 <dfn for="compute the connection status">connected</dfn> or
 <dfn for="compute the connection status">disconnected</dfn>.
+    1. If |account| [=map/contains=] {{IdentityProviderAccount/approved_clients}} and
```
cbiesinger, line 463: I am not sure that map/contains is the right thing here since account is ostensibly an IdentityProviderAccount dictionary... maybe "If |account|.{{IdentityProviderAccount/approved_clients}} is set" or something?

cbiesinger (commented):
lgtm modulo the comment below

spec/index.bs
npm1: I think the shortcut is fine based on the fact that exist here links to map/contains directly (I didnt see a definition of 'present' or 'exist' in dictionary) https://webidl.spec.whatwg.org/#dfn-dictionary-member:~:text=Other%20members%27%20entries%20might%20or%20might%20not%20exist%20in%20the%20dictionary%20value

cbiesinger: ok, fair enough

npm1:
@bvandersloot-mozilla FYI about this editorial but fairly trivial spec bug fix. I'll look to see if we need to add WPT coverage for this or if there is already.

PR #515, opened 2023-11-16
Add a disconnect method

npm1:
Fixes https://github.com/fedidcg/FedCM/issues/496. Adds a `disconnect_endpoint` and a `disconnect` method to allow the RP to request the IDP to disconnect a previously connected federated account. I opted for `disconnect_endpoint` instead of `disconnection_endpoint` because it sounded a bit better to me, but open to change it. The credentialed request is sent with IDP cookies but using CORS as the request mode. If no account is found or there is some other error, the user agent removes all connections associated with the RP, IDP.

***
<a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/pull/515.html" title="Last updated on Dec 1, 2023, 7:20 PM UTC (ad3a7f6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/515/8253c9f...ad3a7f6.html" title="Last updated on Dec 1, 2023, 7:20 PM UTC (ad3a7f6)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -452,7 +452,7 @@ the [=/origin=] of the [=RP=], |idp| is the [=/origin=] of the [=IDP=], and |acc
 an account identifier. It represents the set of triples such that the user has used FedCM to login to
 the |rp| via the |idp| |account|.
 
-Issue: the [=connected accounts set=] should be double keyed in the [=RP=] (e.g. it should include
+Issue: the [=connected accounts set=] should be double keyed in the [=RP=] (i.e., it should include
 both the requester and the embedder, or in other words the iframe and the top-level).
 
 If a user clears browsing data for an |origin| (cookies, localStorage, etc.), the user agent MUST
@@ -530,8 +530,8 @@ by the [[CM]] API.
 <!-- ============================================================ -->
 
 <div algorithm="identity-credential-disconnect">
-When the static {{IdentityCredential/disconnect}} method is invoked given an
-{{IdentityCredentialDisconnectOptions}} |options| perform the following steps:
+When the static {{IdentityCredential/disconnect}} method is invoked, given an
+{{IdentityCredentialDisconnectOptions}} |options|, perform the following steps:
 
     1. Let |globalObject| be the [=current global object=].
     1. Let |document| be |globalObject|'s [=associated Document=].
@@ -559,7 +559,7 @@ When asked to <dfn>attempt to disconnect</dfn> given an {{IdentityCredentialDisc
         passed as |configUrl|. If it fails, [=reject=] |promise| with a "{{NetworkError}}"
         {{DOMException}}.
     1. If there is another pending {{IdentityCredential/disconnect}} call for this |globalObject|
-        (e.g. it has not yet thrown an exception or its associated {{Promise}} has not yet been
+        (e.g., it has not yet thrown an exception or its associated {{Promise}} has not yet been
         resolved), [=reject=] |promise| with a "{{NetworkError}}" {{DOMException}}.
     1. If |configUrl| is not a [=potentially trustworthy origin=], [=reject=] |promise| with a
         "{{NetworkError}}" {{DOMException}}.
@@ -577,7 +577,7 @@ When asked to <dfn>attempt to disconnect</dfn> given an {{IdentityCredentialDisc
         |config|["{{IdentityProviderAPIConfig/disconnect_endpoint}}"], and |globalObject|.
     1. If |disconnectUrl| is failure, [=reject=] |promise| with a "{{NetworkError}}"
         {{DOMException}}.
-    1. [=Send a disconnect request=] with |disconnectUrl|, |options|, and |globalObject| and let
+    1. [=Send a disconnect request=] with |disconnectUrl|, |options|, and |globalObject|, and let
         |accountId| be the result.
     1. Let |idpOrigin| be the [=url/origin=] corresponding to |configUrl|.
     1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
@@ -601,7 +601,7 @@ The <a>send a disconnect request</a> algorithm sends a request to disconnect an 
 previously been used for federated login in the [=RP=].
 
 <div algorithm>
-When asked to <dfn>send a disconnect request</dfn> given a <a spec=url for=/>URL</a>
+When asked to <dfn>send a disconnect request</dfn>, given a <a spec=url for=/>URL</a>
 |disconnectUrl|, and {{IdentityCredentialDisconnectOptions}} |options|, and a |globalObject|,
 perform the following steps. This returns an {{USVString}} or failure.
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -452,6 +452,9 @@ the [=/origin=] of the [=RP=], |idp| is the [=/origin=] of the [=IDP=], and |acc
 an account identifier. It represents the set of triples such that the user has used FedCM to login to
 the |rp| via the |idp| |account|.
 
+Issue: the [=connected accounts set=] should be double keyed in the [=RP=] (e.g. it should include
+both the requester and the embedder, or in other words the iframe and the top-level).
+
 If a user clears browsing data for an |origin| (cookies, localStorage, etc.), the user agent MUST
 [=list/remove=] all triples with an [=/origin=] matching the |origin| from <a>connected accounts set</a>.
 
@@ -495,13 +498,13 @@ To <dfn>create a connection between the RP and the IdP account</dfn> given an
 This specification introduces a new type of {{Credential}}, called an {{IdentityCredential}}:
 
 <pre class="idl">
-  dictionary IdentityCredentialRevokeOptions : IdentityProviderConfig {
+  dictionary IdentityCredentialDisconnectOptions : IdentityProviderConfig {
     USVString accountHint;
   };
 
   [Exposed=Window, SecureContext]
   interface IdentityCredential : Credential {
-    static Promise&lt;undefined&gt; revoke(optional IdentityCredentialRevokeOptions options = {});
+    static Promise&lt;undefined&gt; disconnect(optional IdentityCredentialDisconnectOptions options = {});
     readonly attribute USVString? token;
   };
 </pre>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -452,6 +452,9 @@ the [=/origin=] of the [=RP=], |idp| is the [=/origin=] of the [=IDP=], and |acc
 an account identifier. It represents the set of triples such that the user has used FedCM to login to
 the |rp| via the |idp| |account|.
 
+Issue: the [=connected accounts set=] should be double keyed in the [=RP=] (e.g. it should include
```
TallTed, line 455:
```suggestion
Issue: the [=connected accounts set=] should be double keyed in the [=RP=] (i.e., it should include
```

spec/index.bs
```diff
@@ -517,6 +525,137 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
 The main entrypoint in this specification is through the entrypoints exposed
 by the [[CM]] API.
 
+<!-- ============================================================ -->
+### The disconnect method ### {#browser-api-identity-credential-disconnect}
+<!-- ============================================================ -->
+
+<div algorithm="identity-credential-disconnect">
+When the static {{IdentityCredential/disconnect}} method is invoked given an
+{{IdentityCredentialDisconnectOptions}} |options| perform the following steps:
```
TallTed, line 534:
```suggestion
When the static {{IdentityCredential/disconnect}} method is invoked, given an
{{IdentityCredentialDisconnectOptions}} |options|, perform the following steps:
```

TallTed (changes requested):
Small stuff for clarity

npm1:
Ping

spec/index.bs
```diff
@@ -517,6 +525,137 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
...
+    1. Let |promise| be a new {{Promise}}.
+    1. [=In parallel=], [=attempt to disconnect=] given |options|, |promise|, and |globalObject|.
+    1. Return |promise|.
+</div>
+
+<div algorithm>
+When asked to <dfn>attempt to disconnect</dfn> given an {{IdentityCredentialDisconnectOptions}}
+|options|, a {{Promise}} |promise|, and a |globalObject|, perform the following steps:
+
+    1. Assert: these steps are running [=in parallel=].
```
yi-gu, line 553: Maybe I misunderstood: the following steps are still sequential, right? e.g. you have to fetch `configUrl` first before parsing `disconnectUrl` etc.

npm1: Right, that is not what in parallel means. If you click the link it says what it actually means https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel

spec/index.bs
```diff
@@ -1850,6 +1999,61 @@ For example:
...
+{{DisconnectedAccount}} without an exception.
+
+<dl dfn-type="dict-member" dfn-for="DisconnectedAccount">
+    :   <dfn>account_id</dfn>
+    ::  The {{IdentityProviderAccount/id}} of the account that was successfully disconnected.
+</dl>
+
+The [=IDP=] must return the {{DisconnectedAccount/account_id}} since it may be different from the
+{{disconnect_endpoint_request/account_hint}}, and the ID is the one which allows the
+[=user agent=] to disconnect the account from the [=connected accounts set=].
```
yi-gu, line 2055: I know it was mentioned above but should we call out here that all accounts associated with the RP will be disconnected if something is wrong here?

spec/index.bs
yi-gu: ah got it.

yi-gu:
lgtm

spec/index.bs
npm1: Added

cbiesinger (approved):

spec/index.bs
```diff
@@ -517,6 +525,137 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
...
+    1. Let |idpOrigin| be the [=url/origin=] corresponding to |configUrl|.
+    1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
+    1. If |accountId| is not failure:
+        1. Let |triple| be (|rpOrigin|, |idpOrigin|, |accountId|).
+        1. If [=connected accounts set=] [=list/contains=] |triple|:
+            1. [=list/Remove=] |triple| from the [=connected accounts set=].
+            1. [=Resolve=] |promise|.
+            1. Return.
+    1. For every (|rp|, |idp|, |accountId|) |triple| in the [=connected accounts set=]:
+        1. If |rp| equals |rpOrigin| and |idp| equals |idpOrigin|, [=list/remove=] |triple| from the
```
samuelgoto, line 591: Same here. Lets abstract this away from the callers.

spec/index.bs
```diff
@@ -517,6 +525,137 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
...
+    1. If there does not exist an account |account| such that [=compute the connection status=] of
+        |options|, |account|, and |globalObject| returns
+        [=compute the connection status/connected=], then [=reject=] |promise| with a
+        "{{NetworkError}}" {{DOMException}}. This check can be performed by iterating over the
+        [=connected accounts set=] or by keeping a separate data structure to make this lookup fast.
+    1. Let |config| be the result of running [=fetch the config file=] with
+        |provider| and |globalObject|.
+    1. If |config| is failure, [=reject=] |promise| with a "{{NetworkError}}" {{DOMException}}.
+    1. Let |disconnectUrl| be the result of [=computing the manifest URL=] given |provider|,
+        |config|["{{IdentityProviderAPIConfig/disconnect_endpoint}}"], and |globalObject|.
```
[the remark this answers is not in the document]
npm1: Ok

spec/index.bs
```diff
@@ -517,6 +525,137 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
...
+    1. If |disconnectUrl| is failure, [=reject=] |promise| with a "{{NetworkError}}"
+        {{DOMException}}.
+    1. [=Send a disconnect request=] with |disconnectUrl|, |options|, and |globalObject|, and let
+        |accountId| be the result.
+    1. Let |idpOrigin| be the [=url/origin=] corresponding to |configUrl|.
+    1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
+    1. If |accountId| is not failure:
+        1. Let |triple| be (|rpOrigin|, |idpOrigin|, |accountId|).
+        1. If [=connected accounts set=] [=list/contains=] |triple|:
+            1. [=list/Remove=] |triple| from the [=connected accounts set=].
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -517,6 +541,131 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
...
+    1. If |disconnectUrl| is failure, [=reject=] |promise| with a "{{NetworkError}}"
+        {{DOMException}}.
+    1. [=Send a disconnect request=] with |disconnectUrl|, |options|, and |globalObject|, and let
+        |accountId| be the result.
+    1. Let |idpOrigin| be the [=url/origin=] corresponding to |configUrl|.
+    1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
+    1. [=Remove a connection=] for |accountId|, |rpOrigin|, and |idpOrigin| and let |foundAccount|
+        be the result.
+    1. If |foundAccount|, [=resolve=] |promise|, otherwise [=reject=] |promise| with a
+        "{{NetworkError}}" {{DOMException}}.
```
samuelgoto, line 603: hummm there is something off here i think .... we throw a NetworkError, but the [=Send a disconnect request=] actually succeeded. The |foundAccount| wasn't found within the internals of the browser, but the actual request succeeded ... i can't quite put my finger on what's wrong here, but this feels off to me: it feels like the actual promise succeeded: the network request was made successfully, the connected set was updated correctly (by removing *any* account), etc,  but we are telling the RP that there was an error?

spec/index.bs
```diff
@@ -517,6 +525,137 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
...
+{{IdentityCredentialDisconnectOptions}} |options|, perform the following steps:
+
+    1. Let |globalObject| be the [=current global object=].
+    1. Let |document| be |globalObject|'s [=associated Document=].
+    1. If |document| is not [=allowed to use=] the [=identity-credentials-get=]
+        [=policy-controlled feature=], throw a "{{NotAllowedError}}" {{DOMException}}.
+    1. If |options| does not [=map/contain=] all of the following, then throw a new {{TypeError}}:
+        * {{IdentityProviderConfig/configURL}}
+        * {{IdentityProviderConfig/clientId}}
+        * {{IdentityCredentialDisconnectOptions/accountHint}}
```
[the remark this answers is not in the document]
npm1: This is checking whether or not they are empty, eg. having `configUrl: ""` as opposed to not having it at all. Do you think we should remove this check altogether?

spec/index.bs
npm1: Oh yes I actually want this to depend on whether IdP disconnection server-side succeeded or not. PTAL?

spec/index.bs
```diff
@@ -488,15 +493,34 @@ To <dfn>create a connection between the RP and the IdP account</dfn> given an
     1. [=set/Append=] |triple| to [=connected accounts set=].
 </div>
 
+<div algorithm>
+To <dfn>remove a connection</dfn> given |accountId|, |rpOrigin|, |idpOrigin|, run the following
+steps. It returns whether a connection matching the |accountId| was found and removed:
+    1. If |accountId| is not failure:
```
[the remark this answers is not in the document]

spec/index.bs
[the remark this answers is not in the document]
samuelgoto: > This is checking whether or not they are empty, eg. having configUrl: "" as opposed to not having it at all. 

```
If |options| does not [=map/contain=] all of the following, then throw a new {{TypeError}}:
```

[=map/contain=] just checks for the presence of the key though, not for whether its value is empty-string, right?

> Do you think we should remove this check altogether?

Don't we already fallback gracefully when those values are empty-strings further down into the algorithms?

spec/index.bs
```diff
@@ -517,6 +545,134 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
...
+        [=connected accounts set=] or by keeping a separate data structure to make this lookup fast.
+    1. Let |config| be the result of running [=fetch the config file=] with
+        |provider| and |globalObject|.
+    1. If |config| is failure, [=reject=] |promise| with a "{{NetworkError}}" {{DOMException}}.
+    1. Let |disconnectUrl| be the result of [=computing the manifest URL=] given |provider|,
+        |config|.{{IdentityProviderAPIConfig/disconnect_endpoint}}, and |globalObject|.
+    1. If |disconnectUrl| is failure, [=reject=] |promise| with a "{{NetworkError}}"
+        {{DOMException}}.
+    1. [=Send a disconnect request=] with |disconnectUrl|, |options|, and |globalObject|, and let
+        |accountId| be the result.
```
samuelgoto, line 601:
```suggestion
        |result| be the result.
```

I'm having a hard time parsing this algorithm here under the assumption that |accountId| can be a failure. I'd suggest we rename this to |result|, and check for failure early on and if no failure, assign |result| to a new variable |accountId|.

For example:

```
1. If |result| is failure
    1. [=remove all connections=] given |rpOrigin| and |idpOrigin|.
    1. [=reject=] |promise| with a "{{NetworkError}}" {{DOMException}}.
1. Let |accountId| be |result|. // carries on assuming there wasn't a network error
```

spec/index.bs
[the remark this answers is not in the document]
npm1: ? I think you are not looking at the latest commits? "If any of the following members from |options| are empty". I think configURL we would fail, less clear for clientId and accountHint (we'd at least send fetches, eg fail later than we should)

spec/index.bs
```diff
@@ -517,6 +545,134 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
...
+    1. Let |idpOrigin| be the [=url/origin=] corresponding to |configUrl|.
+    1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
+    1. Let |wasAccountRemoved| be false.
```
samuelgoto, line 604:
```suggestion
```

Is this really necessary? You are already assigning wasAccountRemoved in the next step as the result of the [=remove the connection=] algorithm?

spec/index.bs
samuelgoto: I think this looks better, but I made a few suggestions to try to return early when a network error occurs, so that it becomes easier to reason about the non-error cases. WDYT?

spec/index.bs
```diff
@@ -488,15 +493,38 @@ To <dfn>create a connection between the RP and the IdP account</dfn> given an
...
+following steps. It returns whether the |accountId| connection was successfully removed.
+    1. Let |triple| be (|rpOrigin|, |idpOrigin|, |accountId|).
+    1. If [=connected accounts set=] [=list/contains=] |triple|:
+        1. [=list/Remove=] |triple| from the [=connected accounts set=].
+        1. Return true.
+    1. Return false.
+</div>
+
+<div algorithm>
+To <dfn>remove all connections</dfn> given |rpOrigin| and |idpOrigin|, run the following steps:
```
samuelgoto, line 507: I really like the separation here between the two algorithms: removing the connection by accountId and removing all connections.

spec/index.bs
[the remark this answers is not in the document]
samuelgoto: Do we really care if they are empty? Can't we make the disconnect network request call without them? This seems like an artificial requirement that Rps are going to put garbage in (like "client_id=foobar", "accountHint=goaway) just to make this work.

That is, the disconnect network call is already responsible for handling requests where client_id or accountHint doesn't match anything in their servers, I'm wondering if we should just get out of the way here, as opposed to checking for the empty-ness of the strings here.

spec/index.bs
npm1: Thanks

spec/index.bs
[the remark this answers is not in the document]
npm1: Sure, I think we can remove these checks then

spec/index.bs

spec/index.bs
npm1: Applied the suggestions, thanks!

samuelgoto (approved):

samuelgoto:
LGTM

@bvandersloot-mozilla WDYT?

spec/index.bs
```diff
@@ -494,17 +494,21 @@ To <dfn>create a connection between the RP and the IdP account</dfn> given an
...
 <div algorithm>
-To <dfn>remove a connection</dfn> given |accountId|, |rpOrigin|, |idpOrigin|, run the following
-steps. It returns whether a connection matching the |accountId| was found and removed:
-    1. If |accountId| is not failure:
-        1. Let |triple| be (|rpOrigin|, |idpOrigin|, |accountId|).
-        1. If [=connected accounts set=] [=list/contains=] |triple|:
-            1. [=list/Remove=] |triple| from the [=connected accounts set=].
-            1. Return true.
-    1. For every (|rp|, |idp|, |accountId|) |triple| in the [=connected accounts set=]:
+To <dfn>remove the connection</dfn> given |accountId|, |rpOrigin|, and |idpOrigin|, run the
```
TallTed, line 497:
```suggestion
To <dfn>remove a connection</dfn>: given |accountId|, |rpOrigin|, and |idpOrigin|, run the
```

spec/index.bs
```diff
@@ -594,13 +594,18 @@ When asked to <dfn>attempt to disconnect</dfn> given an {{IdentityCredentialDisc
...
-    1. [=Remove a connection=] for |accountId|, |rpOrigin|, and |idpOrigin| and let |foundAccount|
-        be the result.
-    1. If |foundAccount|, [=resolve=] |promise|, otherwise [=reject=] |promise| with a
-        "{{NetworkError}}" {{DOMException}}.
+    1. If |result| is failure:
+        1. [=Remove all connections=] given |rpOrigin| and |idpOrigin|.
+        1. [=Reject=] |promise| with a "{{NetworkError}}" {{DOMException}}.
+        1. Return.
+    1. Let |accountId| be |result| (note that it is not failure).
+    1. [=Remove the connection=] given |accountId|, |rpOrigin|, and |idpOrigin| and let
```
TallTed, line 605:
```suggestion
    1. [=Remove a connection=] using |accountId|, |rpOrigin|, and |idpOrigin|, and let
```

bvandersloot-mozilla:
I think I follow and LGTM.

samuelgoto:
> I think I follow and LGTM.

Thanks @bvandersloot-mozilla ! @npm1 i'll wait for you to resolve the last remaining bits of feedback from @TallTed and then squash and merge.

PR #519, opened 2023-11-23
Delete proposals directory

npm1:
We have agreed to keep explainers in GH issues to allow for discussion, and all of the proposals have either been added to spec or moved into issues.


npm1:
Looking more closely at the list, I think the only one not properly represented might be https://github.com/fedidcg/FedCM/blob/main/proposals/idp-sign-out.md. Am I missing something or should I file an issue for that one?

samuelgoto:
> Looking more closely at the list, I think the only one not properly represented might be https://github.com/fedidcg/FedCM/blob/main/proposals/idp-sign-out.md. Am I missing something or should I file an issue for that one?

We haven't made progress on front-channel logout for a long time, so if we re-open this then we could just rewrite it from scratch, since so much of our understanding has changed.

Thanks for putting this together, merging.

PR #521, opened 2023-11-24
Update HOWTO-chrome.md with DomainHint and Disconnect

npm1:


explorations/HOWTO-chrome.md
```diff
@@ -218,3 +218,91 @@ if (cred.isAutoSelected !== undefined) {
   const isAutoSelected = cred.isAutoSelected;
 }
 ```
+
+### DomainHint
+
+To use the DomainHint:
+* Ensure that chrome://version shows TODO or higher.
```
npm1, line 225: Waiting for the Chrome version to include a change so I can include the version here.

yi-gu: FedCmDomainHint seems exist in chrome://flags already. Are you waiting for any specific change?

explorations/HOWTO-chrome.md
```diff
@@ -218,3 +218,91 @@ if (cred.isAutoSelected !== undefined) {
...
+* Enable the experimental feature `FedCmDomainHint` in `chrome://flags`.
+
+* Add an array of `domain_hints` to the accounts described in the accounts endpoint:
+
+```
+{
+  accounts: [{
+    id: "accountId",
+    email: "account@email.com",
```
yi-gu, line 234: nit: add a "name" field since it's required

explorations/HOWTO-chrome.md
```diff
@@ -218,3 +218,91 @@ if (cred.isAutoSelected !== undefined) {
...
+    domain_hints: ["corp1", "corp2"],
+    ...
+  }, ...]
```
yi-gu, line 237: Would it be better if you have another two accounts here: one with `corp3` and one without any domain hint. Then in the example below only the first account can be displayed; you can have another example using "any" which can display the first two accounts.

explorations/HOWTO-chrome.md
```diff
@@ -218,3 +218,91 @@ if (cred.isAutoSelected !== undefined) {
...
+satisfy CORS, and in the body of the response it should include a JSON with the
+account ID of the account that has been disconnected.
+
+```json
+{
+  "account_id": "account456Id"
+}
+```
+
+If the browser finds an account matching the ID provided, it will note the
```
yi-gu, line 303: Please elaborate a bit on "finds an account". e.g. from where, how it was there in the first place.

explorations/HOWTO-chrome.md
```diff
@@ -218,3 +218,91 @@ if (cred.isAutoSelected !== undefined) {
...
+disconnection of that account. If the IdP fails or the `account_id` is nowhere
```
yi-gu, line 304: Similarly, elaborate a bit on "IdP fails" what

explorations/HOWTO-chrome.md
```diff
@@ -218,3 +218,91 @@ if (cred.isAutoSelected !== undefined) {
...
+to be found, the browser will remove from its memory all of the federated
+accounts associated with the (RP, IDP). As a reminder, an account needs to be
+in the browser memory in order for both auto reauthn and the User Info API to
```
yi-gu, line 307: Would "memory" be confusing since it could refer to RAM? How about using "local storage"?

explorations/HOWTO-chrome.md
npm1: https://chromium-review.googlesource.com/c/chromium/src/+/5053387

explorations/HOWTO-chrome.md
npm1: Well that is why I had the ... but sure

yi-gu:
lgtm

PR #523, opened 2023-11-29
Add third party cookie relaxations for FedCM

### LINKED ISSUE #517: Allow user agents to use "Connected Accounts Set" with flexibility
Currently the [spec](https://fedidcg.github.io/FedCM/#browser-connected-accounts-set) will treat an account "connected" if a user has explicitly used that account to create a connection between an RP and an IdP. e.g. when a user uses account `foo` to sign in to `rpOrigin` with `idpOrigin`, the user agent should add the tuple {rpOrigin, idpOrigin, "foo"} to the Connected Accounts Set as a sign that the user agent has observed a "sign-in" moment.

For features that require an account being connected, e.g. auto re-authentication, the user agent currently only trusts its own connected accounts set even though it could be incorrect. e.g. if a user has created a federated account on the RP with the IdP without the user agent knowing it (e.g. via another user agent), then the user will be treated as a new user on the RP and therefore won't be eligible for those features. Even if an IdP as the source of truth has claimed that the user has created an account on the RP via `approved_clients`, the user agent would still ignore that signal. The design was mainly from privacy's perspective.

However, the privacy benefit of the design may be lost if the IdP has third-party cookies (3PC) access on the RP site. e.g. pop-up heuristics is (to be) implemented in many major browsers even though it's meant to be temporary. If an IdP has gained 3PC access via heuristics, they could communicate with themselves using 3PC and the privacy bar that FedCM sets would make less sense.

Therefore we should revisit the decision and see if user agents could have some flexibility when using the "Connected Accounts Set" to determine whether to trigger "returning user only" features.

Note: we should not change how "Connected Accounts Set" is calculated to keep it in a clean state such that when user agents remove 3PC access from the IdPs in the future (e.g. deprecating heuristics) we can still hold the high privacy bar.

npm1:
Fixes https://github.com/fedidcg/FedCM/issues/517

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/pull/523.html" title="Last updated on Jul 16, 2026, 6:53 PM UTC (465088e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/523/858d07e...465088e.html" title="Last updated on Jul 16, 2026, 6:53 PM UTC (465088e)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -374,8 +374,8 @@ If a user clears browsing data for an |origin| (cookies, localStorage, etc.), th
 [=same origin=] with |rp|, |idp|, or |embedder|.
 
 <div algorithm>
-To <dfn>compute the connected account key</dfn> given an {{IdentityProviderConfig}} |provider|, an
-{{IdentityProviderAccount}} |account|, and a |globalObject|, run the following steps. It returns a
+To <dfn>compute the connected account key</dfn> given an {{IdentityProviderConfig}} |provider|, a
+{{USVString}} |accountId|, and a |globalObject|, run the following steps. It returns a
 quadruple of the form (embedder, rp, idp, account).
     1. Let |configUrl| be the result of running [=parse url=] with |provider|'s
         {{IdentityProviderConfig/configURL}} and |globalObject|.
@@ -383,7 +383,6 @@ quadruple of the form (embedder, rp, idp, account).
         [=navigable/top-level traversable=]'s [=navigable/active document=]'s [=Document/origin=].
     1. Let |idpOrigin| be the [=url/origin=] corresponding to |configUrl|.
     1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
-    1. Let |accountId| be |account|'s {{IdentityProviderAccount/id}}.
     1. Return (|embedderOrigin|, |rpOrigin|, |idpOrigin|, |accountId|).
 </div>
 
@@ -395,8 +394,16 @@ When asked whether an {{IdentityProviderAccount}} |account| is
         |account|'s {{IdentityProviderAccount/approved_clients}} does not [=list/contain=]
         |provider|'s {{IdentityProviderConfig/clientId}}, return false.
     1. Let |tuple| be the result of running [=compute the connected account key=] given |provider|,
-        |account|, and |globalObject|.
-    1. Return whether [=connected accounts set=] [=list/contains=] |tuple|.
+        |account|'s {{IdentityProviderAccount/id}}, and |globalObject|.
+    1. If [=connected accounts set=] [=list/contains=] |tuple|, return true.
+    1. Let |idpOrigin| be the [=url/origin=] corresponding to |provider|'s
+        {{IdentityProviderConfig/configURL}}.
+    1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
+    1. If |account| [=map/contains=] {{IdentityProviderAccount/approved_clients}},
+        |account|'s {{IdentityProviderAccount/approved_clients}} [=list/contains=] |provider|'s
+        {{IdentityProviderConfig/clientId}}, and third party cookies are available for |idpOrigin|
+        on |rpOrigin|, return true.
+    1. Return false.
 </div>
 
 <div algorithm="compute the connection status">
@@ -409,7 +416,7 @@ This returns <dfn for="compute the connection status">connected</dfn> or
             {{IdentityProviderConfig/clientId}}, return [=compute the connection status/connected=].
         1. Return [=compute the connection status/disconnected=].
     1. Let |tuple| be the result of running [=compute the connected account key=] given |provider|,
-        |account|, and |globalObject|.
+        |account|'s {{IdentityProviderAccount/id}}, and |globalObject|.
     1. If [=connected accounts set=] [=list/contains=] |tuple|, return
         [=compute the connection status/connected=].
     1. Return [=compute the connection status/disconnected=].
@@ -417,15 +424,14 @@ This returns <dfn for="compute the connection status">connected</dfn> or
 
 <div algorithm>
 To <dfn>create a connection between the RP and the IdP account</dfn> given an
-{{IdentityProviderConfig}} |provider|, an {{IdentityProviderAccount}} |account|, and a
-|globalObject| (the [=RP=]'s), run the following steps:
+{{IdentityProviderConfig}} |provider|, a {{USVString}} |accountId|, and a |globalObject| (the
+[=RP=]'s), run the following steps:
     1. Let |configUrl| be the result of running [=parse url=] with |provider|'s
         {{IdentityProviderConfig/configURL}} and |globalObject|.
     1. Let |embedderOrigin| be the |globalObject|'s [=Window/navigable=]'s
         [=navigable/top-level traversable=]'s [=navigable/active document=]'s [=Document/origin=].
     1. Let |idpOrigin| be the [=url/origin=] corresponding to |configUrl|.
     1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
-    1. Let |accountId| be |account|'s {{IdentityProviderAccount/id}}.
     1. Let |tuple| be (|embedderOrigin|, |rpOrigin|, |idpOrigin|, |accountId|).
     1. [=set/Append=] |tuple| to [=connected accounts set=].
 </div>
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -472,6 +472,10 @@ To <dfn>compute the connection status</dfn> given an {{IdentityProviderConfig}}
     1. Let |triple| be (|rpOrigin|, |idpOrigin|, |accountId|).
     1. If [=connected accounts set=] [=list/contains=] |triple| , return
         [=compute the connection status/connected=].
+    1. Otherwise, if account| [=map/contains=] {{IdentityProviderAccount/approved_clients}},
+        |account|'s {{IdentityProviderAccount/approved_clients}} [=list/contains=] |provider|'s
+        {{IdentityProviderConfig/clientId}}, and third party cookies are available for |idpOrigin|
```
cbiesinger, line 477: Should/can we link to something that more explicitly defines what "third party cookies are available" means?

samuelgoto: > Should/can we link to something that more explicitly defines what "third party cookies are available" means?

The most immediate thing that comes to mind is the following API:

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/cookieEnabled

It is not implemented consistently across browsers, but it is a well defined API.

spec/index.bs
```diff
@@ -1411,13 +1415,16 @@ When invoking the {{IdentityProvider/getUserInfo()}} method given an {{IdentityP
...
-        {{DOMException}}. This check can be performed by iterating over the
-        [=connected accounts set=] or by keeping a separate data structure to make this lookup fast.
     1. Let |configUrl| be the result of running [=parse url=] with |provider|'s
         {{IdentityProviderConfig/configURL}} and |globalObject|.
+    1. If both of the following are true, throw a new "{{NetworkError}}" {{DOMException}}:
+        * There does not exist an account |account| such that [=compute the connection status=] of
+            |provider|, |account|, and |globalObject| returns
+            [=compute the connection status/connected=]. This check can be performed by iterating
+            over the [=connected accounts set=] or by keeping a separate data structure to make this
+            lookup fast.
```
TallTed, line 1425:
```suggestion
            over the [=connected accounts set=] or by keeping a separate data structure for faster
            lookup results.
```

spec/index.bs
npm1: That doesn't work, as JS attribute getters should not be called from IDL algorithms in general. The attribute getter definition is pretty generic too, and when I asked @johannhof whether there is something we can link he mentioned that there isn't and that cookie layering is what would provide something, but that's not done yet.

spec/index.bs
```diff
@@ -472,6 +472,10 @@ To <dfn>compute the connection status</dfn> given an {{IdentityProviderConfig}}
     1. Let |triple| be (|rpOrigin|, |idpOrigin|, |accountId|).
     1. If [=connected accounts set=] [=list/contains=] |triple| , return
         [=compute the connection status/connected=].
+    1. Otherwise, if account| [=map/contains=] {{IdentityProviderAccount/approved_clients}},
+        |account|'s {{IdentityProviderAccount/approved_clients}} [=list/contains=] |provider|'s
+        {{IdentityProviderConfig/clientId}}, and third party cookies are available for |idpOrigin|
+        on |rpOrigin|, return [=compute the connection status/connected=].
```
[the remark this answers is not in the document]
npm1: What exactly is the problem with this approach? What you are suggesting does not work as for instance we'd show PP/TOS for these accounts when we should in fact treat them as returning accounts for the PP/TOS purpose as well.

npm1:
Based on discussion, we will need https://github.com/fedidcg/FedCM/pull/526 first before moving forward on this PR.

spec/index.bs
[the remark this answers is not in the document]
npm1: Resolving but note that the connection status returning something does not mean that the connected set gets updated.

npm1 (commented):
This is ready for another review, ptal

spec/index.bs
```diff
@@ -391,9 +391,16 @@ This returns <dfn for="compute the connection status">connected</dfn> or
...
+    1. Let |idpOrigin| be the [=url/origin=] corresponding to |provider|'s
+        {{IdentityProviderConfig/configURL}}.
+    1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
     1. If [=connected accounts set=] [=list/contains=] |triple|, return
         [=compute the connection status/connected=].
-    1. Return [=compute the connection status/disconnected=].
+    1. Otherwise, if account| [=map/contains=] {{IdentityProviderAccount/approved_clients}},
+        |account|'s {{IdentityProviderAccount/approved_clients}} [=list/contains=] |provider|'s
+        {{IdentityProviderConfig/clientId}}, and third party cookies are available for |idpOrigin|
+        on |rpOrigin|, return [=compute the connection status/connected=].
```
yi-gu, line 402: Do you think we should call out that when `connected` is returned this way, UA should not create a connection [here](https://w3c-fedid.github.io/FedCM/#ref-for-create-a-connection-between-the-rp-and-the-idp-account)? It does say "Prompt the user to gather explicit intent to create an account" in step 3 from the link above.

spec/index.bs
```diff
@@ -472,6 +472,10 @@ To <dfn>compute the connection status</dfn> given an {{IdentityProviderConfig}}
     1. Let |triple| be (|rpOrigin|, |idpOrigin|, |accountId|).
     1. If [=connected accounts set=] [=list/contains=] |triple| , return
         [=compute the connection status/connected=].
+    1. Otherwise, if account| [=map/contains=] {{IdentityProviderAccount/approved_clients}},
```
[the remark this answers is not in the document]
TallTed: Resolved? but it appears not applied... `if account|` should be `if |account|`!

spec/index.bs
```diff
@@ -1411,13 +1415,16 @@ When invoking the {{IdentityProvider/getUserInfo()}} method given an {{IdentityP
...
-        [=connected accounts set=] or by keeping a separate data structure to make this lookup fast.
     1. Let |configUrl| be the result of running [=parse url=] with |provider|'s
         {{IdentityProviderConfig/configURL}} and |globalObject|.
+    1. If both of the following are true, throw a new "{{NetworkError}}" {{DOMException}}:
+        * There does not exist an account |account| such that [=compute the connection status=] of
+            |provider|, |account|, and |globalObject| returns
+            [=compute the connection status/connected=]. This check can be performed by iterating
+            over the [=connected accounts set=] or by keeping a separate data structure to make this
+            lookup fast.
+        * There is third party cookie access for |configUrl|'s [=url/origin=] on |document|'s
```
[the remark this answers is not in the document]
TallTed: Resolved, but not applied? `is third` should be `is no third`!

spec/index.bs
TallTed: Resolved but not applied? There are certainly multiple language changes which might be made to make this better; mine was just one suggestion. But *some* change should be made.
```suggestion
            over the [=connected accounts set=] or by keeping a separate data structure to make this
            lookup faster.
```

TallTed:
```suggestion
            over the [=connected accounts set=] or by keeping a separate data structure to make these
            lookups faster.
```

spec/index.bs

npm1:
ptal

spec/index.bs
```diff
@@ -1977,13 +1983,17 @@ The <dfn>in parallel steps for getUserInfo</dfn> given a {{Promise}} |promise| a
...
     1. Let |configUrl| be the result of running [=parse url=] with |provider|'s
         {{IdentityProviderConfig/configURL}} and |globalObject|.
+    1. If both of the following are true, [=queue to reject=] |promise| with "{{NetworkError}}"
+        and return:
+        * There does not exist an account |account| such that [=compute the connection status=] of
+            |provider|, |account|, and |globalObject| returns
+            [=compute the connection status/connected=]. This check can be performed by iterating
+            over the [=connected accounts set=] or by keeping a separate data structure to make these
+            lookups faster.
+        * There is no third party cookie access for |configUrl|'s [=url/origin=] on |document|'s
```
yi-gu, line 1997: we should swap this step with the configUrl parse check below, right?

spec/index.bs
```diff
@@ -396,8 +394,16 @@ When asked whether an {{IdentityProviderAccount}} |account| is
...
         |provider|'s {{IdentityProviderConfig/clientId}}, return false.
     1. Let |tuple| be the result of running [=compute the connected account key=] given |provider|,
-        |account|, and |globalObject|.
-    1. Return whether [=connected accounts set=] [=list/contains=] |tuple|.
+        |account|'s {{IdentityProviderAccount/id}}, and |globalObject|.
+    1. If [=connected accounts set=] [=list/contains=] |tuple|, return true.
+    1. Let |idpOrigin| be the [=url/origin=] corresponding to |provider|'s
+        {{IdentityProviderConfig/configURL}}.
+    1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
+    1. If |account| [=map/contains=] {{IdentityProviderAccount/approved_clients}},
```
yi-gu, line 402: In step 1 above we have checked that the client id is on the list so this may not be needed.

spec/index.bs
npm1: Ah yes since this uses configUrl, done

yi-gu (approved):

spec/index.bs
npm1: Step 1 checks whether the list exists and client_id is not present. So it is possible still for the list to not exist at this point.

yi-gu: Right. Technically speaking we can omit the "approved_clients contains client_id" check here but maybe it's more readable this way? Up to you.

PR #528, opened 2023-12-18
Make typogram diagrams work in dark mode

tabatkins:
Since the typogram library explicitly uses black fill/stroke and a transparent bg, it fails badly when the spec is in dark mode. Ideally this would be fixed upstream, but for now this is a minimal fix that appears to catch everything necessary.

***
<a href="https://pr-preview.s3.amazonaws.com/tabatkins/FedCM/pull/528.html" title="Last updated on Dec 18, 2023, 10:02 PM UTC (8ce0e9a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/528/baf2b3d...tabatkins:8ce0e9a.html" title="Last updated on Dec 18, 2023, 10:02 PM UTC (8ce0e9a)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -74,6 +74,15 @@ dl.domintro dt code {
     padding: 8px;
     border-left: 8px solid brown;
 }
+
+/* temporary fixes to the typogram diagrams
+   to support dark mode properly */
+script + svg :is(polygon, line, rect):not(.grid) {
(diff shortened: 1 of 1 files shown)


npm1 (approved):
Thanks for fixing this!

Issue #537, opened 2024-01-17
Allow setting IDP login status from same-site subresources


cbiesinger:
### Background
The [login status API](https://fedidcg.github.io/FedCM/#browser-api-login-status) currently only allow setting the login status on toplevel loads or for subresources who are same-origin with all their ancestors, both when using the JavaScript API and when using the HTTP header.

However, some IDPs want to use one origin for their FedCM endpoints but log in users on a different origin. With the same-origin restriction, this would require a toplevel redirect or opening a popup, which leads to disruptive user experiences especially if the current login flow does not depend on a top-level navigation.

### Proposal
We are proposing to relax the same-origin restriction to a same-site (that is, same eTLD+1) restriction. For example:

```js
// This is on a toplevel page on URL https://login.idp.example/
// FedCM endpoints are on https://fedcm.idp.example/
// We are logging in to the IDP.
function login(user, password) {
  if (checkPassword(user, password)) {
    window.cookie = 'loggedin=true';
    // This will send a `Set-Login: logged-in` header
    fetch('https://fedcm.idp.example/set-loggedin');
  }
}

function logout() {
  window.cookie = 'loggedin=false';
  // This will send a `Set-Login: logged-out` header
  fetch('https://fedcm.idp.example/set-loggedout');
}
```

The login status headers in this example will currently be ignored, but in the proposal would be processed.

Subsequent FedCM calls to `https://fedcm.idp.example/fedcm.json` would then be allowed:

```js
// This is a toplevel page on https://www.rp.example
// We are logging in to the RP using FedCM with idp.example.

function loginWithIdp() {
  let token = await navigator.credentials.get({identity: {providers: [{
      configURL: "https://fedcm.idp.example/fedcm.json"}]}});
  // Now we are logged in with idp.example!
}
```

Because the initial login state for an IDP in Chrome is “unknown” and not “logged-out”, this example would actually work initially even in the absence of login status setters. However, if the user is not logged in to the IDP during the very first time FedCM is used with that IDP, we would mark the status as logged-out and require a valid login status header or JS call to update this status. Therefore, it is important that IDPs send a header that Chrome considers valid.

Identity Providers should consider making the subresource request with CORS and credentials included so that their endpoint can verify the correct login state.

### Security & Privacy
The same-origin restriction was originally introduced so that an RP can not ‘artificially’ set an IDP’s login status by just making an invisible fetch() request or using an invisible iframe and thus allow potentially-tracking FedCM requests every time.

[...]

PR #538, opened 2024-01-17
Relax the login status requirement from same-origin  to same-site

cbiesinger:
Some IDPs have their login on one subdomain but the FedCM endpoint on a different subdomain, and this change lets them set the login status on the correct origin.

Bug: #537

***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/538.html" title="Last updated on Jan 18, 2024, 7:13 PM UTC (cfdc752)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/fedidcg/FedCM/538/b3674e5...cbiesinger:cfdc752.html" title="Last updated on Jan 18, 2024, 7:13 PM UTC (cfdc752)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -354,6 +354,26 @@ value |value|:
 
 </div>
 
+### Infrastructure algorithm ### {#infra-algorithm}
+
+<div algorithm>
+An [=environment settings object=] (|settings|) is <dfn noexport>same-site with its
+  ancestors</dfn> if the following algorithm returns `true`:
+
+1.  If |settings|'s [=relevant global object=] has no [=associated Document=],
+    return `false`.
+1.  Let |document| be |settings|' [=relevant global object=]'s [=associated Document=].
+1.  If |document| has no [=Document/browsing context=], return `false`.
+1.  Let |origin| be |settings|' [=environment settings object/origin=].
+1.  Let |navigable| be |document|'s [=node navigable=].
+1.  While |navigable| has a non-null [=navigable/parent=]:
+    1.  Set |navigable| to |navigable|'s [=navigable/parent=].
+    1.  If |navigable|'s [=active document=]'s [=Document/origin=] is not
+        [=/same site=] with |origin|, return `false`.
+1.  Return `true`.
+
+</div>
+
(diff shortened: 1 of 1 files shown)


npm1:
Maybe edit your description so that is actually links the issue?

cbiesinger:
> Maybe edit your description so that is actually links the issue?

Could've sworn I already did that... anyway, done now.

cbiesinger:
Ben, does this look reasonable to you?

cbiesinger:
@bvandersloot-mozilla , does this look reasonable to you?

spec/index.bs
```diff
@@ -366,14 +386,15 @@ be the result of [=get a structured field value=] from the response's header
...
 process this header as follows:
 
+
 <div algorithm="process the login status header">
 1. Let |origin| be the response's [=response/URL=]'s [=/origin=].
 1. Let |client| be the [=/request=]'s [=request/client=].
 1. If the request's [=request/destination=] is not `"document"`:
     1. If |client| is null, return.
-    1. If |origin| is not [=same origin=] with the [=/request=]'s
+    1. If |origin| is not [=/same site=] with the [=/request=]'s
```
timcappalli, line 395: what is the significance of `/` in the dfn syntax? I don't see any reference in the bikeshed docs.

cbiesinger: See https://speced.github.io/bikeshed/#autolink-inside -- it links to a dfn that has no for attribute (here, used to disambiguate what it links to)

bvandersloot-mozilla (approved):
Overall this is reasonable due to the scoping of cookies to site anyway.

I wonder if we should consider using a permission policy rather than a strict ancestor check though.

npm1:
> Overall this is reasonable due to the scoping of cookies to site anyway.

Thanks for the review! Will merge this PR once the repo is migrated to the WG.

> I wonder if we should consider using a permission policy rather than a strict ancestor check though.

It would not be very useful for the IdP to be allowed to change its login status if it does not have cookie access, hence why we did not consider adding permissions policy.

PR #551, opened 2024-03-19
Update HOWTO-chrome.md with button flow dev trial instruction

yi-gu:


explorations/HOWTO-chrome.md
```diff
@@ -333,3 +333,67 @@ IdP fails by either returning some network error or saying that the disconnectio
...
+Sec-Fetch-Dest: webidentity
+
+account_id=123&client_id=client1234&nonce=Ct60bD&disclosure_text_shown=true&is_auto_selected=false
+&mode=button
+```
+Note that we currently only include the new parameter in the button flow. Other flow
+types such as "widget" (name TBD) will be added when Chrome ships this feature.
+
+#### Use Other Account API
+This API allows users to “use other accounts” in the account chooser, for example, when IdPs
```
TallTed, line 381:
```suggestion
This API allows users to “use other accounts” in the account chooser when, for example, IdPs
```

npm1, line 381: Remove quotes

explorations/HOWTO-chrome.md
```diff
@@ -333,3 +333,67 @@ IdP fails by either returning some network error or saying that the disconnectio
...
+  "accounts_endpoint" : ...,
+  "modes: {
+    "button": {
+      "supports_use_other_account": true|false,
+    }
+  }
+}
+```
+
+Note: IdP can add "widget" in "modes" to support this feature in the non-button flow. However,
```
npm1, line 398: But does it actually work? If not, then mention that it does not work

explorations/HOWTO-chrome.md
yi-gu: done

explorations/HOWTO-chrome.md
yi-gu: It works. One can trigger the "use other account" row in the widget flow if they specify it in the config file.

yi-gu:
PTAL. Thanks!

explorations/HOWTO-chrome.md
```diff
@@ -333,3 +333,67 @@ IdP fails by either returning some network error or saying that the disconnectio
 was unsuccessful, or if the `account_id` is nowhere to be found, the browser will
 remove from local storage all of the federated accounts associated with the (RP,
 IDP).
+
+### Button Flow
+The button flow differs from the widget flow in several ways. The most significant
+difference is that the button flow is on the critical path of a user's sign-in
```
npm1, line 339: This seems speculative.

explorations/HOWTO-chrome.md
npm1: So it works with the chrome://flags but it won't work with the OT token? Does this mean we need a code change for this? We should probably not add this to the instructions if it is just missing in the code.

yi-gu: IMO we don't need a code change for this. This instruction is for dev trialing the "Use Other Account" API so that we should keep the function. Agreed that we can remove this note.

explorations/HOWTO-chrome.md
yi-gu: Thanks! Updated.

yi-gu (commented):
Thanks! PTAL.

npm1 (approved):

PR #574, opened 2024-05-08
Add sections for the continuation API bundle to the HOWTO

cbiesinger:
I am not adding the scopes API yet because of the discussions around what it should look like.

### DIFF
--- a/explorations/HOWTO-chrome.md
+++ b/explorations/HOWTO-chrome.md
@@ -396,7 +396,7 @@ To use the Use Other Account API:
 ```
 
 ### Continuation API
-This API lets the IdP request that the login-to-RP flow should continue
+This API lets the IdP request that the authorization flow should continue
 in a popup window that is controlled by the IdP. This can be used to request
 additional permission, to ask a user to confirm their account details, or
 for a variety of other use cases.
@@ -410,7 +410,7 @@ To use this feature:
     "continue_on": "https://idp.example/finish_login?account_id=123"
   }
   ```
-* When the login flow finishes, call `IdentityProvider.resolve` to close the
+* When the authorization flow finishes, call `IdentityProvider.resolve` to close the
   popup and provide the token that will be passed to the RP:
   ```js
     IdentityProvider.resolve("this is the token");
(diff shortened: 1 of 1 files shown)


explorations/HOWTO-chrome.md
```diff
@@ -394,3 +394,103 @@ To use the Use Other Account API:
   }
 }
 ```
+
+### Continuation API
+This API lets the IdP request that the login-to-RP flow should continue
```
samuelgoto, line 399:
```suggestion
This API lets the IdP request that the authorization flow should continue
```

explorations/HOWTO-chrome.md
```diff
@@ -394,3 +394,103 @@ To use the Use Other Account API:
...
+To use this feature:
+* Enable the experimental feature `FedCmAuthz` in chrome://flags
+* Return a "continue_on" field with a URL instead of a token
+  from the ID assertion endpoint. For example:
+  ```js
+  {
+    "continue_on": "https://idp.example/finish_login?account_id=123"
+  }
+  ```
+* When the login flow finishes, call `IdentityProvider.resolve` to close the
```
samuelgoto, line 413:
```suggestion
* When the authorization flow finishes, call `IdentityProvider.resolve` to close the
```

explorations/HOWTO-chrome.md
```diff
@@ -394,3 +394,103 @@ To use the Use Other Account API:
...
+* Add a `params` field to the `navigator.credentials.get` call:
+  ```js
+    navigator.credentials.get({
+        identity: {
+          providers: [{
+            configURL: "https://idp.example/config.json",
+            clientId: "123",
+            nonce: nonce,
+            params: {
+              "key": "value"
```
samuelgoto, line 444:
```suggestion
              "key": "value",
              "anything_goes": "yes",
              "really": "yes",
              "scopes": "calendar.readonly",
              "dpop": "something",
              "moar": "sure",
```

Just expanding with a few more examples.

explorations/HOWTO-chrome.md
```diff
@@ -394,3 +394,103 @@ To use the Use Other Account API:
...
+            }
+          }],
+        }
+    });
+  ```
+* These key/value pairs will be sent as-is in the ID assertion request:
+  `key=value&account_id=123&...`
```
samuelgoto, line 451:
```suggestion
  `key=value&account_id=123&...`
```

OTs are ok to be backwards incompatible, but if we could prefix these with `param-` we could avoid a backwards incompatibility problem later.

explorations/HOWTO-chrome.md
```diff
@@ -394,3 +394,103 @@ To use the Use Other Account API:
...
+  
+
+### Scaling .well-known
```
samuelgoto, line 454:
```suggestion
### Multiple configURLs
```

Maybe we should call it this rather than "scaling well-known"?

explorations/HOWTO-chrome.md
cbiesinger: the suggested change looks identical to the existing text to me?

Let me keep this as-is to match what is implemented. When we change the code I'll update this howto.

cbiesinger (commented):
Thanks, I accepted most suggestions but see comment below.

samuelgoto (approved):

PR #610, opened 2024-05-30
Specify the webdriver command to open terms of service and privacy po…

cbiesinger:
…licy

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2297,7 +2297,32 @@ The [=remote end steps=] are:
 1. Let |dialogButton| be the result of [=getting a property=] named
     `"dialogButton"` from |parameters|.
 
-1. If |dialogButton| is not a string that is "`ConfirmIdpLoginContinue`", return
+1. If |dialogButton| is not a string, return
+    a [=error|WebDriver error=] with [=error code=] [=invalid argument=].
+
+1. If |dialogButton| is "`TermsOfService`" or "`PrivacyPolicy`":
+    1. If no FedCM dialog is currently open or the dialog is not a dialog that
+        lets a user select an account, return a [=error|WebDriver error=] with
+        [=error code=] [=no such alert=].
+    1. Let |index| be the result of [=getting a property=] named
+        `"index"` from |parameters|.
+    1. If |index| is not an integer, or is less than zero or greater than
+        or equal to the number of accounts that the user can choose from in the
+        current flow, return a [=error|WebDriver error=] with [=error code=]
+        [=invalid argument=].
+    1. If the account at |index| does not have a terms of service or privacy
+        policy URL, either because none was provided in the [=client metadata
+        endpoint=] or because [=compute the connection status=] for the account
+        returned [=compute the connection status/connected=], return a
+        [=error|WebDriver error=] with [=error code=] [=invalid argument=].
+    1. Otherwise, if |dialogButton| is "`TermsOfService`", open the account's
+        {{IdentityProviderAccount/terms_of_service_url}}, or if |dialogButton|
+        is "`PrivacyPolicy`", open the account's
+        {{IdentityProviderAccount/privacy_policy_url}}, as if the user had
+        clicked the corresponding link in the account.
+    1. Return [=success=] with data `null`.
+
+1. If |dialogButton| is not "`ConfirmIdpLoginContinue`", return
     a [=error|WebDriver error=] with [=error code=] [=invalid argument=].
 
 1. If no FedCM dialog is currently open or the dialog is not a [=confirm IDP


spec/index.bs
```diff
@@ -2297,7 +2297,32 @@ The [=remote end steps=] are:
...
     `"dialogButton"` from |parameters|.
 
-1. If |dialogButton| is not a string that is "`ConfirmIdpLoginContinue`", return
+1. If |dialogButton| is not a string, return
+    a [=error|WebDriver error=] with [=error code=] [=invalid argument=].
+
+1. If |dialogButton| is "`TermsOfService`" or "`PrivacyPolicy`":
+    1. If no FedCM dialog is currently open or the dialog is not a dialog that
+        lets a user select an account, return a [=error|WebDriver error=] with
+        [=error code=] [=no such alert=].
```
npm1, line 2306: Is this the right error code?

spec/index.bs
```diff
@@ -2297,7 +2297,32 @@ The [=remote end steps=] are:
...
+        `"index"` from |parameters|.
+    1. If |index| is not an integer, or is less than zero or greater than
+        or equal to the number of accounts that the user can choose from in the
+        current flow, return a [=error|WebDriver error=] with [=error code=]
+        [=invalid argument=].
+    1. If the account at |index| does not have a terms of service or privacy
+        policy URL, either because none was provided in the [=client metadata
+        endpoint=] or because [=compute the connection status=] for the account
+        returned [=compute the connection status/connected=], return a
+        [=error|WebDriver error=] with [=error code=] [=invalid argument=].
```
npm1, line 2317: Can this be something different from the out of bounds case, so the developer can know what case this is based on the error?

npm1 (commented):
Looks good!

spec/index.bs
cbiesinger: It's the code I've been using for all the webdriver commands. It seemed close enough and I wanted to avoid defining a new error code.

spec/index.bs
cbiesinger: I am not sure there's a better error code in webdriver right now (and I don't think they are extensible)... in terms of our implementation, we also return a string error which differentiates between the two cases.

Although maybe I could use "invalid element state" here? let me know what you think

cbiesinger: hmm looking at CDP this could be a bit tricky to implement if we wanted to switch to invalid element state

npm1: Could it be `invalid selector`?

cbiesinger: Yeah I think what I'd do is:
- change CDP to return a boolean from the method
- change webdriver to return whichever error we want if the method returned false

Either invalid selector or invalid element state would work fine for that. (the problem I was referring to was that CDP does not support invalid element state/invalid selector directly)

npm1: I wonder what we do in existing scenarios where any of those errors are supposed to be surfaced? Since those error codes already exist, surely we would not be the first ones to use them? Anyways lgtm modulo figuring out the best error code to surface here

npm1 (approved):

samuelgoto (approved):
LGTM++

spec/index.bs

npm1 (approved):
Not sure what was the recent change but lgtm

cbiesinger:
The change was reformatting and also changing the error to invalid selector (from invalid argument) here:
```
    1. If the account at |index| does not have a terms of service or privacy
        policy URL, either because none was provided in the [=client metadata
        endpoint=] or because [=compute the connection status=] for the account
        returned [=compute the connection status/connected=], return a
        [=error|WebDriver error=] with [=error code=] [=invalid selector=].
```

cbiesinger:
@bvandersloot-mozilla could you take a look?

kyleant3425 (approved):

spec/index.bs
```diff
@@ -2191,7 +2191,32 @@ The [=remote end steps=] are:
...
+1. If |dialogButton| is "`TermsOfService`" or "`PrivacyPolicy`":
+    1. If no FedCM dialog is currently open or the dialog is not a dialog that
+        lets a user select an account, return a [=error|WebDriver error=] with
+        [=error code=] [=no such alert=].
+    1. Let |index| be the result of [=getting a property=] named
+        `"index"` from |parameters|.
+    1. If |index| is not an integer, or is less than zero or greater than
+        or equal to the number of accounts that the user can choose from in the
+        current flow, return a [=error|WebDriver error=] with [=error code=]
+        [=invalid argument=].
```
TallTed, line 2206:
```suggestion
    1. If |index| is not an integer, or is less than zero, or is greater than
        or equal to the number of accounts that the user can choose from in the
        current flow, then return a [=error|WebDriver error=] with [=error code=]
        [=invalid argument=].
```

spec/index.bs
```diff
@@ -2191,7 +2191,32 @@ The [=remote end steps=] are:
...
+    1. If the account at |index| does not have a terms of service or privacy
+        policy URL, either because none was provided in the [=client metadata
+        endpoint=] or because [=compute the connection status=] for the account
+        returned [=compute the connection status/connected=], return a
+        [=error|WebDriver error=] with [=error code=] [=invalid selector=].
+    1. Otherwise, if |dialogButton| is "`TermsOfService`", open the account's
+        {{IdentityProviderClientMetadata/terms_of_service_url}}, or if
+        |dialogButton| is "`PrivacyPolicy`", open the account's
```
TallTed, line 2214:
```suggestion
    1. If the account at |index| does not have a terms of service or privacy
        policy URL, either because none was provided in the [=client metadata
        endpoint=] or because [=compute the connection status=] for the account
        returned [=compute the connection status/connected=], then return a
        [=error|WebDriver error=] with [=error code=] [=invalid selector=].
    1. Otherwise, if |dialogButton| is "`TermsOfService`", open the account's
        {{IdentityProviderClientMetadata/terms_of_service_url}}, or if
        |dialogButton| is "`PrivacyPolicy`", then open the account's
```

TallTed (changes requested):
Multiple "if" conditions deserve the clarity of a "then"

PR #634, opened 2024-08-16
FPWD preparation

simoneonofri:
as https://github.com/w3c/transitions/issues/647

 - update index.bs with metadata
 - update SVGs
 - update w3c.json


npm1:
What are the svg updates? I don't see a noticeable difference in the images

spec/index.bs
```diff
@@ -1,13 +1,15 @@
...
-Shortname: FedCM
+Shortname: fedcm
 Level: 1
-Status: w3c/CG-DRAFT
-Group: fedidcg
-ED: https://fedidcg.github.io/FedCM/
-Repository: fedidcg/FedCM
-Editor: Nicolás Peña Moreno, Google Inc. https://google.com, npm@chromium.org
-Former Editor: Sam Goto, Google Inc. https://google.com, goto@google.com
+Status: w3c/FPWD
```
npm1, line 5: I didn't know FPWD is a thing? When would this change to WD?

spec/index.bs
```diff
@@ -1,13 +1,15 @@
...
+Group: fedid
+TR: http://www.w3.org/TR/fedcm/
```
npm1, line 7: This does not exist currently

simoneonofri:
some attributes added by the editor and the image ids.

this happened because  they are embedded in the HTML

spec/index.bs
simoneonofri: FPWD is a First Public Working Draft :)

spec/index.bs
simoneonofri: waiting for an ack to create it

npm1 (approved):

samuelgoto (approved):
LGTM++

PR #637, opened 2024-08-19
[link-validator] Missing file context-api.md

simoneonofri:
Hello,

There is a note with a broken reference link.

`Note: This command lets automation verify that the
   [context api](https://github.com/fedidcg/FedCM/blob/main/proposals/context-api.md)
    was applied properly`

I searched inside the repo and could not find it.

I am proposing to remove it, if it not exist anymore. Otherwise feel free to modify the PR.

Thanks,

Simone

P.S. This should be the last check for publication in `/TR`

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2412,10 +2412,6 @@ The [=remote end steps=] are:
     </table>
 </figure>
 
-Note: This command lets automation verify that the
-    [context api](https://github.com/fedidcg/FedCM/blob/main/proposals/context-api.md)
-    was applied properly
-
 The [=remote end steps=] are:
 
 1. If no FedCM dialog is currently open, return a [=error|WebDriver error=] with
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2412,6 +2412,10 @@ The [=remote end steps=] are:
     </table>
 </figure>
 
+ Note: This command lets automation verify that the
+  [context api](#dom-identitycredentialrequestoptions-context)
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -2412,10 +2412,6 @@ The [=remote end steps=] are:
     </table>
 </figure>
 
-Note: This command lets automation verify that the
```
npm1, line 2415: The correct solution is to point to https://w3c-fedid.github.io/FedCM/#dom-identitycredentialrequestoptions-context, not remove the note.

simoneonofri: ok i edited the code to have the correct link and updated also the page on TR

npm1: It looks like PR preview is no longer working on the repo. I think this link works but there is no preview to double check

spec/index.bs
```diff
@@ -2412,9 +2412,9 @@ The [=remote end steps=] are:
...
-    [context api](https://github.com/fedidcg/FedCM/blob/main/proposals/context-api.md)
-    was applied properly
+ Note: This command lets automation verify that the
+  [context api](#dom-identitycredentialrequestoptions-context)
+     was applied properly
```
npm1, line 2417: why the extra space here?

npm1 (approved):

spec/index.bs
simoneonofri: @npm1 none specifically, removed

Issue #640, opened 2024-08-22
Add in status indicator extension for Login Status


pascoej:
It can be useful to know the username of the account whenever sites use `setStatus.` This was present in the fedidcg explainer, but hasn't made it into the spec here yet.

I'm creating this issue because I'd like to consider it for inclusion and plan on opening a PR to allow sites to pass indicator.name and indicator.picture when setting login status.

Example from explainer:
```
navigator.login.setStatus("logged-in", {
  indicator: {
    name: "John Doe",
    picture: "https://website.com/john-doe/profile.png",
  }
});
```

samuelgoto:
I am highly supportive and excited about login status indicators, and I'm glad that you are looking at that too!

Some of the early explorations we looked at was to reuse parts of the url bar to indicate that the user is logged in.

Does this match more or less your intuition as to where we would hang a status indicator (obviously, each browser being able to make their own UX determinations independently)?

<img width="688" alt="Screenshot 2024-08-22 at 15 40 23" src="https://github.com/user-attachments/assets/59a25e5f-ef4d-4c7d-8e33-c9484edaa95e">

PR #642, opened 2024-08-22
Break the Login Status API out of the FedCM spec

### LINKED ISSUE #641: Break the Login Status API spec out of the FedCM spec
Currently, the [Login Status API](https://w3c-fedid.github.io/FedCM/#browser-api-login-status) is bundled inside a section of the FedCM spec, but that's entirely unnecessary: a browser vendor may choose to implement the former but not the latter (but not the other way around).

We should create a spec for the Login Status API that's separate from the FedCM spec so that they can move and be implemented independently.

samuelgoto:
Resolves w3c-fedid/FedCM#641.

### DIFF
--- a/spec/Makefile
+++ b/spec/Makefile
@@ -2,16 +2,24 @@
 
 all: spec idl
 
-spec: index.html
+spec: index.html login-status.html
 
-idl: fedcm.idl
+idl: fedcm.idl login-status.idl
 
 fedcm.idl: index.bs
 	./extract_idl.rb fedcm.idl
 
+login-status.idl: login-status.bs
+	./extract_idl.rb login-status.idl
+
 index.html: index.bs
 	bikeshed --die-on=fatal spec index.bs
+
+login-status.html: login-status.bs
+	bikeshed --die-on=fatal spec login-status.bs
 
 clean:
 	rm index.html
 	rm fedcm.idl
+	rm login-status.html
+	rm login-status.idl
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -46,6 +46,13 @@ spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
 spec: webappsec-fetch-metadata; urlPrefix: https://w3c.github.io/webappsec-fetch-metadata/
     type: dfn
         text: Directly User-Initiated Requests; url: directly-user-initiated
+spec: login-status; urlPrefix: login-status.html
+    type: dfn
+        text: logged-in; url: logged-in
+        text: logged-out; url: logged-out
+        text: unknown; url: unknown
+        text: get the login status; url: get-the-login-status
+        text: set the login status; url: set-the-login-status
 </pre>
 
 <pre class=link-defaults>
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -331,130 +331,6 @@ for an [=IDP=] to adopt the FedCM API. It doesn't introduce security issues on t
 [=RP=] cannot inspect the results from the fetches in any way.
 
 <!-- ============================================================ -->
-## The Login Status API ## {#browser-api-login-status}
-<!-- ============================================================ -->
-
-### Login Status Map ### {#hdr-login-status-map}
-
-Each [=user agent=] keeps a global, persistent <dfn>Login Status
-map</dfn>, an initially empty [=map=]. The [=map/keys=] in this map are
-[=/origin=] (of [=IDPs=]), and the [=map/values=] are enums that can be one of
-"<dfn><code>unknown</code></dfn>", "<dfn><code>logged-in</code></dfn>",
-and "<dfn><code>logged-out</code></dfn>".
-
-<div algorithm>
-To <dfn>get the login status</dfn> for an [=/origin=] |origin|:
-1. If [=Login Status map=][|origin|] exists, return it.
-1. Otherwise, return [=unknown=].
-
-</div>
-
-<div algorithm>
-To <dfn>set the login status</dfn> for an [=/origin=] |origin| to
-value |value|:
-1. Assert that |value| is one of [=logged-in=] or [=logged-out=].
-1. [=map/Set=] [=Login Status map=][|origin|] to |value|.
-
-</div>
-
-### HTTP header API ### {#login-status-http}
-
-[=IDPs=] can set the login status using an HTTP [=response=] [=header=] as follows.
-
-Issue: The HTTP header checking should move into the Fetch spec, since it
-    affects all resource loads.
-
-For each [=http-redirect fetch=] and [=http fetch=]'s [=response=], let |value|
-be the result of [=get a structured field value=] from the response's header
-list with name "<dfn><code>Set-Login</code></dfn>" and type "`item`". If |value| is not null,
-process this header as follows:
-
-<div algorithm="process the login status header">
-1. Let |origin| be the response's [=response/URL=]'s [=/origin=].
-1. Let |client| be the [=/request=]'s [=request/client=].
-1. If the request's [=request/destination=] is not `"document"`:
-    1. If |client| is null, return.
-    1. If |origin| is not [=same origin=] with the [=/request=]'s
-        [=request/origin=], return.
-    1. If |client| is not [=same-origin with its ancestors=], return.
-1. Assert that |value| is a tuple.
-1. Let |token| be the first entry of |value|.
-1. If |token| is `"logged-in"`, [=set the login status=] for |origin|
-    to [=logged-in=].
-1. If |token| is `"logged-out"`, [=set the login status=] for |origin|
-    to [=logged-out=].
-
-</div>
-
-### JavaScript API ### {#login-status-javascript}
-
-[=IDPs=] can also use a JavaScript API to update the stored login status:
-
-
-<pre class="idl">
-enum LoginStatus {
-  "logged-in",
-  "logged-out",
-};
-
-[Exposed=Window, SecureContext] 
-interface NavigatorLogin {
-  Promise&lt;undefined&gt; setStatus(LoginStatus status);
-};
-
-partial interface Navigator {
-  [SecureContext] readonly attribute NavigatorLogin login;
-};
-</pre>
-
-<div algorithm="setStatus">
-When {{NavigatorLogin/setStatus()}} is called with argument |status|:
-1. If the [=current settings object=] is not [=same-origin with its ancestors=],
-    throw a {{SecurityError}} {{DOMException}}.
-1. Let |origin| be the [=current settings object=]'s
-    [=environment settings object/origin=].
-1. Let |value| be [=logged-in=] if |status| is `"logged-in"` or [=logged-out=]
-    if |status| is `"logged-out"`.
-1. [=Set the login status=] for |origin| to |value|.
-
-</div>
-
-### Clearing the Login Status Map data ### {#login-status-clear-data}
-
-User agents MUST also clear the [=Login Status map=] data when:
-    :   the user clears all cookies or site settings data
-    ::  The user agent MUST clear the entire map.
-    :   the user clears all cookies or all site data for a specific origin
-    ::  The user agent MUST remove all entries that would be affected
-            by the deleted cookies, that is, any entry with an origin
-            to which a deleted cookie could be sent to.
-
-            Note: For example, domain cookies may affect subdomains of
-                the deleted origin, e.g. clearing cookies for `google.com`
-                should also reset the login status for `accounts.google.com`,
-                since it may rely on a domain cookie for google.com.
-    :   the user deletes individual cookies (if allowed by the user agent)
-    ::  the behavior is user agent-defined.
-
-            Note: The user agent MAY want to reset the state to [=unknown=],
-                since is impossible to know whether this cookie affects
-                authorization state.
-    : the user agent receives a <a http-header>Clear-Site-Data</a> header with a
-        value of `"cookies"` or `"*"`, and the [=/request=]'s [=request/client=] is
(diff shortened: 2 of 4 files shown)


npm1:
Uhm is this a request from someone? Also I'm not sure but I imagine there needs to be some procedure to add a new spec to the WG (even if it is all extracted from the current spec). Finally, in terms of editor I would suggest to name the one who actually wrote this which I think is @cbiesinger

cbiesinger:
> Also I'm not sure but I imagine there needs to be some procedure to add a new spec to the WG (even if it is all extracted from the current spec).

As far as I know the primary requirement is that the spec is in the charter, which it is (https://www.w3.org/2024/03/wg-fedid-charter.html#deliverables)

I see this was also added to the agenda for the next call

spec/login-status.bs
```diff
@@ -0,0 +1,264 @@
...
+spec: ecma262; urlPrefix: https://tc39.github.io/ecma262/
+    type: dfn
+        text: internal method; url: sec-ordinary-object-internal-methods-and-internal-slots
+
+spec: credential-management-1; urlPrefix: https://w3c.github.io/webappsec-credential-management/
+    type: dfn
+        text: same-origin with its ancestors; url: same-origin-with-its-ancestors
+    type: dfn
+        text: requires user mediation; url: origin-requires-user-mediation
+spec: webdriver; urlPrefix: https://w3c.github.io/webdriver/
```
cbiesinger, line 36: A lot of these spec references are not needed for login status

spec/login-status.bs
```diff
@@ -0,0 +1,264 @@
...
+Title: Login Status API
+Shortname: login-status
+Level: 1
+Status: w3c/FPWD
+Group: fedid
+TR: http://www.w3.org/TR/fedcm/login-status.html
+ED: https://w3c-fedid.github.io/fedcm/login-status.html
+Prepare for TR: true
+Repository: w3c-fedid/FedCM
+Editor: Christian Biesinger, w3cid 50308, Google Inc. https://google.com, cbiesinger@chromium.org
```
cbiesinger, line 11: Apparently my w3cid is 39461

samuelgoto:
> Uhm is this a request from someone?

Yeah, we have always intended to break the Login Status spec out of the FedCM spec, but never got to it: a browser vendor should be able to implement the Login Status API without having to implement the FedCM API.

https://github.com/privacycg/is-logged-in/issues/55#issuecomment-1749419249
https://www.w3.org/2023/09/13-login-status-minutes.html
https://www.w3.org/2024/03/wg-fedid-charter.html#deliverables (separate and orthogonal deliverables)

and more recently:

https://github.com/w3c-fedid/FedCM/issues/640#issue-2481927370

> Also I'm not sure but I imagine there needs to be some procedure to add a new spec to the WG (even if it is all extracted from the current spec)

I'm already working with @hlflanagan and @wseltzer to figure out how to go about this.  We would likely want to also move the spec to its own repo and the login-status repo from the CG to the WG:

https://github.com/fedidcg/login-status

We also want to coordinate with @johnwilander if he would rather have us use the FedID CG repo or the Privacy CG repo as the basis for the FedID WG repo (e.g. keeping the history and the issues). Either works for me, but we can have that discussion separately.

> Finally, in terms of editor I would suggest to name the one who actually wrote this which I think is @cbiesinger

Yep, chatted with @cbiesinger who agreed to edit here too. Done.

We also want to coordinate with @johnwilander and @bvandersloot-mozilla if they still feel comfortable sharing co-editorship here (TPAC 2023 we had [the three of us as co-editors](https://github.com/fedidcg/login-status)).

spec/login-status.bs
samuelgoto: ha, good catch, done.

spec/login-status.bs
```diff
@@ -0,0 +1,264 @@
...
+        text: no such alert; url: dfn-no-such-alert
+        text: error code; url: dfn-error-code
+        text: validating capabilities; url: dfn-validate-capabilities
+spec: webappsec-fetch-metadata; urlPrefix: https://w3c.github.io/webappsec-fetch-metadata/
+    type: dfn
+        text: Directly User-Initiated Requests; url: directly-user-initiated
+</pre>
+
+<pre class=link-defaults>
+spec:infra; type:dfn; text:list
```
npm1, line 51: Remove the unused.

spec/login-status.bs
```diff
@@ -0,0 +1,264 @@
...
+spec:infra; type:dfn; text:user agent
+spec:html; type:dfn; for:environment settings object; text:global object
+spec:html; type:dfn; for:html-origin-def; text:origin
+spec:webidl; type:dfn; text:resolve
+spec:webdriver2; type:dfn; text:error
+spec:fetch; type:dfn; for:/; text:response
+</pre>
+
+<style>
```
npm1, line 60: I think we can remove this altogether?

spec/login-status.bs
```diff
@@ -0,0 +1,264 @@
...
+   to support dark mode properly */
+script + svg :is(polygon, line, rect):not(.grid) {
+  stroke: currentcolor;
+}
+script + svg :is(polygon, text) {
+  fill: currentcolor;
+}
+</style>
+
+<script src="https://fedidcg.github.io/FedCM/static/underscore-min.js"></script>
```
npm1, line 90: Remove?

spec/login-status.bs
```diff
@@ -0,0 +1,264 @@
...
+<script src="https://fedidcg.github.io/FedCM/static/webfont.js"></script>
+<script src="https://fedidcg.github.io/FedCM/static/typogram.js"></script>
+
+<!-- ============================================================ -->
+# Introduction # {#introduction}
+<!-- ============================================================ -->
+
+*This section is non-normative.*
+
+Note: use the introduction from the explainer.
```
npm1, line 101: ?

spec/login-status.bs
```diff
@@ -0,0 +1,264 @@
...
+
+<!-- ============================================================ -->
+# The Login Status API # {#browser-api-login-status}
```
npm1, line 104: This does not make sense as a title anymore because this is the title of the spec.

spec/login-status.bs
```diff
@@ -0,0 +1,264 @@
...
+<!-- ============================================================ -->
+
+## Login Status Map ## {#hdr-login-status-map}
```
npm1, line 107: Hence lower the `#` for all titles I think. Also the new 1`#` titles need the separators

spec/login-status.bs
```diff
@@ -0,0 +1,264 @@
...
+process this header as follows:
+
+<div algorithm="process the login status header">
+1. Let |origin| be the response's [=response/URL=]'s [=/origin=].
+1. Let |client| be the [=/request=]'s [=request/client=].
+1. If the request's [=request/destination=] is not `"document"`:
+    1. If |client| is null, return.
+    1. If |origin| is not [=same origin=] with the [=/request=]'s
+        [=request/origin=], return.
+    1. If |client| is not [=same-origin with its ancestors=], return.
```
npm1, line 149: You seem to have copy pasted an outdated version. We updated this to same-site...

spec/index.bs
```diff
@@ -330,151 +330,6 @@ value (which is used when a resource loaded as a third-party, not first-party).
...
-
-<div algorithm>
-To <dfn>set the login status</dfn> for an [=/origin=] |origin| to
-value |value|:
-1. Assert that |value| is one of [=logged-in=] or [=logged-out=].
-1. [=map/Set=] [=Login Status map=][|origin|] to |value|.
-
-</div>
-
-### Infrastructure algorithm ### {#infra-algorithm}
```
npm1, line 360: Oh this is not a good title. We may want to rename it once we move it

spec/login-status.bs
```diff
@@ -0,0 +1,264 @@
...
+    [=IDP=] login state changes, it should send an explicit [=Set-Login=] header.
+    [=RP=] state should not affect this map since it only reflects [=IDP=] state.
+
+<!-- ====================================================================== -->
+# Acknowledgements # {#acknowledgements}
+<!-- ====================================================================== -->
+
+Note: write down the Acknowledgements section.
+
+<pre class="biblio">
```
npm1, line 233: Remove unneeded?

spec/login-status.bs
```diff
@@ -0,0 +1,264 @@
...
+
+Note: Other website-initiated cookie changes should not affect this map. When
+    [=IDP=] login state changes, it should send an explicit [=Set-Login=] header.
+    [=RP=] state should not affect this map since it only reflects [=IDP=] state.
+
+<!-- ====================================================================== -->
+# Acknowledgements # {#acknowledgements}
+<!-- ====================================================================== -->
+
+Note: write down the Acknowledgements section.
```
npm1, line 231: ?

spec/Makefile
```diff
@@ -2,16 +2,24 @@
 
 all: spec idl
 
-spec: index.html
+spec: index.html login-status.html
```
npm1, line 5: I imagine we will undo these changes if it moves to its own repo.

npm1 (changes requested):

spec/login-status.bs
samuelgoto: done

spec/login-status.bs
samuelgoto: done

spec/login-status.bs
samuelgoto: Are you sure? This seems to be up to date with HEAD?

https://github.com/w3c-fedid/FedCM/blob/main/spec/index.bs#L400

spec/login-status.bs
samuelgoto: Same todo as in the FedCM spec?

https://github.com/w3c-fedid/FedCM/blob/main/spec/index.bs#L2952

spec/login-status.bs
samuelgoto: done

spec/login-status.bs
samuelgoto: done

spec/login-status.bs
samuelgoto: done

spec/Makefile
samuelgoto: yep

spec/index.bs
samuelgoto: Sure. Lets do that on a separate pass to keep the history.

samuelgoto (commented):
Done

Also, added John Wilander as a co-editor.

spec/login-status.bs
npm1: That same same-site?

samuelgoto: Ah, yes, interesting, for some reason I missed this PR in my branch:

https://github.com/w3c-fedid/FedCM/commit/548e7b2a5fc990c7588ed2c9af46416c0f00be07

I sync and merged and I think this is now up to date. Sanity check?

npm1: Now it seems fine. I guess the plan is to create a new repo for this, right? So anyways the change in this repo will only be to remove the existing spec text.

samuelgoto: > . I guess the plan is to create a new repo for this, right? 

Yes

> So anyways the change in this repo will only be to remove the existing spec text.

Yes, that's correct. I wanted to move into a single PR so that it is clear that the break was clean, rather than in two separate PRs into different repos.

npm1 (approved):

spec/index.bs
```diff
@@ -1587,8 +1442,8 @@ success or failure.
                 with the [=/origin=] of the {{IdentityProviderAPIConfig/login_url}}.
 
                 Note: The IDP login flow may set this value to logged-in using
-                    either the [[#login-status-javascript|JavaScript]] or
-                    [[#login-status-http|HTTP header]] API during the login
+                    either the [login-status.html#login-status-javascript](JavaScript) or
+                    [login-status.html#login-status-http](HTTP header) API during the login
```
cbiesinger, line 1446: I wonder if this should be a [==] reference... may need a <dfn export> in the login status spec

Also, don't we need a link to login-status in the anchors section at the top? As-is, there are a lot of failed links: https://github.com/w3c-fedid/FedCM/actions/runs/10690399462/job/29634680467#step:3:839

spec/login-status.bs
```diff
@@ -0,0 +1,207 @@
...
+
+Test Suite: https://github.com/web-platform-tests/wpt/tree/master/fedcm
+</pre>
+
+<pre class=link-defaults>
+spec:infra; type:dfn; text:user agent
+spec:fetch; type:dfn; for:/; text:response
+spec: credential-management-1; urlPrefix: https://w3c.github.io/webappsec-credential-management/
+    type: dfn
+        text: same-origin with its ancestors; url: same-origin-with-its-ancestors</pre>
```
cbiesinger, line 32: I don't think this is actually used?

spec/index.bs
samuelgoto: ah, i think that did the trick! PTAL!?

https://github.com/w3c-fedid/FedCM/pull/642/commits/57509ce9497acdc7857cdeff4a20c9c3ffc8778f

spec/login-status.bs
samuelgoto: I think that's still used here?

https://github.com/w3c-fedid/FedCM/blob/main/spec/index.bs#L840

cbiesinger: Yes, but it's not used in the login status spec?

cbiesinger (approved):
lgtm modulo the same-origin comment

spec/login-status.bs
samuelgoto: ah, right. done.

PR #660, opened 2024-10-02
Specify the mode API

tttzach:
Fixes https://github.com/w3c-fedid/active-mode/issues/2

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/pull/660.html" title="Last updated on Nov 1, 2024, 5:58 PM UTC (494dd9c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/660/721c213...494dd9c.html" title="Last updated on Nov 1, 2024, 5:58 PM UTC (494dd9c)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -797,7 +797,7 @@ the exception thrown.
 
         1. If |mode| is [=active=],
             [=fetch the config file=] and [=show an IDP login dialog=].
-        2. If |mode| is [=passive=], the user agent MUST do one of the following:
+        1. If |mode| is [=passive=], the user agent MUST do one of the following:
             * Return (failure, false).
             * Prompt the user whether to continue. If the user continues, the user
                 agent SHOULD set |loginStatus| to [=unknown=]. This MAY include an
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -804,7 +804,6 @@ the exception thrown.
                 * If the user cancels this dialog, return (failure, true).
                 * If the user triggers this affordance, [=fetch the config file and show an IDP login dialog=]
                    with |provider| and |globalObject|.
-        |provider| and |globalObject|.
     1. Let |requiresUserMediation| be |provider|'s {{IdentityProviderConfig/configURL}}'s [=/origin=]'s
         [=requires user mediation=].
     1. Let |mediation| be |options|'s {{CredentialRequestOptions/mediation}}.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -802,9 +802,9 @@ the exception thrown.
 
             * If the user cancels this dialog, return (failure, true).
             * If the user triggers this affordance,
-                [=fetch the config file and show an IDP login dialog=].
+                [=fetch the config file=] and [=show an IDP login dialog=].
         * If |mode| is [=active=],
-            [=fetch the config file and show an IDP login dialog=].
+            [=fetch the config file=] and [=show an IDP login dialog=].
 
     1. Let |requiresUserMediation| be |provider|'s {{IdentityProviderConfig/configURL}}'s [=/origin=]'s
         [=requires user mediation=].
@@ -894,7 +894,7 @@ the exception thrown.
         1. Set |account| to |accountsList|[0].
         1. If [=compute the connection status=] of |account|, |provider| and |globalObject| returns
             [=compute the connection status/connected=], show a dialog to request user permission to sign
-            in via |account|, and set the result in |permission|. The user agent MAY use |options|'s
+            in via |account|, and set the result in |permission|. The user agent SHOULD use |options|'s
             {{IdentityCredentialRequestOptions/context}} and |options|'s
             {{IdentityCredentialRequestOptions/mode}} to customize the dialog.
         1. Otherwise, let |permission| be the result of running [=request permission to sign-up=]
@@ -1302,7 +1302,7 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
             is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
             |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
             the |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
-        1. The user agent MAY use the
+        1. The user agent SHOULD use the
             {{IdentityCredentialRequestOptions/context}} and |options|'s
             {{IdentityCredentialRequestOptions/mode}} to customize the dialog shown.
     1. If the user does not grant permission, return false.
@@ -1469,18 +1469,6 @@ success or failure.
                     a different browsing context.
             1. If |loginStatus| is [=logged-in=], return success.
             1. Otherwise, return failure.
-</div>
-
-<div algorithm>
-To <dfn>fetch the config file and show an IDP login dialog</dfn> given an {{IdentityProviderConfig}} |provider|, and a |globalObject|, run the following steps. This returns
-success or failure.
-    1. Assert: these steps are running [=in parallel=].
-    1. Let |config| be the result of running [=fetch the config file=]
-              with |provider| and |globalObject|.
-    1. If |config| is failure, return failure.
-    1. [=Show an IDP login dialog=] with |config| and |provider|.
-    1. If that algorithm succeeds, return success.
-    1. Otherwise, return failure.
 </div>
 
 <!-- ============================================================ -->
@@ -2907,4 +2895,3 @@ Note: The WG has labeled the following issues as critical open issues that must 
 </ul>
 </body>
 </html>
-
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -624,7 +624,8 @@ The {{IdentityCredentialRequestOptions}} contains a list of
 {{IdentityProviderConfig}}s that the [=RP=] supports and has
 pre-registered with (i.e. the [=IDP=] has given the [=RP=] a `clientId`).
 The {{IdentityCredentialRequestOptions}} also contains a {{IdentityCredentialRequestOptionsContext}}
-which the user agent can use to provide a more meaningful dialog to users.
+which the user agent can use to provide a more meaningful dialog to users and {{IdentityCredentialRequestOptionsMode}}
+which the user agent can use to specify different behaviors or dialog types.
 
 <xmp class=idl>
 enum IdentityCredentialRequestOptionsContext {
@@ -634,9 +635,15 @@ enum IdentityCredentialRequestOptionsContext {
   "continue"
 };
 
+enum IdentityCredentialRequestOptionsMode {
+  "active",
+  "passive"
+};
+
 dictionary IdentityCredentialRequestOptions {
   required sequence<IdentityProviderRequestOptions> providers;
   IdentityCredentialRequestOptionsContext context = "signin";
+  IdentityCredentialRequestOptionsMode mode = "passive";
 };
 </xmp>
 
@@ -777,24 +784,27 @@ the exception thrown.
     1. Let |loginStatus| be the result of [=get the login status=] with
         the [=/origin=] of |provider|'s {{IdentityProviderConfig/configURL}}.
     1. If |loginStatus| is [=unknown=], a user agent MAY set it to [=logged-out=].
+    1. Let |showIdpLoginDialog| be an affordance to [=show an IDP login dialog=]:
+        1. Let |config| be the result of running [=fetch the config file=]
+                  with |provider| and |globalObject|.
+        1. If |config| is failure, return (failure, true).
+        1. [=Show an IDP login dialog=] with |config| and |provider|.
+        1. If that algorithm returns failure, return (failure, true).
     1. If |loginStatus| is [=logged-out=], the user agent MUST do one of the following:
 
         * Return (failure, false).
         * Prompt the user whether to continue. If the user continues, the user
-            agent SHOULD set |loginStatus| to [=unknown=]. This MAY include an
-            affordance to [=show an IDP login dialog=].
+            agent SHOULD set |loginStatus| to [=unknown=]. This MAY include |showIdpLoginDialog| affordance.
 
             * If the user cancels this dialog, return (failure, true).
-            * If the user triggers this affordance:
-                1. Let |config| be the result of running [=fetch the config file=]
-                    with |provider| and |globalObject|.
-                1. If |config| is failure, return (failure, true).
-                1. [=Show an IDP login dialog=] with |config| and |provider|.
-                1. If that algorithm returns failure, return (failure, true).
+            * If the user continues, trigger |showIdpLoginDialog|.
+        * Trigger |showIdpLoginDialog| directly. The user agent SHOULD set |loginStatus| to [=unknown=].
 
-        Issue: We should perhaps provide a way to let the [=RP=] request that
-            the second option is provided, possibly gated on a user gesture.
-            See [this issue](https://github.com/fedidcg/FedCM/issues/442) for discussion.
+            * Let |mode| be |options|'s {{IdentityCredentialRequestOptions/mode}}.
+            * If |mode| is [=passive=], return (failure, true).
+            * If |mode| is [=active=] and user gesture is not present, return (failure, true).
+            * If |mode| is [=active=] and user gesture is present, trigger |showIdpLoginDialog|.
+
     1. Let |requiresUserMediation| be |provider|'s {{IdentityProviderConfig/configURL}}'s [=/origin=]'s
         [=requires user mediation=].
     1. Let |mediation| be |options|'s {{CredentialRequestOptions/mediation}}.
@@ -884,7 +894,8 @@ the exception thrown.
         1. If [=compute the connection status=] of |account|, |provider| and |globalObject| returns
             [=compute the connection status/connected=], show a dialog to request user permission to sign
             in via |account|, and set the result in |permission|. The user agent MAY use |options|'s
-            {{IdentityCredentialRequestOptions/context}} to customize the dialog.
+            {{IdentityCredentialRequestOptions/context}} and |options|'s
+            {{IdentityCredentialRequestOptions/mode}} to customize the dialog.
         1. Otherwise, let |permission| be the result of running [=request permission to sign-up=]
             algorithm with |account|, |config|, |provider|, and |globalObject|. Also set
             |disclosureTextShown| to true.
@@ -1290,8 +1301,8 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
             is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
             |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
             the |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
-        1. The user agent MAY use the {{IdentityCredentialRequestOptions/context}} to customize the
-            dialog shown.
+        1. The user agent MAY use the {{IdentityCredentialRequestOptions/context}} and |options|'s
+            {{IdentityCredentialRequestOptions/mode}} to customize the dialog shown.
     1. If the user does not grant permission, return false.
     1. [=Create a connection between the RP and the IdP account=] with |provider|, |account|, and
         |globalObject|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -623,8 +623,8 @@ This specification introduces an extension to the {{CredentialRequestOptions}} o
 The {{IdentityCredentialRequestOptions}} contains a list of
 {{IdentityProviderConfig}}s that the [=RP=] supports and has
 pre-registered with (i.e. the [=IDP=] has given the [=RP=] a `clientId`).
-The {{IdentityCredentialRequestOptions}} also contains a {{IdentityCredentialRequestOptionsContext}}
-which the user agent can use to provide a more meaningful dialog to users and {{IdentityCredentialRequestOptionsMode}}
+The {{IdentityCredentialRequestOptions}} also contains an {{IdentityCredentialRequestOptionsContext}},
+which the user agent can use to provide a more meaningful dialog to users, and an {{IdentityCredentialRequestOptionsMode}},
 which the user agent can use to specify different behaviors or dialog types.
 
 <xmp class=idl>
@@ -781,29 +781,21 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
 or a pair (failure, bool), where the bool indicates whether to skip delaying
 the exception thrown.
     1. Assert: These steps are running [=in parallel=].
+    1. Let |mode| be |options|'s {{IdentityCredentialRequestOptions/mode}}.
+    1. If |mode| is [=active=]
+        1. If [=transient activation=] is not present, return (failure, true).
+        1. If [=transient activation=] is present and if there is a pending request where |mode| is [=passive=], return (failure, true) for the pending request.
     1. Let |loginStatus| be the result of [=get the login status=] with
         the [=/origin=] of |provider|'s {{IdentityProviderConfig/configURL}}.
     1. If |loginStatus| is [=unknown=], a user agent MAY set it to [=logged-out=].
-    1. Let |showIdpLoginDialog| be an affordance to [=show an IDP login dialog=]:
-        1. Let |config| be the result of running [=fetch the config file=]
-                  with |provider| and |globalObject|.
-        1. If |config| is failure, return (failure, true).
-        1. [=Show an IDP login dialog=] with |config| and |provider|.
-        1. If that algorithm returns failure, return (failure, true).
-    1. If |loginStatus| is [=logged-out=], the user agent MUST do one of the following:
+    1. If |loginStatus| is [=logged-out=], the user agent SHOULD set |loginStatus| to [=unknown=] and MUST do one of the following:
 
         * Return (failure, false).
-        * Prompt the user whether to continue. If the user continues, the user
-            agent SHOULD set |loginStatus| to [=unknown=]. This MAY include |showIdpLoginDialog| affordance.
+        * Prompt the user whether to continue. This MAY include having to [=allow an affordance to show an IDP login dialog=].
 
             * If the user cancels this dialog, return (failure, true).
-            * If the user continues, trigger |showIdpLoginDialog|.
-        * Trigger |showIdpLoginDialog| directly. The user agent SHOULD set |loginStatus| to [=unknown=].
-
-            * Let |mode| be |options|'s {{IdentityCredentialRequestOptions/mode}}.
-            * If |mode| is [=passive=], return (failure, true).
-            * If |mode| is [=active=] and user gesture is not present, return (failure, true).
-            * If |mode| is [=active=] and user gesture is present, trigger |showIdpLoginDialog|.
+            * If the user continues, trigger [=allow an affordance to show an IDP login dialog=].
+        * If |mode| is [=active=] and [=transient activation=] is present, trigger [=show an IDP login dialog=].
 
     1. Let |requiresUserMediation| be |provider|'s {{IdentityProviderConfig/configURL}}'s [=/origin=]'s
         [=requires user mediation=].
@@ -1467,6 +1459,18 @@ success or failure.
                     a different browsing context.
             1. If |loginStatus| is [=logged-in=], return success.
             1. Otherwise, return failure.
+</div>
+
+<div algorithm>
+To <dfn>allow an affordance to show an IDP login dialog</dfn> given an {{IdentityProviderConfig}} |provider|, and a |globalObject|, run the following steps. This returns
+success or failure.
+    1. Assert: these steps are running [=in parallel=].
+    1. Let |config| be the result of running [=fetch the config file=]
+              with |provider| and |globalObject|.
+    1. If |config| is failure, return failure.
+    1. [=Show an IDP login dialog=] with |config| and |provider|.
+    1. If that algorithm succeeds, return success.
+    1. Otherwise, return failure.
 </div>
 
 <!-- ============================================================ -->
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -799,7 +799,7 @@ the exception thrown.
             1. Let |result| be the result of running
                 [=fetch the config file and show an IDP login dialog=] with
                 |provider| and |globalObject|.
-            1. If |result| is failure, return (failure, false).
+            1. If |result| is failure, return (failure, true).
         1. Otherwise, the user agent MUST do one of the following:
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -777,24 +784,27 @@ the exception thrown.
     1. Let |loginStatus| be the result of [=get the login status=] with
         the [=/origin=] of |provider|'s {{IdentityProviderConfig/configURL}}.
     1. If |loginStatus| is [=unknown=], a user agent MAY set it to [=logged-out=].
+    1. Let |showIdpLoginDialog| be an affordance to [=show an IDP login dialog=]:
+        1. Let |config| be the result of running [=fetch the config file=]
+                  with |provider| and |globalObject|.
+        1. If |config| is failure, return (failure, true).
+        1. [=Show an IDP login dialog=] with |config| and |provider|.
```
samuelgoto, line 791: Don't you need to check somewhere if mode == "active" before calling [=Show an IDP login dialog=]?

tttzach: I was hoping for this to just be a definition of the |showIdpLoginDialog| behavior, the check for mode == "active" comes in line 806.

spec/index.bs
```diff
@@ -777,24 +784,27 @@ the exception thrown.
...
-                    with |provider| and |globalObject|.
-                1. If |config| is failure, return (failure, true).
-                1. [=Show an IDP login dialog=] with |config| and |provider|.
-                1. If that algorithm returns failure, return (failure, true).
-
-        Issue: We should perhaps provide a way to let the [=RP=] request that
-            the second option is provided, possibly gated on a user gesture.
-            See [this issue](https://github.com/fedidcg/FedCM/issues/442) for discussion.
+            * If the user continues, trigger |showIdpLoginDialog|.
+        * Trigger |showIdpLoginDialog| directly. The user agent SHOULD set |loginStatus| to [=unknown=].
```
cbiesinger, line 801: This needs to be written a bit differently. As written it will trigger the IDP login dialog twice (here and on line 806).

Either this line needs to be more informal ("Trigger an IDP login dialog by following these steps") or just describe directly what to do. Since we already allow returning failure above, this could just be something like:

```
* If |options|'s {{IdentityCredentialRequestOptions/mode}} is [=active=] and [=transient activation=] is present, trigger |showIdpLoginDialog|.
```

cbiesinger (commented):
Generally looks good. See comments below. We should probably add a Note: in some appropriate place that the user agent should consider displaying the account chooser in a louder way if mode is active.

spec/index.bs
```diff
@@ -624,7 +624,8 @@ The {{IdentityCredentialRequestOptions}} contains a list of
 {{IdentityProviderConfig}}s that the [=RP=] supports and has
 pre-registered with (i.e. the [=IDP=] has given the [=RP=] a `clientId`).
 The {{IdentityCredentialRequestOptions}} also contains a {{IdentityCredentialRequestOptionsContext}}
```
TallTed, line 626: This goes along with the suggestion below.
```suggestion
The {{IdentityCredentialRequestOptions}} also contains an {{IdentityCredentialRequestOptionsContext}},
```

spec/index.bs
```diff
@@ -624,7 +624,8 @@ The {{IdentityCredentialRequestOptions}} contains a list of
...
-which the user agent can use to provide a more meaningful dialog to users.
+which the user agent can use to provide a more meaningful dialog to users and {{IdentityCredentialRequestOptionsMode}}
+which the user agent can use to specify different behaviors or dialog types.
```
TallTed, line 628:
```suggestion
which the user agent can use to provide a more meaningful dialog to users,
and an {{IdentityCredentialRequestOptionsMode}},
which the user agent can use to specify different behaviors or dialog types.
```

TallTed (changes requested):
Minor, human-facing

spec/index.bs
```diff
@@ -777,24 +784,27 @@ the exception thrown.
     1. Let |loginStatus| be the result of [=get the login status=] with
```
yi-gu, line 784: I'm wondering if we should start the algorithm by checking the mode and user activation requirement. e.g. if there's no user activation when active mode is used, we don't need to update the login status. See https://github.com/yi-gu/FedCM/blob/main/explorations/static/FedCM_Mode_API.jpeg?raw=true

In addition, we should also mention an active mode must replace pending passive mode request unless people consider that as a UA specific choice.

spec/index.bs
tttzach: This is now an algorithm, thanks!

spec/index.bs
```diff
@@ -777,24 +784,27 @@ the exception thrown.
...
         the [=/origin=] of |provider|'s {{IdentityProviderConfig/configURL}}.
     1. If |loginStatus| is [=unknown=], a user agent MAY set it to [=logged-out=].
+    1. Let |showIdpLoginDialog| be an affordance to [=show an IDP login dialog=]:
```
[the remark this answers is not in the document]
tttzach: This is now an algorithm, thanks!

spec/index.bs
```diff
@@ -777,24 +784,27 @@ the exception thrown.
...
+        1. Let |config| be the result of running [=fetch the config file=]
+                  with |provider| and |globalObject|.
+        1. If |config| is failure, return (failure, true).
+        1. [=Show an IDP login dialog=] with |config| and |provider|.
+        1. If that algorithm returns failure, return (failure, true).
     1. If |loginStatus| is [=logged-out=], the user agent MUST do one of the following:
```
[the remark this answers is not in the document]
tttzach: Fixed, thank you!

spec/index.bs
tttzach: I used your suggestion, thanks!

spec/index.bs
tttzach: Fixed, thanks!

spec/index.bs
tttzach: Fixed, thanks!

spec/index.bs
tttzach: Good catch, I made it such that we check for the mode an user activation requirement prior to the login status, and also added a line to represent replacing the pending passive mode request, if any.

spec/index.bs
```diff
@@ -777,24 +784,27 @@ the exception thrown.
...
-                1. Let |config| be the result of running [=fetch the config file=]
-                    with |provider| and |globalObject|.
-                1. If |config| is failure, return (failure, true).
-                1. [=Show an IDP login dialog=] with |config| and |provider|.
-                1. If that algorithm returns failure, return (failure, true).
-
-        Issue: We should perhaps provide a way to let the [=RP=] request that
-            the second option is provided, possibly gated on a user gesture.
-            See [this issue](https://github.com/fedidcg/FedCM/issues/442) for discussion.
+            * If the user continues, trigger |showIdpLoginDialog|.
```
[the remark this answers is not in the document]
tttzach: Got it, thanks!

spec/index.bs
```diff
@@ -784,18 +786,24 @@ the exception thrown.
     1. Let |mode| be |options|'s {{IdentityCredentialRequestOptions/mode}}.
     1. If |mode| is [=active=]
         1. If [=transient activation=] is not present, return (failure, true).
-        1. If [=transient activation=] is present and if there is a pending request where |mode| is [=passive=], return (failure, true) for the pending request.
+        1. If [=transient activation=] is present and if there is a pending
```
TallTed, line 789: redundant if
```suggestion
        1. If [=transient activation=] is present and there is a pending
```

TallTed (changes requested):
trivial, editorial

spec/index.bs
tttzach: Thanks!

spec/index.bs
```diff
@@ -774,27 +783,28 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
 or a pair (failure, bool), where the bool indicates whether to skip delaying
 the exception thrown.
     1. Assert: These steps are running [=in parallel=].
+    1. Let |mode| be |options|'s {{IdentityCredentialRequestOptions/mode}}.
+    1. If |mode| is [=active=]
+        1. If [=transient activation=] is not present, return (failure, true).
+        1. If [=transient activation=] is present and there is a pending
+            request where |mode| is [=passive=], return (failure, true) for the
```
cbiesinger, line 790: Hmm this needs to be written differently (the previous request is not (or may not be) at this point in the algorithm so you can't just return here from there.

Maybe something like, `cancel the previous request as if a {{CredentialRequestOptions/signal}} was given to it and was [=AbortSignal/aborted=]`?

spec/index.bs
```diff
@@ -774,27 +783,28 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
...
-                1. If |config| is failure, return (failure, true).
-                1. [=Show an IDP login dialog=] with |config| and |provider|.
-                1. If that algorithm returns failure, return (failure, true).
-
-        Issue: We should perhaps provide a way to let the [=RP=] request that
-            the second option is provided, possibly gated on a user gesture.
-            See [this issue](https://github.com/fedidcg/FedCM/issues/442) for discussion.
+            * If the user triggers this affordance,
+                [=fetch the config file and show an IDP login dialog=].
+        * If |mode| is [=active=] and [=transient activation=] is present,
```
cbiesinger, line 805: I got a bit lost in the control flow but I think we verify earlier that transient activation is present if mode is active so we probably don't need that check here.

cbiesinger (approved):
Thanks, looks good modulo two minor comments

spec/index.bs
tttzach: Used your suggestion, thanks!

spec/index.bs
tttzach: Yes we do indeed verify earlier, removed the transient activation check. Thanks!

spec/index.bs
```diff
@@ -787,8 +787,9 @@ the exception thrown.
     1. If |mode| is [=active=]
         1. If [=transient activation=] is not present, return (failure, true).
         1. If [=transient activation=] is present and there is a pending
-            request where |mode| is [=passive=], return (failure, true) for the
-            pending request.
+            request where |mode| is [=passive=], cancel the previous request
+            as if a {{CredentialRequestOptions/signal}} was given to it and
+            was [=AbortSignal/aborted=].
```
TallTed, line 792:
```suggestion
            as if a {{CredentialRequestOptions/signal}} of
            [=AbortSignal/aborted=] was given to it.
```

tttzach: Thanks, Ted!

wseltzer:
Discussed in Oct. 15 call. [minutes to be linked]

tttzach:
@aaronpk @bvandersloot-mozilla @philsmart Please take a look!

spec/index.bs
```diff
@@ -1290,8 +1302,9 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
             is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
             |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
             the |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
-        1. The user agent MAY use the {{IdentityCredentialRequestOptions/context}} to customize the
-            dialog shown.
+        1. The user agent MAY use the
```
samuelgoto, line 1305:
```suggestion
        1. The user agent SHOULD use the
```

Should we use stronger language here?

spec/index.bs
```diff
@@ -1458,6 +1471,18 @@ success or failure.
             1. Otherwise, return failure.
 </div>
 
+<div algorithm>
+To <dfn>fetch the config file and show an IDP login dialog</dfn> given an {{IdentityProviderConfig}} |provider|, and a |globalObject|, run the following steps. This returns
```
samuelgoto, line 1475: Wouldn't it be possible to break this into two separate algorithms so that they are more reusable/narrow?

spec/index.bs
```diff
@@ -1458,6 +1471,18 @@ success or failure.
...
+success or failure.
+    1. Assert: these steps are running [=in parallel=].
+    1. Let |config| be the result of running [=fetch the config file=]
+              with |provider| and |globalObject|.
+    1. If |config| is failure, return failure.
```
samuelgoto, line 1480: at this point, haven't we already fetched the config file?

philsmart:
From the perspective of at least the R&E federations I am involved with, 'active' mode is a better fit. This mode supports authentication with Identity Providers (IdPs) that the user has not yet logged into (as signalled by the login status API). 

Supporting multiple IdPs in 'active' mode would be helpful, but it could result in the same account fetching issues as enabling multi-IdP in 'passive' mode. The 'passive' mode addressed this (if I recall correctly) by including only the IdPs you've signed into, which brings us back to the same issue (unable to use an IdP you've not logged into). This can be addressed for now by using an external IdP discovery service and populating FedCM with a single IdP in 'active' mode.

For what it's worth, I agree with renaming 'button' and 'widget' to 'active' and 'passive', it is clearer.

spec/index.bs
tttzach: Hmm this is the algorithm, unless you meant where we use this then no not yet - we haven't already fetched the config file upon checking the login status bit but it does come a couple lines after. Would it make sense to move the config file fetch to before checking the login status bit?

spec/index.bs
tttzach: Ah it looks like these two already exist separately. Removed and used them instead, thanks!

spec/index.bs
tttzach: Switched to stronger language here and similar text elsewhere, thanks!

yi-gu (approved):

cbiesinger (approved):

spec/index.bs
```diff
@@ -774,27 +783,30 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
 or a pair (failure, bool), where the bool indicates whether to skip delaying
 the exception thrown.
     1. Assert: These steps are running [=in parallel=].
+    1. Let |mode| be |options|'s {{IdentityCredentialRequestOptions/mode}}.
+    1. If |mode| is [=active=]
+        1. If [=transient activation=] is not present, return (failure, true).
```
npm1, line 788: transient activation is a Window concept. You need to specify which Window needs to have it. Also `present` is not the right term here since this is a boolean. Maybe "If W has transient activation[...]"

spec/index.bs
```diff
@@ -774,27 +783,30 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
...
+        1. If [=transient activation=] is present and there is a pending
```
npm1, line 789: This part about pending request is also not precise enough. Presumably this is also scoped to something (perhaps top-level browsing context?). In addition, you are never going to reach this because we disallow this from credential management in 8.2 https://w3c.github.io/webappsec-credential-management/#algorithm-request so that will need to be modified as well at some point

spec/index.bs
```diff
@@ -774,27 +783,30 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
...
+            request where |mode| is [=passive=], cancel the previous request
+            as if a {{CredentialRequestOptions/signal}} of
+            [=AbortSignal/aborted=] was given to it.
```
npm1, line 792: Hmm but isnt the abort signal only for the developer to use? This does not make sense to me, is this actually what we do?

spec/index.bs
```diff
@@ -774,27 +783,30 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
...
-                1. Let |config| be the result of running [=fetch the config file=]
-                    with |provider| and |globalObject|.
-                1. If |config| is failure, return (failure, true).
-                1. [=Show an IDP login dialog=] with |config| and |provider|.
-                1. If that algorithm returns failure, return (failure, true).
-
-        Issue: We should perhaps provide a way to let the [=RP=] request that
-            the second option is provided, possibly gated on a user gesture.
-            See [this issue](https://github.com/fedidcg/FedCM/issues/442) for discussion.
+    1. If |loginStatus| is [=logged-out=]
```
npm1, line 796: Missing : at the end

spec/index.bs
```diff
@@ -774,27 +783,30 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
...
+
+        1. If |mode| is [=active=],
+            [=fetch the config file=] and [=show an IDP login dialog=].
+        1. If |mode| is [=passive=], the user agent MUST do one of the following:
```
npm1, line 800: You can use otherwise here (there are only two modes)

spec/index.bs
```diff
@@ -774,27 +783,30 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
...
-                1. [=Show an IDP login dialog=] with |config| and |provider|.
-                1. If that algorithm returns failure, return (failure, true).
-
-        Issue: We should perhaps provide a way to let the [=RP=] request that
-            the second option is provided, possibly gated on a user gesture.
-            See [this issue](https://github.com/fedidcg/FedCM/issues/442) for discussion.
+    1. If |loginStatus| is [=logged-out=]
+
+        1. If |mode| is [=active=],
+            [=fetch the config file=] and [=show an IDP login dialog=].
```
npm1, line 799: Algorithms usually have inputs and outputs

spec/index.bs
```diff
@@ -774,27 +783,30 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
...
+        1. If |mode| is [=passive=], the user agent MUST do one of the following:
+            * Return (failure, false).
+            * Prompt the user whether to continue. If the user continues, the user
+                agent SHOULD set |loginStatus| to [=unknown=]. This MAY include an
+                affordance to [=show an IDP login dialog=].
+
+                * If the user cancels this dialog, return (failure, true).
+                * If the user triggers this affordance,
+                    [=fetch the config file=] and [=show an IDP login dialog=].
```
npm1, line 808: This used to be accurately specified, the simplification is not right

spec/index.bs
```diff
@@ -1290,8 +1303,9 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
             is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
             |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
             the |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
-        1. The user agent MAY use the {{IdentityCredentialRequestOptions/context}} to customize the
-            dialog shown.
+        1. The user agent SHOULD use the
```
npm1, line 1306: Why change may to should?

spec/index.bs
```diff
@@ -883,8 +895,9 @@ the exception thrown.
         1. Set |account| to |accountsList|[0].
         1. If [=compute the connection status=] of |account|, |provider| and |globalObject| returns
             [=compute the connection status/connected=], show a dialog to request user permission to sign
-            in via |account|, and set the result in |permission|. The user agent MAY use |options|'s
-            {{IdentityCredentialRequestOptions/context}} to customize the dialog.
+            in via |account|, and set the result in |permission|. The user agent SHOULD use |options|'s
```
npm1, line 898: Why change may to should?

spec/index.bs
npm1, line 799: Needs to handle failure case similar to how it was being done below

npm1 (changes requested):

spec/index.bs
bvandersloot-mozilla: ditto.

spec/index.bs
bvandersloot-mozilla: ditto

bvandersloot-mozilla (changes requested):
why SHOULD?

spec/index.bs
tttzach: Reverted, thanks!

spec/index.bs
tttzach: Reverted, thanks!

spec/index.bs
tttzach: Went with what we had in the original spec, thanks!

spec/index.bs
tttzach: Done, thanks!

spec/index.bs
tttzach: Went with what we had in the original spec, thanks!

tttzach: Went with what we had in the original spec, thanks!

spec/index.bs
tttzach: Done, thanks!

spec/index.bs
tttzach: No we don't actually use the abort signal for this case, but instead we directly reject the previous request. Though this was why its worded this way: https://github.com/w3c-fedid/FedCM/pull/660#discussion_r1790743904

Maybe we can get rid of `as if a {{CredentialRequestOptions/signal}} of [=AbortSignal/aborted=] was given to it.`?

spec/index.bs
tttzach: Good catch on the scoping... let me know if |W|'s top-level browsing context works or if that should be even more specific. Should modifying 8.2 be in this PR?

spec/index.bs
tttzach: Used your suggestion, thanks!

tttzach (commented):
Thanks for the review, please take another look!

spec/index.bs
```diff
@@ -784,29 +784,34 @@ or a pair (failure, bool), where the bool indicates whether to skip delaying
 the exception thrown.
     1. Assert: These steps are running [=in parallel=].
     1. Let |mode| be |options|'s {{IdentityCredentialRequestOptions/mode}}.
-    1. If |mode| is [=active=]
-        1. If [=transient activation=] is not present, return (failure, true).
-        1. If [=transient activation=] is present and there is a pending
-            request where |mode| is [=passive=], cancel the previous request
-            as if a {{CredentialRequestOptions/signal}} of
-            [=AbortSignal/aborted=] was given to it.
+    1. Let |globalObject| be the [=current global object=].
```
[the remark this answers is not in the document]
tttzach: Fixed, thanks!

spec/index.bs
```diff
@@ -784,29 +784,34 @@ or a pair (failure, bool), where the bool indicates whether to skip delaying
...
             * Prompt the user whether to continue. If the user continues, the user
                 agent SHOULD set |loginStatus| to [=unknown=]. This MAY include an
                 affordance to [=show an IDP login dialog=].
-
                 * If the user cancels this dialog, return (failure, true).
-                * If the user triggers this affordance,
-                    [=fetch the config file=] and [=show an IDP login dialog=].
-
+                * If the user triggers this affordance:
+                    1. Let |config| be the result of running [=fetch the config file=]
```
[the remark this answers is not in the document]
tttzach: Hmm Sam prefers them to be separate: https://github.com/w3c-fedid/FedCM/pull/660#discussion_r1801495711 but I do like the helper function so I brought it back.

spec/index.bs
```diff
@@ -784,29 +784,34 @@ or a pair (failure, bool), where the bool indicates whether to skip delaying
...
-        1. If [=transient activation=] is present and there is a pending
-            request where |mode| is [=passive=], cancel the previous request
-            as if a {{CredentialRequestOptions/signal}} of
-            [=AbortSignal/aborted=] was given to it.
+    1. Let |globalObject| be the [=current global object=].
+    1. Let |W| be |globalObject|'s [=associated Window=].
+    1. If |mode| is [=active=]:
+        1. If |W| does not have [=transient activation=], return (failure, true).
+        1. Otherwise, and if there is a pending request on |W|'s top-level browsing
+            context where |mode| is [=passive=], reject the previous request.
```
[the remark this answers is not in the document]
tttzach: Added, thanks!

spec/index.bs
```diff
@@ -784,29 +784,34 @@ or a pair (failure, bool), where the bool indicates whether to skip delaying
...
-        1. If [=transient activation=] is not present, return (failure, true).
-        1. If [=transient activation=] is present and there is a pending
-            request where |mode| is [=passive=], cancel the previous request
-            as if a {{CredentialRequestOptions/signal}} of
-            [=AbortSignal/aborted=] was given to it.
+    1. Let |globalObject| be the [=current global object=].
+    1. Let |W| be |globalObject|'s [=associated Window=].
+    1. If |mode| is [=active=]:
+        1. If |W| does not have [=transient activation=], return (failure, true).
+        1. Otherwise, and if there is a pending request on |W|'s top-level browsing
```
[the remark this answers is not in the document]
tttzach: Added a link and rephrased it a little, let me know if this works!

spec/index.bs
```diff
@@ -784,29 +784,34 @@ or a pair (failure, bool), where the bool indicates whether to skip delaying
...
     1. Assert: These steps are running [=in parallel=].
     1. Let |mode| be |options|'s {{IdentityCredentialRequestOptions/mode}}.
-    1. If |mode| is [=active=]
-        1. If [=transient activation=] is not present, return (failure, true).
-        1. If [=transient activation=] is present and there is a pending
-            request where |mode| is [=passive=], cancel the previous request
-            as if a {{CredentialRequestOptions/signal}} of
-            [=AbortSignal/aborted=] was given to it.
+    1. Let |globalObject| be the [=current global object=].
+    1. Let |W| be |globalObject|'s [=associated Window=].
```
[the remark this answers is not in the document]
tttzach: Fixed, thanks!

spec/index.bs
```diff
@@ -784,29 +784,26 @@ or a pair (failure, bool), where the bool indicates whether to skip delaying
...
-        1. If [=transient activation=] is not present, return (failure, true).
-        1. If [=transient activation=] is present and there is a pending
-            request where |mode| is [=passive=], cancel the previous request
-            as if a {{CredentialRequestOptions/signal}} of
-            [=AbortSignal/aborted=] was given to it.
+    1. If |mode| is [=active=]:
+        1. Let |W| be |globalObject|'s [=associated Window=].
+        1. If |W| does not have [=transient activation=], return (failure, true).
+        1. Otherwise, if there is a pending request where |mode| is [=passive=] on |W|'s
+            [top-level browsing context](https://html.spec.whatwg.org/#bc-traversable) or on any
```
npm1, line 791:
```suggestion
            [=Window/navigable=]'s [=navigable/top-level traversable=] or on any of its descendants
```
Added suggestion though did not check if the syntax is right but you get the idea

spec/index.bs
```diff
@@ -1472,6 +1469,19 @@ success or failure.
             1. Otherwise, return failure.
 </div>
 
+<div algorithm>
+To <dfn>fetch the config file and show an IDP login dialog</dfn> given an
+{{IdentityProviderConfig}} |provider|, and a |globalObject|, run the following
+steps. This returns success or failure.
```
npm1, line 1475: The return value is not being used in the callers, should they?

spec/index.bs
```diff
@@ -774,27 +783,27 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
...
-                    with |provider| and |globalObject|.
-                1. If |config| is failure, return (failure, true).
-                1. [=Show an IDP login dialog=] with |config| and |provider|.
-                1. If that algorithm returns failure, return (failure, true).
-
-        Issue: We should perhaps provide a way to let the [=RP=] request that
-            the second option is provided, possibly gated on a user gesture.
-            See [this issue](https://github.com/fedidcg/FedCM/issues/442) for discussion.
+    1. If |loginStatus| is [=logged-out=]:
+        1. If |mode| is [=active=], [=fetch the config file and show an IDP login dialog=]
```
npm1, line 797: You need to return failure if this fails.

spec/index.bs
```diff
@@ -774,27 +783,27 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
...
+            with |provider| and |globalObject|.
+        1. Otherwise, the user agent MUST do one of the following:
+            * Return (failure, false).
+            * Prompt the user whether to continue. If the user continues, the user
+                agent SHOULD set |loginStatus| to [=unknown=]. This MAY include an
+                affordance to [=show an IDP login dialog=].
+                * If the user cancels this dialog, return (failure, true).
+                * If the user triggers this affordance, [=fetch the config file and show an IDP login dialog=]
```
npm1, line 805: You need to return failure if this fails.

spec/index.bs
tttzach: Used your suggestion, thanks!

spec/index.bs
tttzach: Fixed, thanks!

spec/index.bs
tttzach: Fixed, thanks!

spec/index.bs
tttzach: Fixed, thanks!

spec/index.bs
```diff
@@ -774,27 +783,34 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
...
-
-        Issue: We should perhaps provide a way to let the [=RP=] request that
-            the second option is provided, possibly gated on a user gesture.
-            See [this issue](https://github.com/fedidcg/FedCM/issues/442) for discussion.
+    1. If |loginStatus| is [=logged-out=]:
+        1. If |mode| is [=active=]: 
+            1. Let |result| be the result of running
+                [=fetch the config file and show an IDP login dialog=] with
+                |provider| and |globalObject|.
+            1. If |result| is failure, return (failure, false).
```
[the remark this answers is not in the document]
tttzach: This should be true. Thanks!

spec/index.bs
```diff
@@ -774,27 +783,34 @@ To <dfn>create an IdentityCredential</dfn> given an {{IdentityProviderRequestOpt
...
+            * Return (failure, false).
+            * Prompt the user whether to continue. If the user continues, the user
+                agent SHOULD set |loginStatus| to [=unknown=]. This MAY include an
+                affordance to [=show an IDP login dialog=].
+                * If the user cancels this dialog, return (failure, true).
+                * If the user triggers this affordance:
+                    1. Let |result| be the result of running
+                        [=fetch the config file and show an IDP login dialog=]
+                        with |provider| and |globalObject|.
+                    1. If |result| is failure, return (failure, false).
```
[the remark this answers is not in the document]
tttzach: Fixed, thanks!

tttzach:
@bvandersloot-mozilla Friendly ping :)

hlflanagan:
Note: discussed during https://github.com/fedidcg/meetings/blob/main/2024/2024-11-12-notes.md ; PR blocked pending @bvandersloot-mozilla review.

bvandersloot-mozilla (approved):

PR #661, opened 2024-10-02
Specify the params API

cbiesinger:
Bug: w3c-fedid/custom-requests#2

***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/661.html" title="Last updated on Oct 8, 2024, 3:49 PM UTC (4e432ab)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/661/721c213...cbiesinger:4e432ab.html" title="Last updated on Oct 8, 2024, 3:49 PM UTC (4e432ab)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1206,7 +1206,7 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
         1. ("disclosure_text_shown", |disclosureTextShown|)
         1. ("is_auto_selected", |isAutoSelected|)
     1. If |provider|'s {{IdentityProviderRequestOptions/params}} is not empty:
-        1. Let |json| be the result of [=serializing a javascript value to a json string=]
+        1. Let |json| be the result of [=serializing a JavaScript value to a JSON string=]
             with |provider|'s {{IdentityProviderRequestOptions/params}}.
         1. Append ("params", |json|) to |list|.
     1. Let |requestBody| be the result of running [=urlencoded serializer=] with |list|.
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -653,7 +653,7 @@ dictionary IdentityProviderRequestOptions : IdentityProviderConfig {
   USVString nonce;
   DOMString loginHint;
   DOMString domainHint;
-  record<USVString, any> params;
+  any params;
 };
 </xmp>
 
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -653,6 +653,7 @@ dictionary IdentityProviderRequestOptions : IdentityProviderConfig {
   USVString nonce;
   DOMString loginHint;
   DOMString domainHint;
+  record<USVString, any> params;
 };
 </xmp>
 
@@ -1198,12 +1199,17 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
     1. Let |tokenUrl| be the result of [=computing the manifest URL=] given |provider|,
         |config|["{{IdentityProviderAPIConfig/id_assertion_endpoint}}"], and |globalObject|.
     1. If |tokenUrl| is failure, return failure.
-    1. Let |requestBody| be the result of running [=urlencoded serializer=] with a list containing:
+    1. Let |list| be a list with the following entries:
         1. ("client_id", |provider|'s {{IdentityProviderConfig/clientId}})
         1. ("nonce", |provider|'s {{IdentityProviderRequestOptions/nonce}})
         1. ("account_id",  |accountId|)
         1. ("disclosure_text_shown", |disclosureTextShown|)
         1. ("is_auto_selected", |isAutoSelected|)
+    1. If |provider|'s {{IdentityProviderRequestOptions/params}} is not empty:
+        1. Let |json| be the result of [=serializing a javascript value to a json string=]
+            with |provider|'s {{IdentityProviderRequestOptions/params}}.
+        1. Append ("params", |json|) to |list|.
+    1. Let |requestBody| be the result of running [=urlencoded serializer=] with |list|.
     1. Let |request| be a new <a spec=fetch for=/>request</a> as follows:
 
         :  [=request/url=]


samuelgoto (approved):

spec/index.bs
```diff
@@ -1198,12 +1199,17 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
     1. If |tokenUrl| is failure, return failure.
-    1. Let |requestBody| be the result of running [=urlencoded serializer=] with a list containing:
+    1. Let |list| be a list with the following entries:
         1. ("client_id", |provider|'s {{IdentityProviderConfig/clientId}})
         1. ("nonce", |provider|'s {{IdentityProviderRequestOptions/nonce}})
         1. ("account_id",  |accountId|)
         1. ("disclosure_text_shown", |disclosureTextShown|)
         1. ("is_auto_selected", |isAutoSelected|)
+    1. If |provider|'s {{IdentityProviderRequestOptions/params}} is not empty:
+        1. Let |json| be the result of [=serializing a javascript value to a json string=]
```
TallTed, line 1209:
```suggestion
        1. Let |json| be the result of [=serializing a JavaScript value to a JSON string=]
```

TallTed (changes requested):
Minor, human-facing

spec/index.bs
```diff
@@ -653,6 +653,7 @@ dictionary IdentityProviderRequestOptions : IdentityProviderConfig {
   USVString nonce;
   DOMString loginHint;
   DOMString domainHint;
+  record<USVString, any> params;
```
bvandersloot-mozilla, line 656: Is there a better option than `any` here? Would a single level do?

cbiesinger: Perhaps I misunderstood your comment at https://github.com/w3c-fedid/custom-requests/issues/2#issuecomment-2342125876. But the reason I picked any was that people in that issue, especially @aaronpk, wanted arbitrary objects.

samuelgoto: Yeah, I think we are better off with arbitrary (JSON-serializable) objects.

samuelgoto: But, wouldn't `object` do @cbiesinger ?

bvandersloot-mozilla: I think, if it is clear we are going to serialize it, it is probably fine. I forgot that conversation where Aaron was clear arbitrary structure was desired.

> But, wouldn't object do 
No because that would exclude a lot, e.g. `params: "a string"`

samuelgoto: On a related note, I was hoping for the following:

```suggestion
  object params;
```

I was thinking that we would take directly an `object`, rather than a `record<USVString, any>`.

samuelgoto: `params: "a string"` seems like a reasonable use case. so maybe it should be an union type?

```suggestion
  USVString or object params;
```

It doesn't seem to be that it is necessary to support:

- Arrays like `params: ["foo", {bar: 1}]`
- Booleans like `params: true`
- Numbers like `params: 3` or `params: 2.1` 
- Functions like `params: () => { return 1; }`

So, would `USVString or object` do?

cbiesinger: (I don't think "any" matches a function type, I don't see that in https://webidl.spec.whatwg.org/#idl-any)

As per https://webidl.spec.whatwg.org/#idl-object I *think* `object` is separate from `record`?

But I don't feel strongly about any of this. Should we discuss on the call tomorrow?

cbiesinger: @aaronpk any thoughts on whether we should allow non-maps for `params`?

samuelgoto: I think that if we leave the semantics of the `params` to be `any` and functionally equivalent to `encodeURIComponent(JSON.stringify(param))` it will do what we want it to do:

```javascript
encodeURIComponent(JSON.stringify("foobar"))
// '%22foobar%22'

encodeURIComponent(JSON.stringify({foo: 1}))
// '%7B%22foo%22%3A1%7D'

encodeURIComponent(JSON.stringify([1, 2, 3]))
// '%5B1%2C2%2C3%5D'
```

bvandersloot-mozilla (approved):

spec/index.bs
bvandersloot-mozilla: `any` is also appropriate, and probably better

elf-pavlik (approved):

wseltzer:
Discussed at 8 October meeting [minutes to be linked]. Agreed to merge when reviews complete.

spec/index.bs
cbiesinger: Changed to any

cbiesinger: Changed to any now.

samuelgoto (approved):
LGTM++

spec/index.bs
samuelgoto: I think this is a reference to an algorithm, so we have to keep the casing (i believe @cbiesinger ?).

samuelgoto:
> Discussed at 8 October meeting [minutes to be linked]. Agreed to merge when reviews complete.

I think we got all reviews complete. Can I merge this? Or is there anything else actionable here?

spec/index.bs
cbiesinger: No, bikeshed is smart.

samuelgoto: ah, thanks for the clarification!! lgtm++.

samuelgoto:
> Can I merge this? Or is there anything else actionable here?

I got the ack from @wseltzer to merge. Done.

spec/index.bs
```diff
@@ -1198,12 +1199,17 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
     1. If |tokenUrl| is failure, return failure.
-    1. Let |requestBody| be the result of running [=urlencoded serializer=] with a list containing:
+    1. Let |list| be a list with the following entries:
         1. ("client_id", |provider|'s {{IdentityProviderConfig/clientId}})
         1. ("nonce", |provider|'s {{IdentityProviderRequestOptions/nonce}})
         1. ("account_id",  |accountId|)
         1. ("disclosure_text_shown", |disclosureTextShown|)
         1. ("is_auto_selected", |isAutoSelected|)
+    1. If |provider|'s {{IdentityProviderRequestOptions/params}} is not empty:
+        1. Let |json| be the result of [=serializing a JavaScript value to a JSON string=]
```
samuelgoto, line 1209: @cbiesinger something that was brought to my attention while reviewing a different PR was that this algorithm [can throw](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#exceptions): that's unlikely and seems like a programming error, but seemed worth noting somewhere here what we should do when this throws.

PR #662, opened 2024-10-03
Specify the continuation API

cbiesinger:
Bug: w3c-fedid/custom-requests#1

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1240,9 +1240,25 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
         1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderToken}}, |token|.
         1. If one of the previous two steps threw an exception, set |credential| to failure
             and return.
+        1. If neither {{IdentityProviderToken/token}} nor {{IdentityProviderToken/continue_on}} was
+            specified, set |credential| to failure and return.
+        1. If {{IdentityProviderToken/token}} was not specified but
+            {{IdentityProviderToken/continue_on}} was specified:
+            1. Let |continueOnUrl| be the result of running [=parse url=] with |token|'s
+                {{IdentityProviderToken/continue_on}} and |globalObject|.
+            1. If |continueOnUrl| is failure, set |credential| to failure and return.
+            1. If |continueOnUrl| is not [=same origin=] with |tokenUrl|, set |credential|
+                to failure and return.
+            1. Let |tokenPair| be the result of [=show a continuation dialog=] with |continueOnUrl|.
+            1. If |tokenPair| is failure, set |credential| to failure and return.
+            1. Let |tokenString| be the first entry of |tokenPair|.
+            1. If the second entry of |tokenPair| is not null, set |accountId| to that second entry.
+        1. Otherwise, let |tokenString| be |token|'s {{IdentityProviderToken/token}}.
+        1. [=Create a connection between the RP and the IdP account=] with |provider|, |accountId|, and
+            |globalObject|.
         1. Let |credential| be a new {{IdentityCredential}} given |globalObject|'s
             <a for="global object">realm</a>.
-        1. Set |credential|'s {{IdentityCredential/token}} to |token|.
+        1. Set |credential|'s {{IdentityCredential/token}} to |tokenString|.
         1. Set |credential|'s {{IdentityCredential/isAutoSelected}} to
             |isAutoSelected|.
     1. Wait for |credential| to be set.
@@ -1251,7 +1267,8 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
 
 <xmp class="idl">
 dictionary IdentityProviderToken {
-  required USVString token;
+  USVString token;
+  USVString continue_on;
 };
 </xmp>
 
@@ -1293,8 +1310,6 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
         1. The user agent MAY use the {{IdentityCredentialRequestOptions/context}} to customize the
             dialog shown.
     1. If the user does not grant permission, return false.
-    1. [=Create a connection between the RP and the IdP account=] with |provider|, |account|, and
-        |globalObject|.
     1. Return true.
 </div>
 
@@ -1458,6 +1473,31 @@ success or failure.
             1. Otherwise, return failure.
 </div>
 
+<div algorithm>
+To <dfn>show a continuation dialog</dfn> given a |continueOnUrl|, run the
+following steps. This returns a failure or a tuple (string, string?) (a token
+and an optional account ID).
+    1. Assert: these steps are running [=in parallel=].
+    1. [=Create a fresh top-level traversable=] with |continueOnUrl|.
+    1. The user agent MAY [=set up browsing context features=] or otherwise
+        affect the presentation of this traversable in an implementation-defined
+        way.
+    1. Wait for one of the following conditions:
+        * The user closes the browsing context: return failure.
+        * {{IdentityProvider}}.{{IdentityProvider/close}} is called in the
+            context of this new traversable:
+            1. Close the traversable.
+            1. Return failure.
+        * {{IdentityProvider}}.{{IdentityProvider/resolve()}} is called in
+            the context of this new traversable.
+            1. Close the traversable.
+            1. Let |token| be the token that was passed to that resolve call.
+            1. If {{IdentityResolveOptions/accountId}} was specified in the
+                resolve call, let |accountId| be that account ID.
+            1. Otherwise, let |accountId| be null.
+            1. Return (|token|, |accountId|).
+
+</div>
 <!-- ============================================================ -->
 ## The IdentityProvider Interface ## {#browser-api-identity-provider-interface}
 <!-- ============================================================ -->
@@ -1473,8 +1513,13 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
     USVString picture;
   };
 
+  dictionary IdentityResolveOptions {
+    USVString accountId;
+  };
+
   [Exposed=Window, SecureContext] interface IdentityProvider {
       static undefined close();
+      static undefined resolve(DOMString token, optional IdentityResolveOptions options = {});
       static Promise&lt;sequence&lt;IdentityUserInfo&gt;&gt; getUserInfo(IdentityProviderConfig config);
   };
 </pre>


spec/index.bs
```diff
@@ -1458,6 +1473,31 @@ success or failure.
...
+following steps. This returns a failure or a tuple (string, string?) (a token
+and an optional account ID).
+    1. Assert: these steps are running [=in parallel=].
+    1. [=Create a fresh top-level traversable=] with |continueOnUrl|.
+    1. The user agent MAY [=set up browsing context features=] or otherwise
+        affect the presentation of this traversable in an implementation-defined
+        way.
+    1. Wait for one of the following conditions:
+        * The user closes the browsing context: return failure.
+        * {{IdentityProvider}}.{{IdentityProvider/close}} is called in the
```
ThisIsMissEm, line 1487: Should this actually be a `reject()` if the completion mechanism is `resolve()` — borrowing naming from Promises?

ThisIsMissEm:
I'm glad to see this idea being implemented!

spec/index.bs
samuelgoto: Ah yeah, that sounds nicer to me too!

samuelgoto (approved):

spec/index.bs
```diff
@@ -1240,9 +1240,25 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
         1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderToken}}, |token|.
         1. If one of the previous two steps threw an exception, set |credential| to failure
             and return.
+        1. If neither {{IdentityProviderToken/token}} nor {{IdentityProviderToken/continue_on}} was
+            specified, set |credential| to failure and return.
+        1. If {{IdentityProviderToken/token}} was not specified but
```
samuelgoto, line 1245: Just small nit and for clarity, but I'd return early here:

```suggestion
        1. If {{IdentityProviderToken/token}} was specified
          1.  let |tokenString| be |token|'s {{IdentityProviderToken/token}}.
        1. else  
```

samuelgoto (commented):
Minor nit

spec/index.bs
ThisIsMissEm: That or something like `abort(reason?: string)` and `finish(token: string, accountId?: string)`

spec/index.bs
ThisIsMissEm: should that be `1. Else if {{IdentityProviderToken/continue_on}} was specified:` ?

I think the `Otherwise` also would need to be removed with this change (L1256)

spec/index.bs
```diff
@@ -1473,8 +1513,13 @@ This specification introduces the {{IdentityUserInfo}} dictionary as well as the
     USVString picture;
   };
 
+  dictionary IdentityResolveOptions {
+    USVString accountId;
+  };
+
   [Exposed=Window, SecureContext] interface IdentityProvider {
       static undefined close();
+      static undefined resolve(DOMString token, optional IdentityResolveOptions options = {});
```
bvandersloot-mozilla, line 1522: Why is the token not in the options?

spec/index.bs
```diff
@@ -1458,6 +1473,31 @@ success or failure.
...
+        affect the presentation of this traversable in an implementation-defined
+        way.
+    1. Wait for one of the following conditions:
+        * The user closes the browsing context: return failure.
+        * {{IdentityProvider}}.{{IdentityProvider/close}} is called in the
+            context of this new traversable:
+            1. Close the traversable.
+            1. Return failure.
+        * {{IdentityProvider}}.{{IdentityProvider/resolve()}} is called in
+            the context of this new traversable.
```
bvandersloot-mozilla, line 1492: The processing model of resolve is underspecified here. I believe there are race conditions on multiple resolves in the spec.

spec/index.bs
```diff
@@ -1240,9 +1240,25 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
+            1. Let |continueOnUrl| be the result of running [=parse url=] with |token|'s
+                {{IdentityProviderToken/continue_on}} and |globalObject|.
+            1. If |continueOnUrl| is failure, set |credential| to failure and return.
+            1. If |continueOnUrl| is not [=same origin=] with |tokenUrl|, set |credential|
+                to failure and return.
+            1. Let |tokenPair| be the result of [=show a continuation dialog=] with |continueOnUrl|.
+            1. If |tokenPair| is failure, set |credential| to failure and return.
+            1. Let |tokenString| be the first entry of |tokenPair|.
+            1. If the second entry of |tokenPair| is not null, set |accountId| to that second entry.
+        1. Otherwise, let |tokenString| be |token|'s {{IdentityProviderToken/token}}.
```
yi-gu, line 1256: Does this mean we'd ignore `continue_on` when `token` is present? Would it apply to `error` in the future?

cbiesinger: The order should be:
- use error if specified
- use token if specified
- use continue_on if specified

(It's what chrome has implemented and what I think makes sense)

spec/index.bs
cbiesinger: IdentityProvider.close already exists (added to support logging in), so we either need to make it do something or make a decision it should not do something in this context.

IdentityProvider.reject makes sense to me but I'd prefer waiting until we have the error API (https://github.com/w3c-fedid/FedCM/pull/498) because that's what adds the error code and URL to the returned credential. We have no place to put the reason until then.

(for what it's worth, we also have no IDP feedback on an API like reject)

spec/index.bs
cbiesinger: Can you elaborate what the race condition is? Should it say "until the first of the following condition happens" or something like that?

spec/index.bs
cbiesinger: I felt that keeping the required argument as a direct function argument was more ergonomic than requiring to put it into the dictionary (somewhat analogous to HTTP structured field items, https://www.rfc-editor.org/rfc/rfc8941.html#name-parsing-an-item). But I'm open to changing it.

Things like addEventListener work that way too (https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener).

spec/index.bs
yi-gu: sounds good!

yi-gu (approved):

spec/index.bs
bvandersloot-mozilla: That sounds like a reasonable conclusion outside of specifying a [task queue](https://html.spec.whatwg.org/multipage/webappapis.html#task-queue) or something like that.

spec/index.bs
bvandersloot-mozilla: 👍  Seems fine!

wseltzer:
Discussed at 8 October meeting [minutes to be linked]

spec/index.bs
```diff
@@ -1251,7 +1267,8 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
 
 <xmp class="idl">
 dictionary IdentityProviderToken {
-  required USVString token;
+  USVString token;
```
samuelgoto, line 1270: Can you put some thought on how/whether we should think about forwards compatibility?

spec/index.bs
philsmart: Is there a use case where the IdP could send all these things in a response from the assertions endpoint such that a priority is required? I assumed that would have resulted in a FedCM error. Or maybe this kind of 'choice' can not be represented in the IdentityProviderToken dictionary.

philsmart: I guess I was thinking there would be a different dictionary for continuation (interaction required) responses.

spec/index.bs
```diff
@@ -1240,9 +1240,25 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
+            specified, set |credential| to failure and return.
+        1. If {{IdentityProviderToken/token}} was specified, let |tokenString|
+            be |token|'s {{IdentityProviderToken/token}}.
+        1.  Otherwise, if {{IdentityProviderToken/continue_on}} was specified:
+            1. Let |continueOnUrl| be the result of running [=parse url=] with |token|'s
+                {{IdentityProviderToken/continue_on}} and |globalObject|.
+            1. If |continueOnUrl| is failure, set |credential| to failure and return.
+            1. If |continueOnUrl| is not [=same origin=] with |tokenUrl|, set |credential|
+                to failure and return.
+            1. Let |tokenPair| be the result of [=show a continuation dialog=] with |continueOnUrl|.
```
npm1, line 1253: I think this is not currently running in parallel because it is inside `processResponseConsumeBody`. So this would need to invoke this in parallel

spec/index.bs
```diff
@@ -1251,7 +1267,8 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
 
 <xmp class="idl">
 dictionary IdentityProviderToken {
```
samuelgoto, line 1269: per @philsmart, this dictionary is poorly named: adding a `continue_on` to an `IdentityProviderToken` doesn't seem quite right. This isn't a web-visible dictionary, so we don't need to be too worried about picking a perfect name, but something that doesn't contain the word `token` would work.

```suggestion
dictionary IdentityAssertionResponse {
```

philsmart: This is looks better to me.

philsmart:
FWIW, I believe the IdP needs to have the option to request user interaction during token creation, for example when the RP has requested a higher level of authentication assurance (step-up authentication). Therefore, I consider this API essential.

spec/index.bs
```diff
@@ -1240,9 +1240,25 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
         1. [=converted to an IDL value|Convert=] |json| to an {{IdentityProviderToken}}, |token|.
```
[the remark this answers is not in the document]

spec/index.bs
```diff
@@ -1240,9 +1240,25 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
         1. If one of the previous two steps threw an exception, set |credential| to failure
             and return.
+        1. If neither {{IdentityProviderToken/token}} nor {{IdentityProviderToken/continue_on}} was
+            specified, set |credential| to failure and return.
+        1. If {{IdentityProviderToken/token}} was specified, let |tokenString|
+            be |token|'s {{IdentityProviderToken/token}}.
+        1.  Otherwise, if {{IdentityProviderToken/continue_on}} was specified:
```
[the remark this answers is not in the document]

spec/index.bs
cbiesinger: Thanks, done.

spec/index.bs
cbiesinger: I will address that in a later PR

philsmart (approved):

spec/index.bs
cbiesinger: Done.

spec/index.bs
```diff
@@ -1237,21 +1237,40 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
+            1. Let |continueOnUrl| be the result of running [=parse url=] with |token|'s
+                {{IdentityAssertionResponse/continue_on}} and |globalObject|.
+            1. If |continueOnUrl| is failure, set |credential| to failure and return.
+            1. If |continueOnUrl| is not [=same origin=] with |tokenUrl|, set |credential|
+                to failure and return.
+            1. Let |tokenPair| be the result of [=show a continuation dialog=] with |continueOnUrl|.
+            1. If |tokenPair| is failure, set |credential| to failure and return.
+            1. Let |tokenString| be the first entry of |tokenPair|.
+            1. If the second entry of |tokenPair| is not null, set |accountId| to that second entry.
+        1. Wait for |tokenString| to be set if necessary.
```
npm1, line 1258: If necessary? This is not clear. I think we wait until either `credential` is set to failure or `tokentString` is set. In the former, we return here (note that the return in the inner algorithm does not also automatically return from here.

hlflanagan:
Note: discussed during https://github.com/fedidcg/meetings/blob/main/2024/2024-11-12-notes.md

bvandersloot-mozilla (approved):

PR #668, opened 2024-10-28
Specify the fields API

cbiesinger:
Bug: w3c-fedid/custom-requests#4

***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/668.html" title="Last updated on Jan 7, 2025, 5:02 PM UTC (0d3a430)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/668/8201e01...cbiesinger:0d3a430.html" title="Last updated on Jan 7, 2025, 5:02 PM UTC (0d3a430)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -653,6 +653,7 @@ dictionary IdentityProviderRequestOptions : IdentityProviderConfig {
   USVString nonce;
   DOMString loginHint;
   DOMString domainHint;
+  sequence<USVString> fields;
   any params;
 };
 </xmp>
@@ -872,7 +873,7 @@ the exception thrown.
     1. For each |acc| in |accountsList|:
         1. If |acc| is [=eligible for auto reauthentication=] given |provider|, and |globalObject|,
             set |registeredAccount| to |acc| and increase |numRegisteredAccounts| by 1.
-    1. Let |permission|, |disclosureTextShown|, and |isAutoSelected| be set to false.
+    1. Let |permission|, |permissionRequested|, and |isAutoSelected| be set to false.
     1. If |mediation| is not "{{CredentialMediationRequirement/required}}", |requiresUserMediation|
         is false, and |numRegisteredAccounts| is equal to 1:
         1. Set |account| to |registeredAccount| and |permission| to true. When doing this, the user
@@ -888,23 +889,24 @@ the exception thrown.
             {{IdentityCredentialRequestOptions/context}} to customize the dialog.
         1. Otherwise, let |permission| be the result of running [=request permission to sign-up=]
             algorithm with |account|, |config|, |provider|, and |globalObject|. Also set
-            |disclosureTextShown| to true.
+            |permissionRequested| to true.
     1. Otherwise:
         1. Set |account| to the result of running the [=select an account=] from the
             |accountsList|.
         1. If |account| is failure, return (failure, true).
         1. If [=compute the connection status=] of |account|, |provider| and |globalObject| is
-            [=compute the connection status/connected=], set |permission| to true.
+            [=compute the connection status/connected=], or if |provider|.
+            {{IdentityProviderRequestOptions/fields}} is [=list/empty=], set |permission| to true.
         1. Otherwise:
             1. Let |permission| be the result of running the [=request permission to sign-up=]
                 algorithm with |account|, |config|, |provider|, and |globalObject|.
-            1. Set |disclosureTextShown| to true.
+            1. Set |permissionRequested| to true.
     1. Wait until the [=user agent=]'s dialogs requesting for user choice or permission to be
         closed, if any are created in the previous steps.
     1. Assert: |account| is not null.
     1. If |permission| is false, then return (failure, true).
     1. Let |credential| be the result of running the [=fetch an identity assertion=] algorithm with
-        |account|'s {{IdentityProviderAccount/id}}, |disclosureTextShown|, |isAutoSelected|,
+        |account|'s {{IdentityProviderAccount/id}}, |permissionRequested|, |isAutoSelected|,
         |provider|, |config|, and |globalObject|.
     1. Return |credential|.
 </div>
@@ -1193,18 +1195,32 @@ the token that will be provided to the [=RP=].
 
 <div algorithm>
 To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
-    |accountId|, a boolean |disclosureTextShown|, a boolean |isAutoSelected|, an
+    |accountId|, a boolean |permissionRequested|, a boolean |isAutoSelected|, an
     {{IdentityProviderRequestOptions}} |provider|, an {{IdentityProviderAPIConfig}} |config|,
     and |globalObject|, run the following steps. This returns an {{IdentityCredential}} or failure.
     1. Let |tokenUrl| be the result of [=computing the manifest URL=] given |provider|,
         |config|["{{IdentityProviderAPIConfig/id_assertion_endpoint}}"], and |globalObject|.
     1. If |tokenUrl| is failure, return failure.
+    1. Let |disclosureShownFor| and |fields| be the empty list.
+    1. If |permissionRequested| is true:
+        1. Set |fields| to |provider|.{{IdentityProviderRequestOptions/fields}}.
+        1. Set |disclosureShownFor| to the subset of strings in |fields| that are
+            in the [=list of recognized fields=].
     1. Let |list| be a list with the following entries:
         1. ("client_id", |provider|'s {{IdentityProviderConfig/clientId}})
         1. ("nonce", |provider|'s {{IdentityProviderRequestOptions/nonce}})
         1. ("account_id",  |accountId|)
-        1. ("disclosure_text_shown", |disclosureTextShown|)
         1. ("is_auto_selected", |isAutoSelected|)
+    1. If |fields| is not empty:
+        1. Let |serializedFields| be the entries of |fields| concatenated with a ","
+            between elements.
+        1. Append ("fields", |serializedFields|) to |list|.
+    1. If |disclosureShownFor| is not empty:
+        1. Let |serializedDisclosure| be the entries of |disclosureShownFor| concatenated
+            with a "," between elements.
+        1. Append ("disclosure_shown_for", |serializedDisclosure|) to |list|.
+    1. If |disclosureShownFor| contains all of "name", "email" and "picture", append
+        ("disclosure_text_shown", true) to |list|.
     1. If |provider|'s {{IdentityProviderRequestOptions/params}} is not empty:
         1. Let |json| be the result of [=serializing a JavaScript value to a JSON string=]
             with |provider|'s {{IdentityProviderRequestOptions/params}}.
@@ -1280,25 +1296,40 @@ The <a>request permission to sign-up</a> algorithm fetches the [=client metadata
 waits for the user to grant permission to use the given account, and returns whether the user
 granted permission or not.
 
-<div algorithm>
+<div algorithm="request permission to sign-up">
 To <dfn>request permission to sign-up</dfn> the user with a given an {{IdentityProviderAccount}} |account|,
 an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}} |provider|, and a
 |globalObject|, run the following steps. This returns a boolean.
     1. Assert: These steps are running [=in parallel=].
     1. Let |metadata| be the result of running [=fetch the client metadata=] with |config|,
         |provider|, and |globalObject|.
+    1. Let |fields| be |provider|.{{IdentityProviderRequestOptions/fields}} or, if not specified,
+        `["name", "email", "picture"]`.
+
+        Note: Unspecified is different from an explicitly specified empty list.
     1. Prompt the user to gather explicit intent to create an account. The user agent MAY use the
         {{IdentityProviderBranding}} to inform the style choices of its UI. Additionally:
-        1. If |metadata| is not failure, |metadata|["{{IdentityProviderClientMetadata/privacy_policy_url}}"]
-            is defined and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
-            |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
-            the |metadata|["{{IdentityProviderClientMetadata/privacy_policy_url}}"] link.
-        1. If |metadata| is not failure, |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"]
-            is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
-            |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
-            the |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
-        1. The user agent MAY use the {{IdentityCredentialRequestOptions/context}} to customize the
-            dialog shown.
+        1. If |provider|.{{IdentityProviderRequestOptions/fields}} is not [=list/empty=]:
+            1. If |metadata| is not failure, |metadata|["{{IdentityProviderClientMetadata/privacy_policy_url}}"]
+                is defined and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
+                |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
+                the |metadata|["{{IdentityProviderClientMetadata/privacy_policy_url}}"] link.
+            1. If |metadata| is not failure, |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"]
+                is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
+                |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
+                the |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
+            1. The user agent MUST prompt the user for permission to share the data in |fields|,
+                interpreting the strings as follows (this is the <dfn>list of recognized fields</dfn>):
+                : `"name"`
+                :: The user's name as given in {{IdentityProviderAccount}}.{{IdentityProviderAccount/name}}.
+                : `"email"`
+                :: The user's email address as given in {{IdentityProviderAccount}}.{{IdentityProviderAccount/email}}.
+                : `"picture"`
+                :: The user's profile picture as given in {{IdentityProviderAccount}}.{{IdentityProviderAccount/picture}}.
+
+                Any other string is ignored for forwards compatibility.
+            1. The user agent MAY use the {{IdentityCredentialRequestOptions/context}} to customize the
+                dialog shown.
     1. If the user does not grant permission, return false.
     1. [=Create a connection between the RP and the IdP account=] with |provider|, |account|, and
         |globalObject|.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1193,18 +1195,32 @@ the token that will be provided to the [=RP=].
...
         1. ("account_id",  |accountId|)
-        1. ("disclosure_text_shown", |disclosureTextShown|)
         1. ("is_auto_selected", |isAutoSelected|)
+    1. If |fields| is not empty:
+        1. Let |serializedFields| be the entries of |fields| concatenated with a ","
+            between elements.
+        1. Append ("fields", |serializedFields|) to |list|.
+    1. If |disclosureShownFor| is not empty:
+        1. Let |serializedDisclosure| be the entries of |disclosureShownFor| concatenated
+            with a "," between elements.
```
TallTed, line 1220:
```suggestion
            with a comma ("`,`") between elements.
```

spec/index.bs
```diff
@@ -1193,18 +1195,32 @@ the token that will be provided to the [=RP=].
...
+        1. Set |disclosureShownFor| to the subset of strings in |fields| that are
+            in the [=list of recognized fields=].
     1. Let |list| be a list with the following entries:
         1. ("client_id", |provider|'s {{IdentityProviderConfig/clientId}})
         1. ("nonce", |provider|'s {{IdentityProviderRequestOptions/nonce}})
         1. ("account_id",  |accountId|)
-        1. ("disclosure_text_shown", |disclosureTextShown|)
         1. ("is_auto_selected", |isAutoSelected|)
+    1. If |fields| is not empty:
+        1. Let |serializedFields| be the entries of |fields| concatenated with a ","
```
TallTed, line 1215:
```suggestion
        1. Let |serializedFields| be the entries of |fields| concatenated with a comma ("`,`")
```

spec/index.bs
```diff
@@ -1280,25 +1296,40 @@ The <a>request permission to sign-up</a> algorithm fetches the [=client metadata
...
-            the |metadata|["{{IdentityProviderClientMetadata/privacy_policy_url}}"] link.
-        1. If |metadata| is not failure, |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"]
-            is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
-            |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
-            the |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
-        1. The user agent MAY use the {{IdentityCredentialRequestOptions/context}} to customize the
-            dialog shown.
+        1. If |provider|.{{IdentityProviderRequestOptions/fields}} is not [=list/empty=]:
+            1. If |metadata| is not failure, |metadata|["{{IdentityProviderClientMetadata/privacy_policy_url}}"]
+                is defined and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
```
TallTed, line 1314:
```suggestion
                is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
```

spec/index.bs
```diff
@@ -1280,25 +1296,40 @@ The <a>request permission to sign-up</a> algorithm fetches the [=client metadata
...
+                |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
+                the |metadata|["{{IdentityProviderClientMetadata/privacy_policy_url}}"] link.
+            1. If |metadata| is not failure, |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"]
+                is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
+                |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
+                the |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
+            1. The user agent MUST prompt the user for permission to share the data in |fields|,
+                interpreting the strings as follows (this is the <dfn>list of recognized fields</dfn>):
```
TallTed, line 1322:
```suggestion
                interpreting the strings in the <dfn>list of recognized fields</dfn> as follow:
```

TallTed (changes requested):

cbiesinger:
Thanks @TallTed, addresssed your comments.

spec/index.bs
```diff
@@ -872,7 +873,7 @@ the exception thrown.
     1. For each |acc| in |accountsList|:
         1. If |acc| is [=eligible for auto reauthentication=] given |provider|, and |globalObject|,
             set |registeredAccount| to |acc| and increase |numRegisteredAccounts| by 1.
-    1. Let |permission|, |disclosureTextShown|, and |isAutoSelected| be set to false.
+    1. Let |permission|, |permissionRequested|, and |isAutoSelected| be set to false.
```
npm1, line 876: Why do you like this new name better? This variable is true when the PP/TOS text is shown to the user. The user may still select an account, and I think that sort of counts as providing permission, so I find this new wording a bit confusing.

spec/index.bs
```diff
@@ -1193,18 +1195,32 @@ the token that will be provided to the [=RP=].
...
+        1. Set |disclosureShownFor| to the subset of strings in |fields| that are
+            in the [=list of recognized fields=].
     1. Let |list| be a list with the following entries:
         1. ("client_id", |provider|'s {{IdentityProviderConfig/clientId}})
         1. ("nonce", |provider|'s {{IdentityProviderRequestOptions/nonce}})
         1. ("account_id",  |accountId|)
-        1. ("disclosure_text_shown", |disclosureTextShown|)
         1. ("is_auto_selected", |isAutoSelected|)
+    1. If |fields| is not empty:
+        1. Let |serializedFields| be the entries of |fields| concatenated with a comma ("`,`")
```
npm1, line 1215: Hmm is this commonly called serialized, or perhaps rename this one?

spec/index.bs
```diff
@@ -1193,18 +1195,32 @@ the token that will be provided to the [=RP=].
...
+            between elements.
+        1. Append ("fields", |serializedFields|) to |list|.
+    1. If |disclosureShownFor| is not empty:
+        1. Let |serializedDisclosure| be the entries of |disclosureShownFor| concatenated
+            with a comma ("`,`") between elements.
+        1. Append ("disclosure_shown_for", |serializedDisclosure|) to |list|.
+    1. If |disclosureShownFor| contains all of "name", "email", and "picture", append
```
npm1, line 1222: Is it actually the case that we only show disclosure text if all of these are present? My intuition was that it would be shown if at least one of these is present.

spec/index.bs
```diff
@@ -1280,25 +1296,40 @@ The <a>request permission to sign-up</a> algorithm fetches the [=client metadata
...
 
-<div algorithm>
+<div algorithm="request permission to sign-up">
 To <dfn>request permission to sign-up</dfn> the user with a given an {{IdentityProviderAccount}} |account|,
 an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}} |provider|, and a
 |globalObject|, run the following steps. This returns a boolean.
     1. Assert: These steps are running [=in parallel=].
     1. Let |metadata| be the result of running [=fetch the client metadata=] with |config|,
         |provider|, and |globalObject|.
+    1. Let |fields| be |provider|.{{IdentityProviderRequestOptions/fields}} or, if not specified,
```
npm1, line 1306: s/specified/present. Alternatively, I think we could use = ['name', 'email', 'picture'] in the IDL

cbiesinger: Is that allowed by WebIDL? (it does not seem to be allowed by Chrome's IDL parser)

spec/index.bs
cbiesinger: Renamed.

spec/index.bs
```diff
@@ -888,23 +889,24 @@ the exception thrown.
...
             algorithm with |account|, |config|, |provider|, and |globalObject|. Also set
-            |disclosureTextShown| to true.
+            |permissionRequested| to true.
     1. Otherwise:
         1. Set |account| to the result of running the [=select an account=] from the
             |accountsList|.
         1. If |account| is failure, return (failure, true).
         1. If [=compute the connection status=] of |account|, |provider| and |globalObject| is
-            [=compute the connection status/connected=], set |permission| to true.
+            [=compute the connection status/connected=], or if |provider|.
```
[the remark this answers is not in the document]

spec/index.bs
cbiesinger: We set permissionRequested to true when we call the [=request permission to sign-up=] algorithm, and that algorithm might not show a disclosure text (sign-up case when the user is signed in to only one account). Hence, the new name fits better.

spec/index.bs
cbiesinger: I'm told WebIDL does not allow such a default (only allows an empty list)

npm1 (approved):

spec/index.bs
```diff
@@ -895,8 +895,13 @@ the exception thrown.
             |accountsList|.
         1. If |account| is failure, return (failure, true).
         1. If [=compute the connection status=] of |account|, |provider| and |globalObject| is
```
TallTed, line 897:
```suggestion
        1. If [=compute the connection status=] of |account|, |provider|, and |globalObject| is
```

spec/index.bs
```diff
@@ -895,8 +895,13 @@ the exception thrown.
...
-            [=compute the connection status/connected=], or if |provider|.
-            {{IdentityProviderRequestOptions/fields}} is [=list/empty=], set |permission| to true.
+            [=compute the connection status/connected=], set |permission| to true.
+        1. Otherwise, if |provider|.{{IdentityProviderRequestOptions/fields}} is [=list/empty=],
+            [=create a connection between the RP and the IdP account=] with |provider|, |account|,
+            and |globalObject| and set |permission| to |true|.
```
TallTed, line 901:
```suggestion
            and |globalObject|, and set |permission| to |true|.
```

spec/index.bs
```diff
@@ -1212,15 +1217,20 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
-    1. If |disclosureShownFor| contains all of "name", "email" and "picture", append
+        1. Let |disclosureString| be the entries of |disclosureShownFor| concatenated
+            with a comma ("`,`") between elements.
+        1. Append ("disclosure_shown_for", |disclosureString|) to |list|.
+    1. If |disclosureShownFor| contains all of "name", "email", and "picture", append
         ("disclosure_text_shown", true) to |list|.
+
+        Note: This parameter exists for backwards compatibility with older identity providers
+            that do not yet support `disclosure_shown_for`. At the time, the disclosure text,
+            if shown, always included name, email and picture. Newer identity providers should
```
TallTed, line 1232:
```suggestion
            if shown, always included name, email, and picture. Newer identity providers should
```

spec/index.bs
```diff
@@ -1212,15 +1217,20 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
+            check `disclosure_shown_for` instead.
```
TallTed, line 1233:
```suggestion
            instead check `disclosure_shown_for`.
```

TallTed (changes requested):

wseltzer:
Discussed in 19 Nov 2024 WG call [link to minutes]. Not yet resolved.

cbiesinger:
I addressed TallTed's comments.

TallTed (approved):

cbiesinger:
I have updated the PR to make it optional for a user agent to show a disclosure text/permission prompt.

spec/index.bs
```diff
@@ -1378,7 +1378,8 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
         metadata=] with |config|, |provider|, and |globalObject|.
     1. Prompt the user to gather explicit intent to create an account. The user agent MAY use the
         {{IdentityProviderBranding}} to inform the style choices of its UI. Additionally:
-        1. If |fields| is not [=list/empty=]:
+        1. The user agent SHOULD <dfn>support showing a permission prompt</dfn> as follows.
```
npm1, line 1381: This is not a step. Can you merge it into the previous bullet with Additionally? If the user agent <dfn> supports...</dfn>, run the following:

spec/index.bs
```diff
@@ -1378,7 +1378,8 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
...
+        1. If |fields| is not [=list/empty=] and the user agent [=supports showing a permission prompt=]:
```
npm1, line 1382: And with the previous change, this change is not needed.

spec/index.bs

spec/index.bs

cbiesinger:
I've now also added a note that IDPs should support continue_on.

spec/index.bs
```diff
@@ -1377,9 +1377,15 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
...
     1. Prompt the user to gather explicit intent to create an account. The user agent MAY use the
-        {{IdentityProviderBranding}} to inform the style choices of its UI. Additionally:
-        1. The user agent SHOULD <dfn>support showing a permission prompt</dfn> as follows.
-        1. If |fields| is not [=list/empty=] and the user agent [=supports showing a permission prompt=]:
+        {{IdentityProviderBranding}} to inform the style choices of its UI. Additionally, if the user agent
+        <dfn>supports showing a permission prompt</dfn>:
+
+        Note: Identity providers should support showing their own permission prompt using
+            {{IdentityAssertionResponse/continue_on}} when the `disclosure_shown_for` parameter
+            does not contain the fields required by the IDP. This is to support user agents that
```
npm1, line 1385: nit: support galore, use synonym sometimes?

npm1 (commented):
The last bullet point about context/mode seems unrelated to fields, should it go with the rest or totally outside of the new dfn?

spec/index.bs
TallTed:
```suggestion
            does not contain the fields required by the IDP. This is to enable user agents that
```

cbiesinger:
Because this came up in the meeting -- if a user agent does not (want to) support this feature, then it would not "support[] showing a permission prompt" and would not fall into the "MUST" cases for the UI. That was at least the intention, if I missed a case please do point it out!

spec/index.bs
```diff
@@ -1338,26 +1364,49 @@ The <a>request permission to sign-up</a> algorithm fetches the [=client metadata
...
+        1. If |fields| is not [=list/empty=]:
+            1. If |metadata| is not failure, |metadata|["{{IdentityProviderClientMetadata/privacy_policy_url}}"]
+                is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
+                |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
+                the |metadata|["{{IdentityProviderClientMetadata/privacy_policy_url}}"] link.
+            1. If |metadata| is not failure, |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"]
+                is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
+                |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
+                the |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
+            1. The user agent MUST prompt the user for permission to share the data in |fields|,
```
bvandersloot-mozilla, line 1397: Does this force the display of a dialog with these fields shown?

cbiesinger: No because line 1380 makes this step only apply if the user agent supports showing a permission prompt.

bvandersloot-mozilla: Ah I missed that because tabs are hard and that line was broken up in the diff. 👍

cbiesinger: The autogenerated "Diff" link (in the initial comment in the PR) may be helpful for cases like this

philsmart:
I realise 'fields' are meaningful parts of data structures, but the naming seems less obvious to an IdP. They sound more specifically like 'requested (identity) attributes/claims'. Would it be worth looking at the naming of this, at least when it is POSTed to the assertions endpoint.

I guess overlapping with OIDCs use of Scopes, and SAML's use of RequestedAttributes.

samuelgoto:
> I realise 'fields' are meaningful parts of data structures, but the naming seems less obvious to an IdP. They sound more specifically like 'requested (identity) attributes/claims'. Would it be worth looking at the naming of this, at least when it is POSTed to the assertions endpoint.
> 
> I guess overlapping with OIDCs use of Scopes, and SAML's use of RequestedAttributes.

I think the closest to `fields` is OIDC's `claims`, rather than OAuth's `scope`. We picked `fields` because we figured it would be worth making it (a) OIDC-agnostic (e.g. it has to make sense for SAML too) and (b) align with the HTML spec, where `<forms>` have `<fieldset>` and `fields`. I do expect `fields` to connect to HTML's autofill at some point, so that's why I still think it is properly named. Happy to hear other suggestions, but just wanted to give some context of the considerations we made when picking it.

philsmart:
Thanks, @samuelgoto. That's a good point about being protocol agnostic. From an identity perspective, my understanding is that the RP will request certain attributes, and those are returned as verified (by the IdP) claims or assertions. 

I guess this pops up across the Identity and Access landscape in different guidelines and specifications e.g. [LDAP](https://datatracker.ietf.org/doc/html/rfc7643#section-4.1), [SAML](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf), SCIM, [OIDC](https://openid.net/specs/openid-connect-core-1_0.html#Terminology), [NIST SP800-63](https://pages.nist.gov/800-63-4/sp800-63.html)('Access to the service requires a defined list of attributes'), [NIST SP800-63c](https://pages.nist.gov/800-63-4/sp800-63c.html#common)('The federation transaction can also provide the RP with a set of identity attributes'), and so on. I personally wouldn't oppose reusing an existing term here, but I am not very familiar with alignment to the HTML specification, so maybe 'fields' is best.

bvandersloot-mozilla (approved):
Since this is not required, looks fine to me!

ThisIsMissEm:
@wseltzer @hlflanagan was this pull request reverted? I don't see this in the current editors draft: https://w3c-fedid.github.io/FedCM/ — there's just a bunch of `ISSUE: extension` blocks in the current draft.

Has Fields API shipped or not? The [status list](https://github.com/w3c-fedid/FedCM/wiki/Status-of-FPWD%E2%80%90identified-Issues-(Consensus-Blockers-for-CR)) says Fields APi is "now merged as part of FedCM"

cbiesinger:
The API has shipped in Chrome. And I see the changes from this PR in https://w3c-fedid.github.io/FedCM/#dom-identityproviderrequestoptions-fields and other places?

ThisIsMissEm:
@cbiesinger yeah, but the ID Assertion algorithm is missing them in the returned credential: https://w3c-fedid.github.io/FedCM/#fetch-an-identity-assertion

so you can request fields, but you never get them back?

cbiesinger:
We don't specify the format of the token. It's up to the IDP to include the value for the requested fields in whichever token format they use.

Are you suggesting that we could include the value of the fields from the accounts endpoint in the credential object? that would be possible, but is not what we were originally thinking. I would suggest filing a new issue for that use case.

ThisIsMissEm:
@cbiesinger I was expecting the `credentials` object to contain a top-level `fields` property, whether that's populated from the accounts endpoint or the "`Convert` json to an `IdentityAssertionResponse`" is unknown to me, but I suspect it may have to be the ID Assertion endpoint, since the IdP may need to enforce client permissions around retrieving certain fields, e.g,. needing `scope` of `account:email` in order to read the account's email (such that it shows in a consent flow)

PR #678, opened 2024-11-04
Specify "Use another account".

cbiesinger:
This depends on PR #660 (Mode API).

Bug: w3c-fedid/active-mode#3

***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/678.html" title="Last updated on Jun 17, 2025, 4:08 PM UTC (6ced827)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/678/25b4735...cbiesinger:6ced827.html" title="Last updated on Jun 17, 2025, 4:08 PM UTC (6ced827)">Diff</a>


spec/index.bs
```diff
@@ -1267,10 +1281,17 @@ dictionary IdentityProviderToken {
 <!-- ============================================================ -->
 
 <div algorithm>
-To <dfn>select an account</dfn> given an |accountsList|, run the following steps. This returns an
-{{IdentityProviderAccount}} or failure.
+To <dfn>select an account</dfn> given an |accountsList|, {{IdentityProviderModeSettings}}
```
TallTed, line 1284:
```suggestion
To <dfn>select an account</dfn> given an |accountsList|, some {{IdentityProviderModeSettings}}
```

spec/index.bs
```diff
@@ -1267,10 +1281,17 @@ dictionary IdentityProviderToken {
...
+|modeSettings|, an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderConfig}} |provider|,
+and a |globalObject|, run the following steps. This returns an {{IdentityProviderAccount}} or
+failure.
     1. Assert |accountsList|'s [=list/size=] is greater than 1.
     1. Display an account chooser displaying the options from |accountsList|.
+    1. If |modeSettings|.{{IdentityProviderModeSettings/supports_use_other_account}} is true,
+        the account chooser MUST provide an affordance to use another account. If that
+        affordance is triggered:
+        1. [=Show an IDP login dialog=] with |config|, |provider| and |globalObject|.
+        1. If that returned sucess, go back to the [=fetch accounts step=].
```
TallTed, line 1294:
```suggestion
        1. If that returned success, go back to the [=fetch accounts step=].
```

npm1 (changes requested):
In the case the user is signed in to just one account, it does not appear to show the use other account feature. So the logic needs to change to account for that

spec/index.bs
```diff
@@ -804,6 +804,10 @@ the exception thrown.
     1. Let |config| be the result of running [=fetch the config file=] with
         |provider| and |globalObject|.
     1. If |config| is failure, return (failure, false).
+    1. Let |modeSettings| be |config|.{{IdentityProviderAPIConfig/modes}}.
+        {{IdentityProviderModes/active}} if |options|.{{IdentityCredentialRequestOptions/mode}}
```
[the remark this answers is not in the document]
cbiesinger: This will only be linked once PR #660 (mode API) lands. I don't know how to do PRs that depend on one another.

cbiesinger:
> In the case the user is signed in to just one account, it does not appear to show the use other account feature. So the logic needs to change to account for that

I updated the PR to add the additional logic where required, PTAL

spec/index.bs
```diff
@@ -1267,10 +1287,17 @@ dictionary IdentityProviderToken {
...
+|modeSettings|, an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderConfig}} |provider|,
+and a |globalObject|, run the following steps. This returns an {{IdentityProviderAccount}} or
+failure.
     1. Assert |accountsList|'s [=list/size=] is greater than 1.
     1. Display an account chooser displaying the options from |accountsList|.
+    1. If |modeSettings|.{{IdentityProviderModeSettings/supports_use_other_account}} is true,
+        the account chooser MUST provide an affordance to use another account. If that
+        affordance is triggered:
+        1. [=Show an IDP login dialog=] with |config|, |provider| and |globalObject|.
+        1. If that returned success, go back to the [=fetch accounts step=].
```
npm1, line 1300: Otherwise return failure here I think?

spec/index.bs
```diff
@@ -1299,6 +1327,12 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
             the |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
         1. The user agent MAY use the {{IdentityCredentialRequestOptions/context}} to customize the
             dialog shown.
+        1. If |modeSettings|.{{IdentityProviderModeSettings/supports_use_other_account}} is true,
+            the account chooser MUST provide an affordance to use another account unless such an
+            affordance was provided in a previous step (e.g. if [=select an account=] was
+            invoked). If that affordance is triggered:
+            1. [=Show an IDP login dialog=] with |config|, |provider| and |globalObject|.
+            1. If that returned success, go back to the [=fetch accounts step=].
```
npm1, line 1335: ditto otherwise

spec/index.bs
```diff
@@ -886,18 +891,23 @@ the exception thrown.
...
             {{IdentityCredentialRequestOptions/context}} to customize the dialog.
-        1. Otherwise, let |permission| be the result of running [=request permission to sign-up=]
-            algorithm with |account|, |config|, |provider|, and |globalObject|. Also set
-            |disclosureTextShown| to true.
+            If |modeSettings|.{{IdentityProviderModeSettings/supports_use_other_account}} is true,
+                the account chooser MUST provide an affordance to use another account. If that
+                affordance is triggered:
+                1. [=Show an IDP login dialog=] with |config|, |provider| and |globalObject|.
+                1. If that returned success, go back to the [=fetch accounts step=].
+                1. Otherwise, let |permission| be the result of running [=request permission to sign-up=]
```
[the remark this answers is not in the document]
cbiesinger: The current Chrome implementation and IMO the better experience is not to terminate the flow in this situation.

spec/index.bs
cbiesinger: Ditto

spec/index.bs
cbiesinger: Ditto

spec/index.bs
```diff
@@ -886,18 +891,23 @@ the exception thrown.
             [=compute the connection status/connected=], show a dialog to request user permission to sign
             in via |account|, and set the result in |permission|. The user agent MAY use |options|'s
             {{IdentityCredentialRequestOptions/context}} to customize the dialog.
+            If |modeSettings|.{{IdentityProviderModeSettings/supports_use_other_account}} is true,
+                the account chooser MUST provide an affordance to use another account. If that
```
npm1, line 895: this indentation is confusing, this is supposed to be part of the current statement so should not be indented more. But actually I think it is cleaner to use a separate statement for this. To do that we need to move the 'show a dialog' to an inner statement
```
1. If compute returns connected:
  1. Show a dialog
  1. If support_use_other_account
1. Otherwise, ...
```

npm1 (approved):
lgtm with nit

spec/index.bs
cbiesinger: Thanks, done.

spec/index.bs
```diff
@@ -887,21 +888,27 @@ the exception thrown.
...
-            in via |account|, and set the result in |permission|. The user agent MAY use |options|'s
-            {{IdentityCredentialRequestOptions/context}} to customize the dialog.
+            [=compute the connection status/connected=]:
+            1. Show a dialog to request user permission to sign in via |account|, and set the result
+                in |permission|. The user agent MAY use |options|'s
+                {{IdentityCredentialRequestOptions/context}} to customize the dialog.
+            1. If |modeSettings|.{{IdentityProviderModeSettings/supports_use_other_account}} is true,
+                that dialog MUST provide an affordance to use another account. If that
+                affordance is triggered:
+                1. [=Show an IDP login dialog=] with |config|, |provider| and |globalObject|.
```
TallTed, line 898:
```suggestion
                1. [=Show an IDP login dialog=] with |config|, |provider|, and |globalObject|.
```

spec/index.bs
```diff
@@ -1302,9 +1309,10 @@ waits for the user to grant permission to use the given account, and returns whe
...
 
 <div algorithm>
-To <dfn>request permission to sign-up</dfn> the user with a given an {{IdentityProviderAccount}} |account|,
-an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}} |provider|, and a
-|globalObject|, run the following steps. This returns a boolean.
+To <dfn>request permission to sign-up</dfn> the user with a given an {{IdentityProviderAccount}}
+|account|, some {{IdentityProviderModeSettings}} |modeSettings|, an {{IdentityProviderAPIConfig}}
+|config|, an {{IdentityProviderRequestOptions}} |provider|, and a |globalObject|, run the following
+steps. This returns a boolean.
     1. Assert: These steps are running [=in parallel=].
```
TallTed, line 1316: The parallel steps would be better presented as bullets (which might bullets within a numbered step), rather than as numbered steps. As this stands, I am unsure exactly *which* steps are `running in parallel`.

spec/index.bs
```diff
@@ -1320,6 +1328,12 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
             the |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
         1. The user agent MAY use the {{IdentityCredentialRequestOptions/context}} to customize the
             dialog shown.
+        1. If |modeSettings|.{{IdentityProviderModeSettings/supports_use_other_account}} is true,
+            the account chooser MUST provide an affordance to use another account unless such an
+            affordance was provided in a previous step (e.g. if [=select an account=] was
+            invoked). If that affordance is triggered:
+            1. [=Show an IDP login dialog=] with |config|, |provider| and |globalObject|.
```
TallTed, line 1335:
```suggestion
            affordance was provided in a previous step (e.g., if [=select an account=] was
            invoked). If that affordance is triggered:
            1. [=Show an IDP login dialog=] with |config|, |provider|, and |globalObject|.
```

TallTed (changes requested):

spec/index.bs
domenic: This would not be correct. Please check the definition of "in parallel": https://html.spec.whatwg.org/multipage/infrastructure.html#in-parallel

TallTed: The definition of "in parallel" is not sufficient for me (and I have some familiarity with the subject) to know whether the numbered steps (1)-(20) of which this Assert is (1) are to run *in parallel* with each other (which appears possible, and indeed, the most likely meaning) *or* that this sequence of 20 steps is to occur *in parallel* with other operations (sequences, single steps, etc.) in the spec.

Presuming that one goal of writing this specification is to have it be comprehensible by readers who are new to the subject, I suggest that some rewording would be helpful.

domenic: A PR to the HTML spec is welcome, but I think it's pretty clear personally:

> means those steps are to be run, one after another

TallTed: "In parallel" does not mean "in sequence", no matter what spec is written as if it does.

Also, I highly doubt that most readers of this spec will go to the HTML spec, so while I may well submit a PR there (if I can find where to do so; it wasn't obvious when I looked for it yesterday), I think it better to provide more clarity in the FedCM spec.

cbiesinger: For better or worse, "in parallel" is an existing term in specland. Ted, do you have a specific suggestion for how to improve this?

In general we should assume that readers either know what these terms mean or click through to their definition, IMO.

cbiesinger: (https://github.com/whatwg/html/blob/main/source is where you would send PRs for the HTML spec)

TallTed: Unfortunately, GitHub's web interface (my primary tool) can't handle docs the size of that WHATWG source. But, I was able to view it, and noted the "parallel queue".
```
<p>A <dfn export>parallel queue</dfn> represents a queue of algorithm steps that must be run in 
series.</p>
```
I think that's a more apt description of this 20-step sequence. Something like 
```suggestion
    1. Assert: These steps comprise a [=parallel queue=] that should run [=in parallel=] with any other active algorithms and/or [=parallel queues=].
```
(That does presume that plurals of defined terms are properly handled; the last might need change from `[=parallel queues=]` to `[=parallel queue=]`.)

spec/index.bs
```diff
@@ -1364,13 +1365,14 @@ steps. This returns a boolean.
             is defined, and the |provider|'s {{IdentityProviderConfig/clientId}} is not in the list of
             |account|["{{IdentityProviderAccount/approved_clients}}"], then the user agent MUST display
             the |metadata|["{{IdentityProviderClientMetadata/terms_of_service_url}}"] link.
-        1. The user agent MAY use the {{IdentityCredentialRequestOptions/context}} to customize the
-            dialog shown.
+        1. The user agent MAY use the
+            {{IdentityCredentialRequestOptions/context}} and |options|'s
```
TallTed, line 1369: Should this not be as above (lines 909–910), i.e., —
```suggestion
        1. The user agent MAY use the |options|'s
            {{IdentityCredentialRequestOptions/context}} and |options|'s
```

hlflanagan:
Discussed on https://github.com/fedidcg/meetings/blob/main/2024/2024-11-26-notes.md. Needs further review.

spec/index.bs
TallTed: @domenic — Your thumb-down emoji doesn't communicate even as well as the "in parallel" that is meant to be understood as some language other than English. If you have an argument against my points, please do me the courtesy of writing it out.

wseltzer:
Discussed at January 14, 2025 meeting [minutes]

bvandersloot-mozilla:
Why does this behavior need to be gated behind an IDP opt in?

cbiesinger:
> Why does this behavior need to be gated behind an IDP opt in?

As mentioned in one of the meetings, some IDPs do not support multiple accounts for one person so this would not be useful for them.

spec/index.bs
TallTed: I am at a loss. @domenic, you are not participating usefully. Please refrain from emoting unexplained "thumb down". If you have an argument against what I've said, please put into words, so I and others can consider your point(s) against my own.

bvandersloot-mozilla:
Follow up: is there a reason it varies between active and passive if it is expected to be per-IDP?

cbiesinger:
We are happy to make this a single toplevel boolean `supports_use_other_account`. I will update the PR accordingly.

spec/index.bs
cbiesinger: Because I am not touching those lines (anymore), I do not want to change that as part of this PR but I can do that separately.

cbiesinger:
@bvandersloot-mozilla @npm1 please take another look, I have updated the PR to use a single toplevel boolean.

spec/index.bs
cbiesinger: Actually, that algorithm does not have an |options| parameter, so this is not a simple change.

spec/index.bs
```diff
@@ -914,18 +917,25 @@ the exception thrown.
...
+            [=compute the connection status/connected=]:
+            1. Show a dialog to request user permission to sign in via |account|, and set the result
+                in |permission|. The user agent MAY use |options|'s
+                {{IdentityCredentialRequestOptions/context}} and |options|'s
+                {{IdentityCredentialRequestOptions/mode}} to customize the
+                dialog.
+            1. If |supportsUseOtherAccount| is true, that dialog MUST provide
+                an affordance to use another account. If that affordance is triggered:
+                1. [=Show an IDP login dialog=] with |config|, |provider|, and |globalObject|.
+                1. If that returned success, go back to the [=fetch accounts step=].
```
npm1, line 929: Should there be an otherwise here?

spec/index.bs
```diff
@@ -1370,10 +1382,16 @@ dictionary IdentityAssertionResponse {
...
+To <dfn>select an account</dfn> given an |accountsList|, a boolean
+|supportsUseOtherAccount|, an {{IdentityProviderAPIConfig}} |config|, an
+{{IdentityProviderConfig}} |provider|, and a |globalObject|, run the following
+steps. This returns an {{IdentityProviderAccount}} or failure.
     1. Assert |accountsList|'s [=list/size=] is greater than 1.
     1. Display an account chooser displaying the options from |accountsList|.
+    1. If |supportsUseOtherAccount| is true, the account chooser MUST provide
+        an affordance to use another account. If that affordance is triggered:
+        1. [=Show an IDP login dialog=] with |config|, |provider| and |globalObject|.
+        1. If that returned success, go back to the [=fetch accounts step=].
```
npm1, line 1394: Should there be an otherwise here?

spec/index.bs
```diff
@@ -1426,6 +1446,12 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
             1. The user agent MAY use the
                 {{IdentityCredentialRequestOptions/context}} and |provider|'s
                 {{IdentityCredentialRequestOptions/mode}} to customize the dialog shown.
+            1. If |supportsUseOtherAccount| is true, the account chooser MUST
```
npm1, line 1449: I wonder if this should instead be covered by showing an initial step to select an account if this is true, otherwise go direct to request permission?

spec/index.bs
cbiesinger: Done

spec/index.bs
cbiesinger: Done.

spec/index.bs
cbiesinger: You're suggesting that we should require showing the account chooser first if "supports use other account" is true, even if there is only one account signed in to the IDP?

npm1: Yea... that matches the Chrome implementation (for active mode which the only case we are keeping anyways)

bvandersloot-mozilla: That sounds reasonable to me.

spec/index.bs
```diff
@@ -949,6 +949,14 @@ the exception thrown.
                 1. Otherwise, go back to the [=fetch accounts step=] to get an updated
                     value of |providerMap| for this [=IDP=].
         1. Otherwise, |value| is a [=list=] of accounts. [=list/Extend=] |allAccounts| with |value|.
+        1. If |options|.{{IdentityCredentialRequestOptions/mode}} is `"active"` and the provider's
+            |config|.{{IdentityProviderAPIConfig/supports_use_other_account}} is true, add an
+            affordance to trigger [=show an IDP login dialog=] to let the user sign in to another
+            account. If that affordance is triggered:
+                1. Let |result| be the result of that algorithm.
+                1. If |result| is failure, go back to [=show accounts=].
```
npm1: This is not 'back', this step is after. Is this right?

spec/index.bs
```diff
@@ -949,6 +949,14 @@ the exception thrown.
...
+                1. Otherwise, go back to the [=fetch accounts step=] to get an updated
```
npm1, line 958: Hmm I guess this works in that this feature only works in the spec currently with a single account, but we would need to make this cleaner when we add multi IDP support I think.

spec/index.bs
```diff
@@ -985,6 +997,11 @@ An extension may use the following instead of the [=create identity credential/s
         1. Otherwise:
             1. Show UI to allow the user to select an account chooser displaying the options from
                 <var ignore="">accountsList</var>.
+            1. If |supportsUseOtherAccount| is true, the account chooser MUST provide
+                an affordance to use another account. If that affordance is triggered:
+                1. [=Show an IDP login dialog=] with |config|, |provider| and |globalObject|.
+                1. If that returned success, go back to the [=fetch accounts step=].
+                1. Otherwise, go back to the account chooser.
```
npm1: What is the 'account chooser'?

spec/index.bs
cbiesinger: Eh. Because this gets triggered by user action at a later point, it is going back temporally. But I've removed the "back" here.

spec/index.bs
cbiesinger: Yeah maybe. Possibly something else should be labelled "fetch accounts step". Anyway I'll resolve this since it sounds you are fine with this for now.

spec/index.bs
cbiesinger: The thing mentioned on line 998, right above (which I just noticed is not a grammatically correct sentence). But OK, I added a dfn and a link to the dfn.

npm1 (approved):
lgtm. Awaiting review from Ben/Mozilla

spec/index.bs
cbiesinger: I am going to resolve this thread because this PR no longer touches that section and "in parallel" was largely orthogonal to this change anyway.

spec/index.bs
```diff
@@ -983,8 +995,13 @@ An extension may use the following instead of the [=create identity credential/s
                 and |globalObject|. Also set |permissionRequested| to true if the user agent
                 [=supports showing a permission prompt=].
         1. Otherwise:
-            1. Show UI to allow the user to select an account chooser displaying the options from
+            1. Show <dfn>UI to allow the user to select an account</dfn> displaying the options from
                 <var ignore="">accountsList</var>.
+            1. If |supportsUseOtherAccount| is true, the account chooser MUST provide
```
bvandersloot-mozilla:
```suggestion
            1. If |supportsUseOtherAccount| is true, the account chooser SHOULD provide
```
This seems a bit better, but feel free to drop it.

cbiesinger:
Thanks, done. @npm1 I think this is ready to merge!

PR #686, opened 2024-12-10
Add single-get multi IDP support for passive mode

npm1:
Adds support for multiple IDPs on passive mode by:
* Not rejecting if the list passed is not of size 1.
* Still rejecting if the list passed is not of size 1 and mode is active (it is possible this PR is later updated to remove this restriction)
* Performing some per-IDP checks first, and computing a map based on the results
* Showing UI for the relevant IDPs based on the results from the computations: optional UIs for logged out IDPs, mandatory UI for mismatch IDPs, and all accounts for IDPs that returned some accounts.
* Add a `configURL` to `IdentityCredential` so the RP knows which IDP the credential corresponds to.

Also some unrelated errors noticed while drafting this PR are fixed.

Relevant issue: https://github.com/w3c-fedid/multi-idp/issues/2

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/pull/686.html" title="Last updated on Mar 25, 2025, 2:23 PM UTC (eba0ee9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/686/382d5d4...eba0ee9.html" title="Last updated on Mar 25, 2025, 2:23 PM UTC (eba0ee9)">Diff</a>


npm1:
ptal

spec/index.bs
```diff
@@ -780,65 +778,119 @@ The <a>create an IdentityCredential</a> algorithm invokes the various FedCM fetc
...
-            1. If |result| is failure, return (failure, true).
-        1. Otherwise, the user agent MUST do one of the following:
-            * Return (failure, false).
-            * Prompt the user whether to continue. If the user continues, the user
-                agent SHOULD set |loginStatus| to [=unknown=]. This MAY include an
+    1. Let |mediation| be |options|'s {{CredentialRequestOptions/mediation}}.
+    1. Let |providerMap| be a new [=map=].
+    1. For each |provider| in |providerList|:
+        Note: The fetches performed in these steps to populate |providerMap| can be peformed at the
+            same time, e.g. the second [=IDP=]'s fetches can be started before the first ones finish,
```
TallTed, line 800:
```suggestion
            same time, e.g., the second [=IDP=]'s fetches can be started before the first ones finish,
```

spec/index.bs
```diff
@@ -780,65 +778,119 @@ The <a>create an IdentityCredential</a> algorithm invokes the various FedCM fetc
...
+        1. If |providerMap| [=map/contains=] |providerOrigin|, return (failure, true).
+        1. Let |loginStatus| be the result of [=get the login status=] with |providerOrigin|.
+        1. If |loginStatus| is [=unknown=], a user agent MAY set it to [=logged-out=].
+        1. If |loginStatus| is [=logged-out=]:
+            1. If |mode| is {{IdentityCredentialRequestOptionsMode/active}}:
+                1. Let |result| be the result of running
+                    [=fetch the config file and show an IDP login dialog=] with
+                    |provider| and |globalObject|.
+                1. If |result| is failure, return (failure, true).
+            1. Otherwise, set |providerMap|[|providerOrigin|] to "logged-out" and [=continue=].
```
TallTed, line 812: `set |providerMap|[|providerOrigin|] to` appears to be incorrect markup or punctuation, though similar occurs a few times (also flagged, below) in this PR.

It results in 
> set _providerMap_[_providerOrigin_] to

I'm not familiar enough with `.bs` to correct it.

spec/index.bs
```diff
@@ -780,65 +778,119 @@ The <a>create an IdentityCredential</a> algorithm invokes the various FedCM fetc
...
+                A user agent may decide to skip this step if no credentials were
+                sent to server.
+
+                Note: For example, if the fetch failed due to a DNS error, no
+                    credentials were sent and therefore the [=IDP=] did not learn
+                    the user's identity. In this situation, we do not know whether
+                    the user is signed in or not and so we may not want to reset
+                    the status.
+            1. <dfn>Mismatch dialog step</dfn>: If |loginStatus| is [=logged-in=], set
+                |providerMap|[|providerOrigin|] to "mismatch" and [=continue=].
```
TallTed, line 832: Again, `set |providerMap|[|providerOrigin|] to` appears likely to be incorrect markup or punctuation.

spec/index.bs
```diff
@@ -780,65 +778,119 @@ The <a>create an IdentityCredential</a> algorithm invokes the various FedCM fetc
...
+        1. Assert: |accountsList| is not failure and the size of |accountsList| is not 0.
+        1. [=Set the login status=] for |providerOrigin| to [=logged-in=].
+        1. For each |acc| in |accountsList|:
+            1. If |acc|["{{IdentityProviderAccount/picture}}"] is present, [=fetch the account picture=]
+                with |acc| and |globalObject|. If the [=user agent=] displays this picture to
+                the user at any point, it MUST reuse the result of this fetch instead of redownloading
+                the picture.
```
TallTed, line 839: Again, `If |acc|["{{IdentityProviderAccount/picture}}"] is` appears likely to be incorrect markup or punctuation.

spec/index.bs
```diff
@@ -780,65 +778,119 @@ The <a>create an IdentityCredential</a> algorithm invokes the various FedCM fetc
...
+
+            Note: We require downloading the pictures here before we potentially filter the account
+                list so that the identity provider cannot determine what hints were provided
+                based on which fetches occurred.
```
TallTed, line 843: I think this is intended to be within the immediately preceding (1), not its parent, so it wants more indent.
```suggestion
               Note: We require downloading the pictures here before we potentially filter the account
                   list so that the identity provider cannot determine what hints were provided
                   based on which fetches occurred.
```

TallTed (changes requested):

spec/index.bs
npm1: done

spec/index.bs
npm1: done

spec/index.bs
```diff
@@ -731,25 +735,19 @@ algorithm is invoked, the user agent MUST execute the following steps. This retu
 {{IdentityCredential}} (or throws an error to the caller).
 
     1. Assert: These steps are running [=in parallel=].
-    1. If the [=list/size=] of
```
cbiesinger, line 734: Shouldn't this algorithm still check that the list is not empty?

spec/index.bs
```diff
@@ -780,65 +778,119 @@ The <a>create an IdentityCredential</a> algorithm invokes the various FedCM fetc
...
-                |provider| and |globalObject|.
-            1. If |result| is failure, return (failure, true).
-        1. Otherwise, the user agent MUST do one of the following:
-            * Return (failure, false).
-            * Prompt the user whether to continue. If the user continues, the user
-                agent SHOULD set |loginStatus| to [=unknown=]. This MAY include an
+    1. Let |mediation| be |options|'s {{CredentialRequestOptions/mediation}}.
+    1. Let |providerMap| be a new [=map=].
+    1. For each |provider| in |providerList|:
+        Note: The fetches performed in these steps to populate |providerMap| can be peformed at the
```
cbiesinger, line 799: peformed -> performed

spec/index.bs
```diff
@@ -780,65 +778,119 @@ The <a>create an IdentityCredential</a> algorithm invokes the various FedCM fetc
...
+            1. [=Set the login status=] of |providerOrigin| to [=logged-out=].
+                A user agent may decide to skip this step if no credentials were
+                sent to server.
+
+                Note: For example, if the fetch failed due to a DNS error, no
+                    credentials were sent and therefore the [=IDP=] did not learn
+                    the user's identity. In this situation, we do not know whether
+                    the user is signed in or not and so we may not want to reset
+                    the status.
+            1. <dfn>Mismatch dialog step</dfn>: If |loginStatus| is [=logged-in=], set
```
cbiesinger, line 831: Hmm this `<dfn>` has to move somewhere else I think? I don't think the "goto the mismatch dialog step" references will be right otherwise (an infinite loop?).

Alternatively, maybe those places can just directly set map[origin] to mismatch

spec/index.bs
```diff
@@ -780,65 +778,119 @@ The <a>create an IdentityCredential</a> algorithm invokes the various FedCM fetc
...
+                1. If {{IdentityProviderRequestOptions/domainHint}} is "any":
+                    1. If |account|'s {{IdentityProviderAccount/domain_hints}} is empty, remove
+                        |account| from |accountList|.
+                1. Otherwise, remove |account| from |accountList| if |account|'s
+                    {{IdentityProviderAccount/domain_hints}} does not [=list/contain=] |provider|'s
+                    {{IdentityProviderRequestOptions/domainHint}}.
+            1. If |accountList| is now empty, go to the [=mismatch dialog step=].
+        1. Set |providerMap|[|providerOrigin|] to |accountsList|.
+    1. If |providerMap| is [=map/is empty=], return (failure, false).
+    1. The user agent will now create a prompt to the user by building on the values from |providerMap|.
```
cbiesinger, line 860: I can't tell, is this an informal note describing what the next steps do, or something normative?

spec/index.bs
```diff
@@ -780,65 +778,119 @@ The <a>create an IdentityCredential</a> algorithm invokes the various FedCM fetc
...
-            Note: For example, if the fetch failed due to a DNS error, no
-                credentials were sent and therefore the [=IDP=] did not learn
-                the user's identity. In this situation, we do not know whether
-                the user is signed in or not and so we may not want to reset
-                the status.
-        1. <dfn>Mismatch dialog step</dfn>: If |loginStatus| is [=logged-in=], show a
-            dialog to the user. The contents of this dialog are defined by the user
-            agent. This dialog SHOULD provide an affordance for the user to trigger
-            the [=show an IDP login dialog=] algorithm with |config| and |provider|;
+        1. Otherwise, if |value| is "mismatch", add contents indicating this [=IDP=]
```
cbiesinger, line 891: The new phrasing seems a bit odd (what does "indicating this IDP" mean here?)

spec/index.bs
```diff
@@ -851,6 +903,8 @@ the exception thrown.
 
             1. Wait until one of the following occurs:
 
+                * If the user selects another option, do nothing.
```
cbiesinger, line 906: What other option...?

spec/index.bs
```diff
@@ -859,79 +913,45 @@ the exception thrown.
...
-            1. Set |permissionRequested| to true.
+                    1. Otherwise, go back to the [=fetch accounts step=] to get an updated
+                        value of |providerMap| for this [=IDP=].
+        1. Otherwise, |value| is a [=list=] of accounts. All accounts are aggregated into a single
+            |accountsList| and these are then presented to the user as follows:
+            1. If |accountsList|'s size is 1:
+                1. Set |selectedAccount| to |accountsList|[0].
+                1. If [=compute the connection status=] of |selectedAccount|, the relevant |provider|,
+                    and |globalObject| returns [=compute the connection status/connected=], show a
+                    dialog to request user permission to sign in via |selectedAccount|, and set the
```
cbiesinger, line 924: This is not right, because there may be more accounts in the other IDPs.

Is this section assuming that `value` is a list of accounts from all IDPs combined? That is not what the algorithm does, and it is incorrect even if so because you should not go directly to request permission if some IDPs are in mismatch state (right?)

spec/index.bs
npm1: I think it still works but the goto is not really needed anymore so I removed this dfn and made the actual thing explicit since it's clearer.

spec/index.bs
npm1: It is informative. Since it is confusing I removed it

spec/index.bs
npm1: Changed to `|providerOrigin|`

spec/index.bs
npm1: Any other UI option. I simplified this to only mention what happens if this option is selected for each UI option.

spec/index.bs
npm1: Yea they are combined. Made this more explicit and added the mismatch condition

spec/index.bs
```diff
@@ -849,89 +910,53 @@ the exception thrown.
...
+                {{IdentityCredentialRequestOptions/mode}} to customize the dialog.
+            1. Otherwise, let |permission| be the result of running [=request permission to sign-up=]
+                algorithm with |selectedAccount|, the relevant |config|, the relevant |provider|,
+                and |globalObject|. Also set |permissionRequested| to true if the user agent
+                [=supports showing a permission prompt=].
         1. Otherwise:
-            1. Let |permission| be the result of running the [=request permission to sign-up=]
-                algorithm with |account|, |config|, |provider|, and |globalObject|.
-            1. Set |permissionRequested| to true.
+            1. Set |selectedAccount| to the result of running the [=select an account=] from |allAccounts|.
```
cbiesinger, line 937: There is a bit of a mismatch between the algorithm here and "select an account". The latter displays an account chooser, but you already built UI here.

I am actually not sure that "select an account" is still useful. It is a pretty short algorithm and you already have the UI built here. So maybe this should just be "Wait for the user to select an account" or something, and then set |selectedAccount| acccordingly.

spec/index.bs
```diff
@@ -849,89 +910,53 @@ the exception thrown.
...
+                [=create a connection between the RP and the IdP account=] with |provider|, |account|,
+                and |globalObject|, and set |permission| to true.
+                Note: The connection would normally be created in the [=request permission to sign-up=]
+                    algorithm, but we do not want to show an extra dialog in this case.
+            1. Otherwise:
+                1. Let |permission| be the result of running the [=request permission to sign-up=]
+                    algorithm with |selectedAccount|, the relevant |config|, the relevant |provider|,
+                    and |globalObject|.
+                1. Set |permissionRequested| to true.
+    1. Also include an affordance to close the dialog. If the user closes this dialog, return (failure,
```
cbiesinger, line 952: This should probably be before "1. If |allAccounts| is not [=list/empty=], add UI to present the account options to the user as follows:", because those steps have various things that wait for user input.

spec/index.bs
npm1: Ok, moved it here and fixed the issue

spec/index.bs
npm1: I think there I've removed waits in the previous steps. What there are are steps that are to be executed if a certain UI selection is made by the user, which is covered by the wait in the following step.

cbiesinger: In the size 1 case, a wait is implied in "request permission to sign up" (because it returns an account), and `show a dialog to request user permission to sign in via |selectedAccount|, and set the result in |permission|` also implies a wait

npm1: Oh yes I see what you mean. Moved up

wseltzer:
Discussed at Jan 21, 2025 meeting [minutes](https://github.com/fedidcg/meetings/blob/main/2025/2025-01-21-notes.md#notes).

spec/index.bs
```diff
@@ -849,89 +910,56 @@ the exception thrown.
...
+                {{IdentityCredentialRequestOptions/mode}} to customize the dialog.
+            1. Otherwise, let |permission| be the result of running [=request permission to sign-up=]
+                algorithm with |selectedAccount|, the relevant |config|, the relevant |provider|,
+                and |globalObject|. Also set |permissionRequested| to true if the user agent
+                [=supports showing a permission prompt=].
         1. Otherwise:
-            1. Let |permission| be the result of running the [=request permission to sign-up=]
-                algorithm with |account|, |config|, |provider|, and |globalObject|.
-            1. Set |permissionRequested| to true.
+            1. Assert |allAccounts|'s [=list/size=] is greater than 1.
```
cbiesinger, line 939: I actually think this assert is wrong -- it is possible that size==1 if providerMap contains mismatch

spec/index.bs
```diff
@@ -849,89 +910,56 @@ the exception thrown.
...
+            1. Show UI to allow the user to select an account chooser displaying the options from
+                |accountsList|.
+            1. If the user selects an account, perform the following steps:
```
cbiesinger, line 942: this needs an else?

spec/index.bs
```diff
@@ -849,89 +910,56 @@ the exception thrown.
...
+                1. Let |selectedAccount| be the {{IdentityProviderAccount}} chosen.
```
cbiesinger, line 943: does this have to be "Set" instead of let? AIUI Let is only for declaring variables.

spec/index.bs
```diff
@@ -934,23 +936,24 @@ the exception thrown.
...
+                    and |globalObject|, and set |permission| to true.
+                    Note: The connection would normally be created in the [=request permission to sign-up=]
+                        algorithm, but we do not want to show an extra dialog in this case.
+                1. Otherwise:
+                    1. Let |permission| be the result of running the [=request permission to sign-up=]
+                        algorithm with |selectedAccount|, the relevant |config|, the relevant |provider|,
+                        and |globalObject|.
+                    1. Set |permissionRequested| to true.
     1. Wait until the [=user agent=]'s dialogs requesting for user choice or permission to be
         closed, if any are created in the previous steps.
```
TallTed, line 958:
```suggestion
    1. If [=user agent=] created any dialogs requesting user choice or permission in the previous steps,
        wait until they are closed.
```

TallTed (changes requested):

spec/index.bs
npm1: The 'else' would be handled by the wait until the dialog is closed

cbiesinger: In that case, you need to do something about the assert on line 958

cbiesinger (approved):

spec/index.bs
npm1: Ah yea needs to go after the permission check, thanks.

bvandersloot-mozilla:
This overall seems reasonable, but raises questions for me about how a browser with an IDP picker integrates with the spec. Right now it works with single IDP, however, I think that I want to be able to show account information in the IDP picker where there is already a linkage.  

How does this handle where a response is in the connected account set? Does it still show the dialog?

npm1:
Hey Ben, sorry not sure I follow your questions but trying to answer:

> I think that I want to be able to show account information in the IDP picker where there is already a linkage.

Regarding IDP picker with account information: sure, this is intended to be allowed. You can show any UI before fetching the accounts, so in particular you can show UI using some information that you cache from an account in the connected accounts set. However, I do worry that this effectively removes auto reauthentication since you can only really know the returning accounts after fetching. You could guess when it may happen based on the connected accounts set but that seems brittle.

> How does this handle where a response is in the connected account set? Does it still show the dialog?

Not sure I follow. If there are accounts in the connected account set, this just means it may be possible to perform auto reauthentication. Of course, the user agent is encouraged to nudge the user towards these accounts as well as accounts whose `approved_clients` imply that this is an IDP account that has already been previously used to login to the RP. But in general, unless there is exactly one account in the connected accounts set AND the RP requests auto reauthentication, then the IDP should try to perform the login automatically, whereas in any other case we let the user confirm the account they wish to use first.

cbiesinger:
@bvandersloot-mozilla is your question maybe this: does the algorithm allow fetching and displaying accounts only from IDPs who have accounts in the connected set, and only showing the IDP name for the other IDPs (fetching accounts once you click that)?

bvandersloot-mozilla:
@cbiesinger, that is precisely what I meant, thanks!

npm1:
Added support for this (during the wait, the user agent may set the provider list to a subset of itself)

npm1:
Ping @bvandersloot-mozilla. Let me know if you want me to add active mode support here as well or any other comments so I can merge this

bvandersloot-mozilla:
It seems like we should add active multi-idp, however this is probably reasonable to merge as is

wseltzer:
Discussed [March 25](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-03-25-notes.md)

PR #711, opened 2025-03-31
Set providers size limit

npm1:
Fixes https://github.com/w3c-fedid/multi-idp/issues/6

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -794,6 +794,7 @@ the exception thrown.
             {{IdentityCredentialRequestOptionsMode/passive}} on |globalObject|'s [=Window/navigable=]'s
             [=navigable/top-level traversable=] or on any of its descendants, reject the pending
             request with a "{{NetworkError}}" {{DOMException}}.
+    1. If |providerList|'s [=list/size=] is greater than 10, return (failure, true).
     1. The user agent MAY wait an arbitrary amount of time before continuing with the next steps.
         The user agent MAY also return (failure, false) at this point or after some arbitrary wait.
         The user agent MAY also set |providerList| to a subset of itself after some arbitrary wait.


cbiesinger (approved):

spec/index.bs
```diff
@@ -794,6 +794,7 @@ the exception thrown.
             {{IdentityCredentialRequestOptionsMode/passive}} on |globalObject|'s [=Window/navigable=]'s
             [=navigable/top-level traversable=] or on any of its descendants, reject the pending
             request with a "{{NetworkError}}" {{DOMException}}.
+    1. If |providerList|'s [=list/size=] is greater than 10, return (failure, true).
```
TallTed, line 797: This seems arbitrarily small. I understand some of the risks if it is *too* large, but it seems to me it should not be hard-coded at the spec level. Relevant admin users should be able to adjust this as appropriate to a given deployment.

npm1: We can add enterprise settings in order to make this larger for admins that control browser deployments if needed. But I think it is more interoperable to have a concrete specific (and yes, arbitrary) value here so that the behavior is consistent across browsers and developers do not accidentally request a number of IdPs which works on some browsers but not on others.

TallTed: What happens when there are ten large IdPs and some number of small? Or is there an unstated goal to limit the overall number of IdPs?

npm1: Not sure what the size of the IdP has to do with anything? We want to limit the number of IdPs that a single RP can request in one go. Each IdP in the request can potentially introduce 5+ fetches so it would be too slow to include too many IdPs.

TallTed: Imagine, limiting the size of the IdP pool to 5 providers. Who might they be? Probably GoogleIdP, XIdP, AppleIdP, MicrosoftIdP, maybe GitHubIdP. What about JoeSmithIdP? What about DisneyIdP?

npm1: The RP needs to pick their IdPs that FedCM can work with. This often involves an SDK from each to handle the federated login. Are there examples of sites today embedding SDKs from more than 10 IDPs?

TallTed: > Are there examples of sites today embedding SDKs from more than 10 IDPs?

Today? Surely not. The FedCM specification process is still early.

Going forward? Let several hundred IdPs bloom, globally, and let a new one pop up whenever a potential provider sees a suitable pool of clients/customers/users.

It seems problematic to me to specify FedCM as such a tiny oligopoly. It's *somewhat* better than the current set of silos with overlaps at the IdP/RP layer and different user experience from each IdP, but only somewhat.

npm1: Federation is not only done through FedCM, so the question about today is about the current state of the world.

I'll reiterate that what you are imagining is not feasible without IDP registration, because the RP would need to embed one SDK per IDP, and we'd need to perform multiple fetches per IDP. So a `get()` with hundreds of IDPs is not feasible.

This limit is intended to only be limiting to people attempting to abuse the API. We can always change it or increase it if there are real-world requests where they would need a larger limit. Today, the largest I've seen is 5, so the limit I am proposing is double that.

bvandersloot-mozilla: But IDP registration is something we want to do, and that use case is entirely boxed out by this approach to limiting only to 10.

TallTed:
I think this issue's title should be changed from `Set providers size limit` to `Set limit on ID provider count` or similar.

npm1:
> I think this issue's title should be changed from `Set providers size limit` to `Set limit on ID provider count` or similar.

I think your suggestion is unclear. Let me set providers in quote, does that help

spec/index.bs
bvandersloot-mozilla: Ah, I see, this is the provided idp config list, not the actual set that shows up in the UI. I was a little turned around. Then I don't have strong opinions against this, but it really isn't needed in Firefox because of the IDP chooser.

npm1: Ok. We don't need to add this to the spec, as I think the spec already provides enough leeway for the user agent to reject a request for whatever reason. But I figured it might be helpful to have a shared limit to reduce possible compat issues.

TallTed:
"Set `providers` size limit" == "Set limit on _size_ of `providers`", but this is not what is meant.

"Set limit on _number_ of `providers`" is what is meant, and should be the title.

TallTed:
> "Set limit on size of providers" is exactly what this PR does

No. "Set limit on size of *list of* providers" or "Set limit on size of *set of* providers"  is exactly what this PR does.

npm1:
https://w3c-fedid.github.io/FedCM/#dom-identitycredentialrequestoptions-providers

hlflanagan:
Discussed during the FedID WG/CG meeting 22 April 2025 <https://github.com/w3c-fedid/meetings/blob/main/2025/2025-04-22-FedCM-agenda.md>

hlflanagan:
Discussed during the FedID WG/CG meeting 22 April 2025 <https://github.com/w3c-fedid/meetings/blob/main/2025/2025-04-22-FedCM-agenda.md>

Issue #703, opened 2025-03-06
FedCM "core": move "extensions" out of the current spec into an independent spec


samuelgoto:
As we've been discussing in the CG/WG the last few calls, a few of us started looking into a possible subset of the FedCM specification, which we've been calling "core": a subset of FedCM that 2 independent implementations (in this specific exercise, Chrome and Firefox) agree to implement and ship.

We have been debating exactly what that means, but one good starting point is to "move" features in the current spec that are "clearly" not in "core" (as defined as a feature that any implementation chooses not to ship) into a separate spec, lets call it "extensions".

There are probably more, but here is a concrete list that came up:

- [`passive mode`](https://w3c-fedid.github.io/FedCM/#dom-identitycredentialrequestoptionsmode-passive)
- The Disclosure Prompt
    - The [Client Metadata Endpoint](https://w3c-fedid.github.io/FedCM/#idp-api-client-id-metadata-endpoint)
        - the [`client_metadata_endpoint`](https://w3c-fedid.github.io/FedCM/#dom-identityproviderapiconfig-client_metadata_endpoint)
        - the [`fetch the client metadata algorithm`](https://w3c-fedid.github.io/FedCM/#fetch-the-client-metadata)
    - The [Fields API](https://w3c-fedid.github.io/FedCM/#dom-identityproviderrequestoptions-fields)
    - Disclosure in the [`id_assertion_endpoint`](https://w3c-fedid.github.io/FedCM/#idp-api-id-assertion-endpoint):
        - [`disclosure_text_shown`](https://w3c-fedid.github.io/FedCM/#dom-id_assertion_endpoint_request-disclosure_text_shown) 
        - [`disclosure_shown_for`](https://w3c-fedid.github.io/FedCM/#dom-id_assertion_endpoint_request-disclosure_shown_for)
        - [`fields`](https://w3c-fedid.github.io/FedCM/#dom-id_assertion_endpoint_request-fields)
    - The [request permission to sign-up algorithm](https://w3c-fedid.github.io/FedCM/#request-permission-signup)
- anything else? 

There are a lot of other features that are being incubated (e.g. [idp registration](https://github.com/w3c-fedid/idp-registration) and [lighweight mode](https://github.com/fedidcg/LightweightFedCM)) that aren't yet in the spec, so as we review the PR we can decide in the PR itself (rather than here) whether it fits into the "core" or "extensions" specification.

[...]

PR #712, opened 2025-04-07
Move extension parts into issues in spec

npm1:
First step in https://github.com/w3c-fedid/FedCM/issues/703: mark the extension parts of the core spec as issues within the spec.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/pull/712.html" title="Last updated on Apr 9, 2025, 7:33 PM UTC (319316d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/712/c9b0418...319316d.html" title="Last updated on Apr 9, 2025, 7:33 PM UTC (319316d)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -199,9 +199,6 @@ The [[#idp-api]] defines a set of HTTP APIs that [=IDP=]s expose as well as the 
 |           |   |              |           |     |                |           |
 |           |   |              |           |     +------------->*-+ Accounts  |
 |           |   |              |           |     |                |           |
-|           |   |              |           |     +------------->*-+ Client    |
-|           |   |              |           |     |                | Metadata  |
-|           |   |              |           |     |                |           |
 |           |   |              |           |     +------------->*-+ Assertion |
 |           |   |              |           |     |                |           |
 |           |   |              |           |     |                |           |
@@ -212,6 +209,19 @@ The [[#idp-api]] defines a set of HTTP APIs that [=IDP=]s expose as well as the 
 |           |                  |           |                      |           |
 +-----------+                  +-----------+                      +-----------+
 </script>
+
+<div class="issue" heading="extension">
+An extension spec may add a client metadata endpoint in between accounts and assertion:
+<script type="text/typogram">
+|           |   |              |           |     +------------->*-+ Accounts  |
+|           |   |              |           |     |                |           |
+|           |   |              |           |     +------------->*-+ Client    |
+|           |   |              |           |     |                | Metadata  |
+|           |   |              |           |     |                |           |
+|           |   |              |           |     +------------->*-+ Assertion |
+|           |   |              |           |     |                |           |
+</script>
+</div>
 
 The user agent intermediates in such a matter that makes it impractical for the
 API to be used for tracking purposes, while preserving the functionality of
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1284,39 +1311,19 @@ The <a>fetch an identity assertion</a> algorithm is invoked after the user has g
 use FedCM with a specific [=IDP=] account. It fetches the [=identity assertion endpoint=] to obtain
 the token that will be provided to the [=RP=].
 
-<div algorithm>
+<div algorithm="fetch an identity assertion">
 To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
-    |accountId|, a boolean |permissionRequested|, a boolean |isAutoSelected|, an
+    |accountId|, a boolean <var ignore="">permissionRequested</var>, a boolean |isAutoSelected|, an
```
bvandersloot-mozilla, line 1316: 👍

spec/index.bs
```diff
@@ -1284,39 +1311,19 @@ The <a>fetch an identity assertion</a> algorithm is invoked after the user has g
...
-    1. If |fields| is not empty:
-        1. Let |fieldsString| be the entries of |fields| concatenated with a comma ("`,`")
-            between elements.
-        1. Append ("fields", |fieldsString|) to |list|.
-    1. If |disclosureShownFor| is not empty:
-        1. Let |disclosureString| be the entries of |disclosureShownFor| concatenated
-            with a comma ("`,`") between elements.
-        1. Append ("disclosure_shown_for", |disclosureString|) to |list|.
-    1. If |disclosureShownFor| contains all of "name", "email", and "picture", append
-        ("disclosure_text_shown", true) to |list|.
```
bvandersloot-mozilla, line 1314: If I remember correctly, @cbiesinger said that there is compat risk for not sending these fields. An open question is if we should or should not include these as ("fields",""), ("disclosure_shown_for",""), and ("disclosure_text_shown", false) for non-extension clients. WDYT?

spec/index.bs
```diff
@@ -2000,9 +2036,11 @@ Every {{IdentityProviderAccount}} is expected to have members with the following
     :   <dfn>picture</dfn>
     ::  URL for the account's picture.
     :   <dfn>approved_clients</dfn>
-    ::  A list of [=RP=]s (that gets matched against the requesting {{IdentityProviderConfig/clientId}}) this account is already registered with.
-        Used in the [=request permission to sign-up=] to allow the [=IDP=] to control whether to show
-        the Privacy Policy and the Terms of Service.
+    ::  A list of [=RP=]s (that gets matched against the requesting {{IdentityProviderConfig/clientId}}) this account is already registered with. Use to [=compute the connection status=] of an account.
```
bvandersloot-mozilla, line 2039:
```suggestion
    ::  A list of [=RP=]s (that gets matched against the requesting {{IdentityProviderConfig/clientId}}) this account is already registered with. Used to [=compute the connection status=] of an account.
```

bvandersloot-mozilla (commented):
One open question here, but overall, this matches the spirit of what we want. There are lots of editorial changes I might make, but normatively it looks right.

spec/index.bs
npm1: Per offline discussion, keeping the current algorithm but added an issue to discuss this change

npm1:
Please take another look

TallTed (approved):

samuelgoto (approved):
LGTM++

This is probably not the perfect delineation between what's "core" vs what's "extension", and there are probably better ways to structure this, but I do think this is a meaningful step forward that can allow us to tackle this definition in an incremental and gradual way, now that we have identified what we need to fix.

So, LGTM++ that this is a great starting point and baseline that we can work from!

bvandersloot-mozilla (approved):

PR #717, opened 2025-04-15
State that resolve() does nothing outside of continue_on

cbiesinger:
***
<a href="https://pr-preview.s3.amazonaws.com/cbiesinger/WebID/pull/717.html" title="Last updated on Apr 17, 2025, 6:08 PM UTC (410c725)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/717/c85489a...cbiesinger:410c725.html" title="Last updated on Apr 17, 2025, 6:08 PM UTC (410c725)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1736,7 +1736,7 @@ See the [=show an IDP login dialog=] algorithm for more details.
 {{IdentityProvider/resolve}} is used while a [=show a continuation
 dialog|continuation dialog=] is shown to return a specific token to the [=RP=]
 as described below. If called at any other time, reject the returned
-promise with a "{{InvalidStateError}}" {{DOMException}}.
+promise with a "{{NotAllowedError}}" {{DOMException}}.
 
 An {{IdentityUserInfo}} represents user account information from a user. This information is exposed
 to the [=IDP=] once the user has already used the FedCM API to login in the [=RP=]. That is, it is
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1735,7 +1735,8 @@ See the [=show an IDP login dialog=] algorithm for more details.
 
 {{IdentityProvider/resolve}} is used while a [=show a continuation
 dialog|continuation dialog=] is shown to return a specific token to the [=RP=]
-as described below. If called at any other time, it does nothing.
+as described below. If called at any other time, reject the returned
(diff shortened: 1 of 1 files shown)


npm1 (approved):
Link the issue associated with this PR in the description?

LGTM. Even though I think this could be specified a bit more formally, I think it's clear enough. Let's see what Ben thinks

cbiesinger:
I couldn't find the issue, do you happen to have it?

npm1:
Oh I think this was something we noticed over chat, there appears to be no issue

spec/index.bs
```diff
@@ -1733,6 +1733,10 @@ their phone number. To allow for such flows, the [=IDP=] must call
 
 See the [=show an IDP login dialog=] algorithm for more details.
 
+{{IdentityProvider/resolve}} is used while a [=show a continuation
+dialog|continuation dialog=] is shown to return a specific token to the [=RP=]
+as described below. If called at any other time, it does nothing.
```
samuelgoto, line 1738:
```suggestion
as described below. If called at any other time, it throws an Error.
```

Shouldn't we be throwing an error perhaps?

Not sure, and also, not implemented, but just wanted to raise whether "does nothing" is the right thing ...

cbiesinger: Fair point. Changed it to reject with an InvalidStateError.

npm1: Yes rejecting is better. Is this what the implementation does?

spec/index.bs
```diff
@@ -1735,7 +1735,8 @@ See the [=show an IDP login dialog=] algorithm for more details.
 
 {{IdentityProvider/resolve}} is used while a [=show a continuation
 dialog|continuation dialog=] is shown to return a specific token to the [=RP=]
-as described below. If called at any other time, it does nothing.
+as described below. If called at any other time, reject the returned
+promise with a "{{InvalidStateError}}" {{DOMException}}.
```
TallTed, line 1739:
```suggestion
promise with an "{{InvalidStateError}}" {{DOMException}}.
```

spec/index.bs
cbiesinger: After checking the implementation, it does reject with an error, but it uses NotAllowedError. I changed this PR to use that error as well.

(This makes Ted's suggestion moot)

Issue #734, opened 2025-06-07
Assert in step 7.12 of the Create Identity credential algorithm can fail


togamid:
The variable "loginStatus" is only set once in the [Create Identity Credential algorithm](https://w3c-fedid.github.io/FedCM/#create-identity-credential), in step 7.3. It may also be set in 7.4, but does not have to. This means that it can have the values "logged-in", "logged-out" and "unknown" in  step 7.11 , which deals with a failed request to the accounts endpoint. 

This is problematic in 7.11.2, which says "If loginStatus is logged-in, set providerMap[providerOrigin] to "mismatch" and continue.". If I read this correctly, the algorithm only executes the "continue" and skips to the next iteration if the original loginStatus was "logged-in". If this is correct, the algorithm advances to 7.12 if the loginStatus was "logged-out" or "unknown", where the assert fails.

It would probably be better to split 7.11.2 into three steps. One sets the providerMap to mismatch if the loginStatus was "logged-in", one sets it to "logged-out" in the other cases and the third executes the "continue".

However, this creates a privacy issue. If the user agent does not set a loginStatus of "unknown" to "logged-out" in 7.4, the algorithm can reach 7.11.3 without ever showing UI. This might lead to a silent failure in step 13 despite IdP cookies having been sent, which circumvents the last part of the answer to question one in the [Privacy Questionaire](https://github.com/w3c-fedid/FedCM/blob/main/privacy_security_questionnaire.md).

Such a change also has to be carefully reviewed for follow up consequences. One thing I already noticed is that it might allow the RP to determine whether the original login status was "unknown" by provoking a failure which skips the randomization in step 13.

Another thing I noticed is that the algorithm seems to handle the case of a providerMap that just contains "logged-out" weirdly, potentially displaying (or not displaying) empty UI in step 18 and eventually failing another assert in step 21.

npm1:
Yes, I think the `continue` in 7.11.2 needs to be its own step.

`However, this creates a privacy issue. If the user agent does not set a loginStatus of "unknown" to "logged-out" in 7.4, the algorithm can reach 7.11.3 without ever showing UI`

Yes, that is the intention of the "unknown" status: if a user agent wants to accept the tradeoff of allowing one silent failure when the very first accounts fetch occurs, then it can use this status. Otherwise, it should not use this status.

`One thing I already noticed is that it might allow the RP to determine whether the original login status was "unknown" by provoking a failure which skips the randomization in step 13.`

[...]

PR #737, opened 2025-06-12
Fix spec bugs reported by togamid

npm1:
There are a few spec bugs in the `create an IdentityCredential` algorithm. Fixes https://github.com/w3c-fedid/FedCM/issues/734. Thanks to @togamid for reporting (and potentially reviewing)!

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -833,11 +833,13 @@ the exception thrown.
         1. Let |loginStatus| be the result of [=get the login status=] with |providerOrigin|.
         1. If |loginStatus| is [=unknown=], a user agent MAY set it to [=logged-out=].
         1. If |loginStatus| is [=logged-out=]:
-            1. If |mode| is {{IdentityCredentialRequestOptionsMode/active}}:
+            1. If |mode| is {{IdentityCredentialRequestOptionsMode/active}} and |mediation| is not
+                "{{CredentialMediationRequirement/silent}}":
                 1. Let |result| be the result of running
                     [=fetch the config file and show an IDP login dialog=] with
                     |provider| and |globalObject|.
                 1. If |result| is failure, return (failure, true).
+                1. Otherwise, set |loginStatus| to [=logged-in=].
             1. Otherwise, set |providerMap|[|providerOrigin|] to "logged-out" and [=continue=].
         1. Let |requiresUserMediation| be |providerOrigin|'s [=requires user mediation=].
         1. If |requiresUserMediation| is true and |mediation| is
@@ -845,54 +847,13 @@ the exception thrown.
         1. Let |config| be the result of running [=fetch the config file=] with
             |provider| and |globalObject|.
         1. If |config| is failure, [=continue=].
-        1. <dfn>Fetch accounts step</dfn>: Let |accountsList| be the result of
-            [=fetch the accounts=] with |config|, |provider|, and |globalObject|.
-        1. If |accountsList| is failure, or the size of |accountsList| is 0:
-            1. [=Set the login status=] of |providerOrigin| to [=logged-out=].
-                A user agent may decide to skip this step if no credentials were
-                sent to server.
-
-                Note: For example, if the fetch failed due to a DNS error, no
-                    credentials were sent and therefore the [=IDP=] did not learn
-                    the user's identity. In this situation, we do not know whether
-                    the user is signed in or not and so we may not want to reset
-                    the status.
-            1. If |loginStatus| is [=logged-in=], set |providerMap|[|providerOrigin|] to "mismatch"
-                and [=continue=].
-        1. Assert: |accountsList| is not failure and the size of |accountsList| is not 0.
-        1. [=Set the login status=] for |providerOrigin| to [=logged-in=].
-        1. For each |acc| in |accountsList|:
-            1. If |acc|["{{IdentityProviderAccount/picture}}"] is present, [=fetch the account picture=]
-                with |acc| and |globalObject|. If the [=user agent=] displays this picture to
-                the user at any point, it MUST reuse the result of this fetch instead of redownloading
-                the picture.
-
-                Note: We require downloading the pictures here before we potentially filter the account
-                    list so that the identity provider cannot determine what hints were provided
-                    based on which fetches occurred.
-        1. If |provider|'s {{IdentityProviderRequestOptions/loginHint}} is not empty:
-            1. For every |account| in |accountList|, remove |account| from |accountList| if |account|'s
-                {{IdentityProviderAccount/login_hints}} does not [=list/contain=] |provider|'s
-                {{IdentityProviderRequestOptions/loginHint}}.
-            1. If |accountList| is now empty, set |providerMap|[|providerOrigin|] to "mismatch" and
-                continue.
-        1. If |provider|'s {{IdentityProviderRequestOptions/domainHint}} is not empty:
-            1. For every |account| in |accountList|:
-                1. If {{IdentityProviderRequestOptions/domainHint}} is "any":
-                    1. If |account|'s {{IdentityProviderAccount/domain_hints}} is empty, remove
-                        |account| from |accountList|.
-                1. Otherwise, remove |account| from |accountList| if |account|'s
-                    {{IdentityProviderAccount/domain_hints}} does not [=list/contain=] |provider|'s
-                    {{IdentityProviderRequestOptions/domainHint}}.
-            1. If |accountList| is now empty, set |providerMap|[|providerOrigin|] to "mismatch" and
-                continue.
-        1. If |config|.{{IdentityProviderAPIConfig/account_label}} is present:
-            1. For every |account| in |accountList|, remove |account| from |accountList| if |account|'s
-                {{IdentityProviderAccount/label_hints}} does not [=list/contain=]
-                |config|.{{IdentityProviderAPIConfig/account_label}}.
-            1. If |accountList| is now empty, set |providerMap|[|providerOrigin|] to "mismatch" and
-                continue.
-        1. Set |providerMap|[|providerOrigin|] to |accountsList|.
+        1. [=Fetch accounts=] given |config|, |provider|, and |globalObject| and let |accountsList|
+            be the result.
+        1. If |accountsList| is empty:
+            1. If |loginStatus| is [=logged-in=], set |providerMap|[|providerOrigin|] to "mismatch".
+            1. Otherwise (|loginStatus| is [=unknown=]): set |providerMap|[|providerOrigin|] to
+                "logged-out".
+        1. Otherwise, set |providerMap|[|providerOrigin|] to |accountsList|.
     1. If |providerMap| [=map/is empty=], return (failure, false).
     1. Let |registeredAccount| and |numRegisteredAccounts| be null and 0, respectively.
     1. Let |selectedAccount| be null.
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -902,31 +863,24 @@ the exception thrown.
...
-                * If the user triggers this affordance:
-                    1. Let |result| be the result of running
-                        [=fetch the config file and show an IDP login dialog=]
-                        with the relevant |provider| and |globalObject|.
-                    1. If |result| is failure, return (failure, true).
+            * Prompt the user whether to continue with this [=IDP=]. If the user continues:
+                1. Let |result| be the result of running [=show an IDP login dialog=] with the
+                    relevant |config|, |provider|, and |globalObject|.
+                1. If |result| is failure, return (failure, true).
         1. Otherwise, if |value| is "mismatch", add contents indicating this |providerOrigin|
```
togamid, line 958: I don't think the second note (new line 893) is correct any more, as step 7.11.2 sets providerMap to "logged-out" if the login Status was "unknown", which potentially ends the algorithm without UI in 16.1 even though credentials were sent to the server.
This is difficult/impractical but not impossible to do ore than once thanks to the "Clear-Site-Data" header resetting the login status.

togamid, line 958: This step also triggers and shows if mediation is "silent", thanks to the carveouts in steps 13 and 14. However, the specification of the "silent" option indicates that it will fail and not involve the user. 
It probably needs to be reviewed if it is possible to somehow make the "silent" option and the effort to always show UI if user credentials were sent compatible.

togamid: If there isn't some error in my logic and this is in fact correct, the issue should probably be discussed in a separate issue as it seems like a pretty fundamental contradiction that would need to be resolved

npm1: Added clarification that this holds when the login status was not "unknown". I don't follow the Clear-Site-Data comment. Yes it would set the status back to unknown (for browsers supporting that status), but it will also clear all credentials from that origin, so the next 'credentialed fetch' would be trivial.

npm1: For mediation:silent, the idea is to try to not involve the user. This is the only case where it is needed to avoid a silent timing attack. Which contradiction are you referring to, the fact that silent shows UI?

togamid: Yes. The description of silent says that it does not try to contact the user and instead fails. The note says that regardless of what happens, it always shows UI to the user. Those two descriptions/goals seem like they contradict each other.

togamid: If a user visits the tracker website, they might safe the cookie as a JS variable and immediately reset it after it has been cleared and the login status was set to "unknown". A similar effect could be achieved by sending two HTTP requests. One answer resets the login status with the "Clear-Site-Data" header. The other is delayed a bit and immediately resets the cookie. Both attacks involve the user visiting the tracker site, but it can not just be assumed that a tracker can only use the "unknown" path once.

npm1: Ah ok, I think that is not ideal. Filed https://github.com/w3c-fedid/FedCM/issues/746 to tackle separately, feel free to chime in there regarding that issue.

npm1: We discussed in the FedID CG call. Updated to remove the mismatch UI if mediation was silent. This is ok privacy-wise because we would only fetch when there were accounts in the connected accounts set. Which BTW I think was not something in the spec before, so added it. Let me know what you think! (not sure you were in the call)

spec/index.bs
```diff
@@ -833,66 +833,27 @@ the exception thrown.
...
-            1. If |accountList| is now empty, set |providerMap|[|providerOrigin|] to "mismatch" and
-                continue.
-        1. If |config|.{{IdentityProviderAPIConfig/account_label}} is present:
-            1. For every |account| in |accountList|, remove |account| from |accountList| if |account|'s
-                {{IdentityProviderAccount/label_hints}} does not [=list/contain=]
-                |config|.{{IdentityProviderAPIConfig/account_label}}.
-            1. If |accountList| is now empty, set |providerMap|[|providerOrigin|] to "mismatch" and
-                continue.
-        1. Set |providerMap|[|providerOrigin|] to |accountsList|.
+        1. [=Fetch accounts=] given |config|, |provider|, and |globalObject| and let |accountsList|
```
```suggestion
        1. [=Fetch accounts=] given |config|, |provider|, and |globalObject|, and let |accountsList|
```

spec/index.bs
```diff
@@ -902,31 +863,24 @@ the exception thrown.
...
                 and set |requiresUserMediation| be |providerOrigin|'s [=requires user mediation=].
+    1. Let |permission|, |permissionRequested|, and |isAutoSelected| be set to false.
     1. If |mediation| is not "{{CredentialMediationRequirement/required}}", |requiresUserMediation|
         is false, |numRegisteredAccounts| is equal to 1, and |providerMap|'s [=map/values=] do not
         [=map/contain=] "mismatch":
-        1. Set |selectedAccount| to |registeredAccount| and |permission| to true. When doing this, the user
-            agent MAY show some UI to the user indicating that they are being
-            <dfn>auto-reauthenticated</dfn>.
-        1. Set |isAutoSelected| to true.
+        1. Set |selectedAccount| to |registeredAccount| and set |permission| and |isAutoSelected| to
```
```suggestion
        1. Set |selectedAccount| to |registeredAccount|, and set |permission| and |isAutoSelected| to
```

spec/index.bs
```diff
@@ -936,27 +890,33 @@ the exception thrown.
...
             * If the [=show an IDP login dialog=] algorithm was triggered:
 
                 1. Let |result| be the result of that algorithm.
                 1. If |result| is failure, return (failure, true). The user
                     agent MAY show a dialog to the user before or after
                     returning failure indicating this failure.
-                1. Otherwise, go back to the [=fetch accounts step=] to get an updated
-                    value of |providerMap| for this [=IDP=].
+                1. Otherwise, [=fetch accounts=] with the relevant |config|, |provider|, and
+                    |globalObject| and let |accountsList| be the result.
```
```suggestion
                    |globalObject|, and let |accountsList| be the result.
```

spec/index.bs
```diff
@@ -965,6 +925,52 @@ the exception thrown.
     1. Return |credential|.
 </div>
 
+<div algorithm="fetch accounts">
+When asked to <dfn for="create identity credential">fetch accounts</dfn> given an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}} |provider|, and a |globalObject|, run the following steps. This returns a possibly empty list of accounts.
```
```suggestion
When asked to <dfn for="create identity credential">fetch accounts</dfn> given
an {{IdentityProviderAPIConfig}} |config|,
an {{IdentityProviderRequestOptions}} |provider|,
and a |globalObject|,
run the following steps. This returns a possibly empty list of accounts.
```

spec/index.bs
```diff
@@ -965,6 +925,52 @@ the exception thrown.
...
+  1. If |accountsList| is failure, or the size of |accountsList| is 0:
+      1. [=Set the login status=] of |providerOrigin| to [=logged-out=].
+          A user agent may decide to skip this step if no credentials were
+          sent to server.
+
+          Note: For example, if the fetch failed due to a DNS error, no
+              credentials were sent and therefore the [=IDP=] did not learn
+              the user's identity. In this situation, we do not know whether
+              the user is signed in or not and so we may not want to reset
+              the status.
```
TallTed, line 1044: There should be some styling applied to this `Note:`, so it is set aside from the list item of which it is a part — similarly to what has been done manually for the HTML source.

spec/index.bs
```diff
@@ -965,6 +925,52 @@ the exception thrown.
...
+  1. [=Set the login status=] for |providerOrigin| to [=logged-in=].
+  1. For each |acc| in |accountsList|:
+      1. If |acc|["{{IdentityProviderAccount/picture}}"] is present, [=fetch the account picture=]
+          with |acc| and |globalObject|. If the [=user agent=] displays this picture to
+          the user at any point, it MUST reuse the result of this fetch instead of redownloading
+          the picture.
+
+          Note: We require downloading the pictures here before we potentially filter the account
+              list so that the identity provider cannot determine what hints were provided
+              based on which fetches occurred.
```
TallTed, line 1056: As above.

spec/index.bs
```diff
@@ -965,6 +925,52 @@ the exception thrown.
...
+          {{IdentityProviderAccount/login_hints}} does not [=list/contain=] |provider|'s
+          {{IdentityProviderRequestOptions/loginHint}}.
+  1. If |provider|'s {{IdentityProviderRequestOptions/domainHint}} is not empty:
+      1. For every |account| in |accountList|:
+          1. If {{IdentityProviderRequestOptions/domainHint}} is "any":
+              1. If |account|'s {{IdentityProviderAccount/domain_hints}} is empty, remove
+                  |account| from |accountList|.
+          1. Otherwise, remove |account| from |accountList| if |account|'s
+              {{IdentityProviderAccount/domain_hints}} does not [=list/contain=] |provider|'s
+              {{IdentityProviderRequestOptions/domainHint}}.
```
```suggestion
          1. Otherwise, if |account|'s {{IdentityProviderAccount/domain_hints}}
             does not [=list/contain=] |provider|'s
             {{IdentityProviderRequestOptions/domainHint}},
             remove |account| from |accountList|.
```

spec/index.bs
```diff
@@ -965,6 +925,52 @@ the exception thrown.
...
+  1. If |config|.{{IdentityProviderAPIConfig/account_label}} is present:
+      1. For every |account| in |accountList|, remove |account| from |accountList| if |account|'s
+          {{IdentityProviderAccount/label_hints}} does not [=list/contain=]
+          |config|.{{IdentityProviderAPIConfig/account_label}}.
```
```suggestion
  1. If |config|.{{IdentityProviderAPIConfig/account_label}} is present:
      1. For every |account| in |accountList|, if |account|'s
          {{IdentityProviderAccount/label_hints}} does not [=list/contain=]
          |config|.{{IdentityProviderAPIConfig/account_label}},
          remove |account| from |accountList| .
```

spec/index.bs
```diff
@@ -965,6 +925,52 @@ the exception thrown.
...
+          the user at any point, it MUST reuse the result of this fetch instead of redownloading
+          the picture.
+
+          Note: We require downloading the pictures here before we potentially filter the account
+              list so that the identity provider cannot determine what hints were provided
+              based on which fetches occurred.
+  1. If |provider|'s {{IdentityProviderRequestOptions/loginHint}} is not empty:
+      1. For every |account| in |accountList|, remove |account| from |accountList| if |account|'s
+          {{IdentityProviderAccount/login_hints}} does not [=list/contain=] |provider|'s
+          {{IdentityProviderRequestOptions/loginHint}}.
```
```suggestion
      1. For every |account| in |accountList|, if |account|'s
          {{IdentityProviderAccount/login_hints}} does not [=list/contain=]
          |provider|'s {{IdentityProviderRequestOptions/loginHint}},
          remove |account| from |accountList| .
```

TallTed (changes requested):
Small changes, most to maintain "if condition, take action" form instead of inverted "take action, if condition" form, which often leads to mis-steps.

spec/index.bs
npm1: ? Looks fine to me. Also this is the same as all our other notes.

spec/index.bs
npm1: ditto

spec/index.bs
TallTed: Ah, never mind. Looks like ReSpec applies some sneaky formatting. I should have looked at the pr-preview before making this comment.

PR #752, opened 2025-06-24
Document the webdriver command to use another account.

cbiesinger:


spec/index.bs
```diff
@@ -2662,14 +2662,26 @@ The [=remote end steps=] are:
...
+    1. Act as if the user had clicked the "continue" button in the [=confirm IDP
+        login dialog=] and initiate the login flow.
+    1. Return [=success=] with data `null`.
+
+1. If the dialog is [=UI to allow the user to select an account=] and it
+    has an affordance to use another account:
+    1. Act as if the user had triggered the "Use another account" affordance
+        and initiate the [=IDP=] login flow as described above.
+
+        Note: Because multiple IDPs are only supported in passive mode, and
```
npm1, line 2678: Hmm should this command have an IDP for logging in though? Otherwise how do you know which IDP to log in to for the dialog in 2668?

cbiesinger: As the note says, there can only be one IDP

If we support the combination of use other account and multiple IDPs at some point, we can add an index parameter similar to what https://github.com/w3c-fedid/FedCM/pull/610/files was trying to do.

cbiesinger: And the case on line 2668 is pre-existing, just reformatted basically

Issue #753, opened 2025-06-27
IDP section is non-normative and out of date


mooreds:
[Per this comment](https://github.com/w3c-fedid/idp-registration/issues/9#issuecomment-3008250712):

> The IdP section of the spec is severely out of date, and the MDN docs are even worse. Your best bet for understanding the IdP interface is to read the browser section and look at what requests are actually sent.

There's more discussion on that issue, but it is probably better to have as a specific issue.

As an employee of an IDP, it is problematic to have a spec where figuring out details for our portion of the standard is accomplished by reverse engineering the browser section of the spec.

It's unclear why the IDP API is non-normative when the IDP is required to participate in a FedCM flow.

Additionally, if the MDN docs are not authoritative, they should be marked so.

So this is really two issues:

* the IDP section of the spec is non-normative but seems to be required for true implementation
* the IDP section is out of date

Issue #697, opened 2025-02-09
login_url not documented in Section 3.2


togamid:
The parameter "login_url" which is a required part of the response of the config endpoint according to [section 2.2.5](https://w3c-fedid.github.io/FedCM/#fetch-config-file) is not mentioned in [section 3.2](https://w3c-fedid.github.io/FedCM/#idp-api-config-file ), making the example in section 3.2 invalid.

The "login_url" should probably be added in section 3.2 unless there's a good reason to exclude it there.

cbiesinger:
Yes, that was an oversight, thanks for letting us know!

samuelgoto:
> Yes, that was an oversight, thanks for letting us know!

+1!

@togamid any chance you'd be interested in sending a spec PR to fix this? We can get to this at some point, but if you are interested in trying, feel free to send a spec PR!

togamid:
I can make the changes and create a PR in a few days

cbiesinger:
hi @togamid , have you had a chance to work on that PR yet?

PR #757, opened 2025-07-06
Synchronize IdP section with rest of spec

togamid:
The IdP section of the specification is currently out of date. This PR adds functionalities to the section that have been only added to other parts.

Adresses issues https://github.com/w3c-fedid/FedCM/issues/753 and https://github.com/w3c-fedid/FedCM/issues/697

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -2056,9 +2056,15 @@ The file is parsed expecting a {{IdentityProviderWellKnown}} JSON object.
 The {{IdentityProviderWellKnown}} JSON object has the following semantics:
 
 <dl dfn-type="dict-member" dfn-for="IdentityProviderWellKnown">
-    :   <dfn>provider_urls</dfn> (required)
-    ::  A list of URLs that points to valid [[#idp-api-config-file]]s.
+    :   <dfn>provider_urls</dfn> 
+    ::  A list containing exactly one URL pointing to a valid [[#idp-api-config-file]].
+    :   <dfn>accounts_endpoint</dfn>
+    ::  A URL that points to the same location as the {{IdentityProviderAPIConfig/accounts_endpoint}} in the [[#idp-api-config-file]]s.
+    :   <dfn>login_url</dfn>
+    ::  A URL that points to the same location as the {{IdentityProviderAPIConfig/login_url}} in the [[#idp-api-config-file]]s.
 </dl>
+
+Either <b>provider_urls</b> or both <b>accounts_endpoint</b> and <b> login_url </b> are required.
 
 <!-- ============================================================ -->
 ## The config file ## {#idp-api-config-file}
@@ -2091,16 +2097,14 @@ The response body must be a JSON object that can be [=converted to an IDL value|
 The {{IdentityProviderAPIConfig}} object's members have the following semantics:
 
 <dl dfn-type="dict-member" dfn-for="IdentityProviderAPIConfig">
-    :   <dfn>accounts_endpoint</dfn>
+    :   <dfn>accounts_endpoint</dfn> (required)
     ::  A URL that points to an HTTP API that complies with the [[#idp-api-accounts-endpoint]] API.
-    :   <dfn>client_metadata_endpoint</dfn>
-    ::  A URL that points to an HTTP API that complies with the [[#idp-api-client-id-metadata-endpoint]] API.
-    :   <dfn>id_assertion_endpoint</dfn>
+    :   <dfn>id_assertion_endpoint</dfn> (required)
     ::  A URL that points to an HTTP API that complies with the [[#idp-api-id-assertion-endpoint]] API.
     :   <dfn>disconnect_endpoint</dfn>
     ::  A URL that points to an HTTP API that complies with the [[#idp-api-disconnect-endpoint]]
         API.
-    :   <dfn>login_url</dfn>
+    :   <dfn>login_url</dfn> (required)
     ::  A URL that can be used to log in to this IdP. This is a user-facing URL that can be shown in
         a popup window.
     :   <dfn>branding</dfn>
@@ -2108,7 +2112,17 @@ The {{IdentityProviderAPIConfig}} object's members have the following semantics:
     :   <dfn>account_label</dfn>
     ::  An optional string that needs to match the {{IdentityProviderAccount/label_hints}} provided
         by the IdP.
+    :   <dfn>supports_use_other_account</dfn>
+    ::  A boolean determining whether the user should be shown the possibility to log into a different account even if they are alredy logged in.
 </dl>
+
+<div class="issue" heading="extension">
+    An extension may add the following:
+<dl dfn-type="argument" dfn-for="id_assertion_endpoint_request">
+    :   <dfn>client_metadata_endpoint</dfn>
+    ::  A URL that points to an HTTP API that complies with the [[#idp-api-client-id-metadata-endpoint]] API.
+</dl>
+</div>
 
 The {{IdentityProviderBranding}} enables an [=IDP=] to express their branding
 preferences, which may be used by [=user agents=] to customize the permission prompt.
@@ -2164,6 +2178,7 @@ For example:
   "client_metadata_endpoint": "/metadata",
   "id_assertion_endpoint": "/assertion",
   "disconnect_endpoint": "/disconnect",
+  "login_url": "/login",
   "branding": {
     "background_color": "green",
     "color": "#FFEEAA",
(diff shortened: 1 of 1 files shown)


togamid:
As this is my first contribution to a W3C standard I don't have a linked W3C account. I think it should be possible to mark the changes as non-substantive as the IdP section is considered non-normative.

spec/index.bs
```diff
@@ -2056,10 +2056,16 @@ The file is parsed expecting a {{IdentityProviderWellKnown}} JSON object.
...
-    ::  A list of URLs that points to valid [[#idp-api-config-file]]s.
+    :   <dfn>provider_urls</dfn> 
+    ::  A list containing exactly one URL pointing to a valid [[#idp-api-config-file]].
+    :   <dfn>accounts_endpoint</dfn>
+    ::  A URL that points to the same location as the {{IdentityProviderAPIConfig/accounts_endpoint}} in the [[#idp-api-config-file]]s.
+    :   <dfn>login_url</dfn>
+    ::  A URL that points to the same location as the {{IdentityProviderAPIConfig/login_url}} in the [[#idp-api-config-file]]s.
 </dl>
 
+Either <b>provider_urls</b> or both <b>accounts_endpoint</b> and <b> login_url </b> are required.
```
```suggestion
Either <b>provider_urls</b> or both <b>accounts_endpoint</b> and <b>login_url</b> are required.
```

spec/index.bs
```diff
@@ -2056,10 +2056,16 @@ The file is parsed expecting a {{IdentityProviderWellKnown}} JSON object.
 The {{IdentityProviderWellKnown}} JSON object has the following semantics:
 
 <dl dfn-type="dict-member" dfn-for="IdentityProviderWellKnown">
-    :   <dfn>provider_urls</dfn> (required)
-    ::  A list of URLs that points to valid [[#idp-api-config-file]]s.
+    :   <dfn>provider_urls</dfn> 
+    ::  A list containing exactly one URL pointing to a valid [[#idp-api-config-file]].
```
npm1:
```suggestion
    ::  A list containing exactly one URL pointing to [[#idp-api-config-file]].
```
Seems to read better given the title

spec/index.bs
```diff
@@ -2056,10 +2056,16 @@ The file is parsed expecting a {{IdentityProviderWellKnown}} JSON object.
...
+    :   <dfn>accounts_endpoint</dfn>
+    ::  A URL that points to the same location as the {{IdentityProviderAPIConfig/accounts_endpoint}} in the [[#idp-api-config-file]]s.
```
npm1:
```suggestion
    ::  A URL that points to the same location as the {{IdentityProviderAPIConfig/accounts_endpoint}} in [[#idp-api-config-file]].
```

spec/index.bs
```diff
@@ -2056,10 +2056,16 @@ The file is parsed expecting a {{IdentityProviderWellKnown}} JSON object.
...
+    :   <dfn>login_url</dfn>
+    ::  A URL that points to the same location as the {{IdentityProviderAPIConfig/login_url}} in the [[#idp-api-config-file]]s.
```
npm1:
```suggestion
    ::  A URL that points to the same location as the {{IdentityProviderAPIConfig/login_url}} in [[#idp-api-config-file]].
```

spec/index.bs
```diff
@@ -2091,24 +2097,32 @@ The response body must be a JSON object that can be [=converted to an IDL value|
 The {{IdentityProviderAPIConfig}} object's members have the following semantics:
 
 <dl dfn-type="dict-member" dfn-for="IdentityProviderAPIConfig">
-    :   <dfn>accounts_endpoint</dfn>
+    :   <dfn>accounts_endpoint</dfn> (required)
```
npm1, line 2100: Is this needed? Whether required or not is in the idl dfn.

spec/index.bs
```diff
@@ -2091,24 +2097,32 @@ The response body must be a JSON object that can be [=converted to an IDL value|
...
-    :   <dfn>login_url</dfn>
+    :   <dfn>login_url</dfn> (required)
     ::  A URL that can be used to log in to this IdP. This is a user-facing URL that can be shown in
         a popup window.
     :   <dfn>branding</dfn>
     ::  A set of {{IdentityProviderBranding}} options.
     :   <dfn>account_label</dfn>
     ::  An optional string that needs to match the {{IdentityProviderAccount/label_hints}} provided
         by the IdP.
+    :   <dfn>supports_use_other_account</dfn>
```
npm1: nit: place in same order as the idl

spec/index.bs
```diff
@@ -2091,24 +2097,32 @@ The response body must be a JSON object that can be [=converted to an IDL value|
...
+    ::  A boolean determining whether the user should be shown the possibility to log into a different account even if they are alredy logged in.
```
npm1:
```suggestion
    ::  A boolean determining whether the user should be shown the possibility to log into a different account even if they are already logged in. Only used in active mode calls.
```

spec/index.bs
```diff
@@ -2350,18 +2365,22 @@ It will also contain the following parameters in the request body `application/x
...
-    ::  The request nonce
+    ::  The request nonce.
     :   <dfn>account_id</dfn>
     ::  The account identifier that was selected.
-    :   <dfn>fields</dfn>
-    ::  The list of fields that the [=RP=] has requested in {{IdentityProviderRequestOptions/fields}}.
+    :   <dfn>is_auto_selected</dfn>
+    ::  A boolean indicating whether the account was automatically selected as opposed to being actively chosen by the user.
+    :   <dfn>params</dfn>
+    ::  A serialized JSON object that can have an arbitrary structure.
```
npm1: Maybe add where it is used

spec/index.bs
togamid: I wanted to make it clear that there can be multiple config endpoints and all of them have to match the accounts_endpoint in the well-known file. That's an important distinction to specifying the provider_url and a major reason why one would choose to use the accounts_endpoint + login_url form over the provider_urls form.

spec/index.bs
togamid: See comment regarding accounts_endpoint

spec/index.bs
togamid: I chose to include it here because the config endpoint has a large number of initially confusing optional properties. When developing an IdP, it is convenient to quickly see which of these are strictly necessary and which functionalities can be implemented later. 

But I only got the idea to include the "required" because of the well-known endpoint documentation, so I would be fine with removing it.

spec/index.bs
npm1: Okay, perhaps still remove 'the' since the title already has 'the'.

spec/index.bs
npm1: ditto

spec/index.bs
togamid: I thought the point of the params value was that it doesn't have a clearly defined use and instead can be used as an extension point. There are many cases where it is useful, for example to supplement the non-core fields property or implement signed requests, but the spec leaves that to agreements between RPs and IdPs.

spec/index.bs
```diff
@@ -2057,11 +2057,11 @@ The {{IdentityProviderWellKnown}} JSON object has the following semantics:
 
 <dl dfn-type="dict-member" dfn-for="IdentityProviderWellKnown">
     :   <dfn>provider_urls</dfn> 
-    ::  A list containing exactly one URL pointing to a valid [[#idp-api-config-file]].
+    ::  A list containing exactly one URL pointing to [[#idp-api-config-file]].
     :   <dfn>accounts_endpoint</dfn>
-    ::  A URL that points to the same location as the {{IdentityProviderAPIConfig/accounts_endpoint}} in the [[#idp-api-config-file]]s.
+    ::  A URL that points to the same location as {{IdentityProviderAPIConfig/accounts_endpoint}} in the [[#idp-api-config-file]]s.
```
npm1: Sorry, the `the` I suggested removing was the last one, not the one you removed.
```suggestion
    ::  A URL that points to the same location as the {{IdentityProviderAPIConfig/accounts_endpoint}} in [[#idp-api-config-file]]s.
```

spec/index.bs
```diff
@@ -2057,11 +2057,11 @@ The {{IdentityProviderWellKnown}} JSON object has the following semantics:
...
     :   <dfn>login_url</dfn>
-    ::  A URL that points to the same location as the {{IdentityProviderAPIConfig/login_url}} in the [[#idp-api-config-file]]s.
+    ::  A URL that points to the same location as {{IdentityProviderAPIConfig/login_url}} in the [[#idp-api-config-file]]s.
```
npm1: ditto

spec/index.bs
npm1: Alternatively we could modify the title of `#idp-api-config-file` but this seems easier

npm1 (approved):

spec/index.bs
npm1: Ack we can keep

mooreds:
Can we close https://github.com/w3c-fedid/FedCM/issues/753 because this got merged? In other words, is the IDP portion of the spec up to date?

PR #761, opened 2025-07-15
 FedCM Object Token Enhancement Specification

pottis:
This new specification extends the FedCM API to allow IdPs to return structured objects directly as tokens, improving developer experience while maintaining backward compatibility.

### DIFF
--- a/explorations/fedcm_object_token_spec.md
+++ b/explorations/fedcm_object_token_spec.md
@@ -1,0 +1,160 @@
+ FedCM Object Token Enhancement Specification
+
+## 1. Introduction
+
+The Federated Credential Management (FedCM) API currently requires Identity Providers (IdPs) to return tokens as USVString values through the identity assertion endpoint. This requires serializing complex data structures to JSON strings, which Relying Parties (RPs) must then parse before use. This new specification extends the FedCM API to allow IdPs to return structured objects directly as tokens, improving developer experience while maintaining backward compatibility.
+
+## 2. Background
+
+Currently, IdPs must return tokens in this format:
+
+{ "token": "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9..." }
+
+This requires IdPs to serialize complex objects into strings and RPs to parse these strings back into objects. 
+With this new enhancement, IdPs can return objects directly:
+
+{ "object_token": { "sub": "1234567890", "name": "John Doe", "additional_claims": { "permissions": ["read", "write"] } } }
+
+
+## 3. API Changes
+
+### 3.1. IDL Changes
+
+The `IdentityCredential` interface is extended with a new `objectToken` attribute:
+
+[ Exposed=Window, SecureContext, RuntimeEnabled=FedCm ]
+interface IdentityCredential : Credential { 
+// Existing attribute for backwards compatibility 
+readonly attribute USVString token;
+
+// New attribute for structured token data
+[RuntimeEnabled=FedCmObjectToken] readonly attribute object? objectToken;
+
+// Existing attributes
+......
+};
+
+
+The `IdentityProviderToken` dictionary is extended to support object tokens:
+
+dictionary IdentityProviderToken { 
+required USVString token; 
+any object_token; 
+};
+
+
+### 3.2. Network Protocol Changes
+
+The identity assertion endpoint response format is extended to support an `object_token` field:
+
+#### Existing format (maintained for backwards compatibility):
+{ "token": "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9..." }
+
+
+#### New format (allows direct object return):
+{ "object_token": { "sub": "1234567890", "name": "John Doe", "email": "john.doe@example.com", "additional_data": { "permissions": ["read", "write"], "groups": ["users", "premium"] } } }
+
+
+## 4. Behavioral Requirements
+
+### 4.1. Browser Behavior
+
+1. When processing the identity assertion endpoint response:
+   - If `token` is present in the response, set `IdentityCredential.token` to its value
+   - If `object_token` is present in the response, set `IdentityCredential.objectToken` to the parsed object
+   - If both are present, both attributes should be set accordingly
+   - If neither is present but no error is reported, treat as an invalid response
+
+2. For backwards compatibility:
+   - When `object_token` is present but `token` is not, the browser MUST generate a string representation of the object and set it as `IdentityCredential.token`
+   - This string representation MUST be valid JSON of the object
+
+3. The browser MUST NOT expose the `objectToken` attribute when the `FedCmObjectToken` feature flag is disabled
+
+### 4.2. Identity Provider Requirements
+
+1. IdPs MAY return either:
+   - A `token` property containing a string value
+   - An `object_token` property containing a JSON object
+   - Both properties (where `token` might be a string representation of `object_token` or a different value)
+
+2. If providing an `object_token`:
+   - The value MUST be a valid JSON object
+   - The object SHOULD follow standard token structures (e.g., JWT claims where applicable)
+   - The object MAY contain nested objects and arrays
+
+3. Content Type Requirements:
+   - The response MUST use a JSON-compatible Content-Type (e.g., `application/json`)
+
+### 4.3. Relying Party Behavior
+
+1. RPs SHOULD check for the existence of `objectToken` before falling back to `token`
+2. RPs SHOULD implement appropriate feature detection:
+
+const supportsObjectTokens = 'objectToken' in IdentityCredential.prototype;
+
+## 5. Implementation Guidelines
+
+### 5.1. Feature Detection
+
+if ('objectToken' in IdentityCredential.prototype) { 
+// Object tokens are supported 
+} else { 
+// Fall back to string tokens 
+}
+
+
+### 5.2. Handling Both Token Types
+
+const credential = await navigator.credentials.get({ identity: { providers: [{ configURL: 'https://idp.example/fedcm.json' }] } });
+let userData; 
+if (credential.objectToken) { 
+  // Use structured data directly userData = credential.objectToken; 
+}
+else
+{ 
+ // Try parsing the string token (if it's JSON) 
+ try 
+ { 
+    userData = JSON.parse(credential.token); 
+ }
+ catch
+ { 
+   // Handle as opaque token 
+   userData = { token: credential.token }; 
+ } 
+}
+
+
+## 6. Examples
+
+### 6.1. Identity Provider Implementation
+
+// IdP token endpoint handler 
+app.post('/token', (req, res) => { // Authenticate request and generate token data const userData = { sub: "user123", name: "Jane Doe", email: "jane@example.com", groups: ["users"], exp: Math.floor(Date.now() / 1000) + 3600 };
+// Return structured object directly 
+res.json({ object_token: userData, // Optional: also include string version for backwards compatibility token: JSON.stringify(userData) }); });
+
+
+### 6.2. Relying Party Implementation
+
+// Request credentials 
+const credential = await navigator.credentials.get({ identity: { providers: [{ configURL: 'https://idp.example/fedcm.json', clientId: 'client123' }] } });
+// Check if object tokens are supported 
+if (credential.objectToken) { 
+ console.log(Hello ${credential.objectToken.name}!); 
+ // Access nested properties directly 
+ if (credential.objectToken.groups?.includes('admin')) { 
+   showAdminInterface(); }
+ } else { 
+ // Legacy handling with string token const parsedToken = parseJwt(credential.token); console.log(Hello ${parsedToken.name}!); 
+}
+
+
+## 8. Migration Path
+
+For a smooth transition period, both IdPs and RPs should support both token formats:
+
+- IdPs should consider providing both `token` and `object_token` properties during the transition
+- RPs should implement graceful degradation when `objectToken` is not available
+- RPs using older browsers should continue to work with IdPs that only provide `object_token` due to the automatic string representation fallback


explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+
+## 1. Introduction
+
+The Federated Credential Management (FedCM) API currently requires Identity Providers (IdPs) to return tokens as USVString values through the identity assertion endpoint. This requires serializing complex data structures to JSON strings, which Relying Parties (RPs) must then parse before use. This new specification extends the FedCM API to allow IdPs to return structured objects directly as tokens, improving developer experience while maintaining backward compatibility.
+
+## 2. Background
+
+Currently, IdPs must return tokens in this format:
+
+{ "token": "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9..." }
```
TallTed:
```suggestion
`{ "token": "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9..." }`
```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+
+This requires IdPs to serialize complex objects into strings and RPs to parse these strings back into objects. 
+With this new enhancement, IdPs can return objects directly:
+
+{ "object_token": { "sub": "1234567890", "name": "John Doe", "additional_claims": { "permissions": ["read", "write"] } } }
```
TallTed:
```suggestion
`{ "object_token": { "sub": "1234567890", "name": "John Doe", "additional_claims": { "permissions": ["read", "write"] } } }`
```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+// Existing attribute for backwards compatibility 
+readonly attribute USVString token;
+
+// New attribute for structured token data
+[RuntimeEnabled=FedCmObjectToken] readonly attribute object? objectToken;
+
+// Existing attributes
+......
+};
+
```
TallTed:
~~~suggestion
```
[ Exposed=Window, SecureContext, RuntimeEnabled=FedCm ]
interface IdentityCredential : Credential { 
// Existing attribute for backwards compatibility 
readonly attribute USVString token;

// New attribute for structured token data
[RuntimeEnabled=FedCmObjectToken] readonly attribute object? objectToken;

// Existing attributes
......
};
```
~~~

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+
+The `IdentityProviderToken` dictionary is extended to support object tokens:
+
+dictionary IdentityProviderToken { 
+required USVString token; 
+any object_token; 
+};
+
```
TallTed:
~~~suggestion
```
dictionary IdentityProviderToken { 
required USVString token; 
any object_token; 
};
```
~~~

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+
+### 3.2. Network Protocol Changes
+
+The identity assertion endpoint response format is extended to support an `object_token` field:
+
+#### Existing format (maintained for backwards compatibility):
+{ "token": "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9..." }
```
TallTed:
```suggestion
`{ "token": "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9..." }`
```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+
+
+#### New format (allows direct object return):
+{ "object_token": { "sub": "1234567890", "name": "John Doe", "email": "john.doe@example.com", "additional_data": { "permissions": ["read", "write"], "groups": ["users", "premium"] } } }
```
TallTed:
```suggestion
`{ "object_token": { "sub": "1234567890", "name": "John Doe", "email": "john.doe@example.com", "additional_data": { "permissions": ["read", "write"], "groups": ["users", "premium"] } } }`
```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+   - This string representation MUST be valid JSON of the object
+
+3. The browser MUST NOT expose the `objectToken` attribute when the `FedCmObjectToken` feature flag is disabled
+
+### 4.2. Identity Provider Requirements
+
+1. IdPs MAY return either:
+   - A `token` property containing a string value
+   - An `object_token` property containing a JSON object
+   - Both properties (where `token` might be a string representation of `object_token` or a different value)
```
TallTed:
```suggestion
   - Both properties (where `token` might, and might not, be a string representation of `object_token`)
```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+## 5. Implementation Guidelines
+
+### 5.1. Feature Detection
+
+if ('objectToken' in IdentityCredential.prototype) { 
+// Object tokens are supported 
+} else { 
+// Fall back to string tokens 
+}
+
```
TallTed:
~~~suggestion
```
if ('objectToken' in IdentityCredential.prototype) { 
// Object tokens are supported 
} else { 
// Fall back to string tokens 
}
```
~~~

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+ { 
+    userData = JSON.parse(credential.token); 
+ }
+ catch
+ { 
+   // Handle as opaque token 
+   userData = { token: credential.token }; 
+ } 
+}
+
```
TallTed:
~~~suggestion
```
const credential = await navigator.credentials.get({ identity: { providers: [{ configURL: 'https://idp.example/fedcm.json' }] } });
let userData; 
if (credential.objectToken) { 
  // Use structured data directly userData = credential.objectToken; 
}
else
{ 
 // Try parsing the string token (if it's JSON) 
 try 
 { 
    userData = JSON.parse(credential.token); 
 }
 catch
 { 
   // Handle as opaque token 
   userData = { token: credential.token }; 
 } 
}
```
~~~

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+## 6. Examples
+
+### 6.1. Identity Provider Implementation
+
+// IdP token endpoint handler 
+app.post('/token', (req, res) => { // Authenticate request and generate token data const userData = { sub: "user123", name: "Jane Doe", email: "jane@example.com", groups: ["users"], exp: Math.floor(Date.now() / 1000) + 3600 };
+// Return structured object directly 
+res.json({ object_token: userData, // Optional: also include string version for backwards compatibility token: JSON.stringify(userData) }); });
+
```
TallTed:
~~~suggestion
```
// IdP token endpoint handler 
app.post('/token', (req, res) => { // Authenticate request and generate token data const userData = { sub: "user123", name: "Jane Doe", email: "jane@example.com", groups: ["users"], exp: Math.floor(Date.now() / 1000) + 3600 };
// Return structured object directly 
res.json({ object_token: userData, // Optional: also include string version for backwards compatibility token: JSON.stringify(userData) }); });
```
~~~

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+// Check if object tokens are supported 
+if (credential.objectToken) { 
+ console.log(Hello ${credential.objectToken.name}!); 
+ // Access nested properties directly 
+ if (credential.objectToken.groups?.includes('admin')) { 
+   showAdminInterface(); }
+ } else { 
+ // Legacy handling with string token const parsedToken = parseJwt(credential.token); console.log(Hello ${parsedToken.name}!); 
+}
+
```
TallTed:
~~~suggestion
```
// Request credentials 
const credential = await navigator.credentials.get({ identity: { providers: [{ configURL: 'https://idp.example/fedcm.json', clientId: 'client123' }] } });
// Check if object tokens are supported 
if (credential.objectToken) { 
 console.log(Hello ${credential.objectToken.name}!); 
 // Access nested properties directly 
 if (credential.objectToken.groups?.includes('admin')) { 
   showAdminInterface(); }
 } else { 
 // Legacy handling with string token const parsedToken = parseJwt(credential.token); console.log(Hello ${parsedToken.name}!); 
}
```
~~~

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+
+## 7. Migration Path
+
+For a smooth transition period, both IdPs and RPs should support both token formats:
+
+- IdPs should consider providing both `token` and `object_token` properties during the transition
+- RPs should implement graceful degradation when `objectToken` is not available
+- RPs using older browsers should continue to work with IdPs that only provide `object_token` due to the automatic string representation fallback
```
TallTed:
```suggestion
- RPs using older browsers should continue to work with IdPs that only provide `object_token`, due to the automatic string representation fallback
```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
+ FedCM Object Token Enhancement Specification
```
npm1: This is a proposal, not a specification. Here and elsewhere

explorations/fedcm_object_token_spec.md
npm1: Maybe ``` and multiline

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+{ "object_token": { "sub": "1234567890", "name": "John Doe", "additional_claims": { "permissions": ["read", "write"] } } }
+
+
+## 3. API Changes
+
+### 3.1. IDL Changes
+
+The `IdentityCredential` interface is extended with a new `objectToken` attribute:
+
+[ Exposed=Window, SecureContext, RuntimeEnabled=FedCm ]
```
npm1: ditto ```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+interface IdentityCredential : Credential {
```
npm1: partial interface

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+
+[ Exposed=Window, SecureContext, RuntimeEnabled=FedCm ]
+interface IdentityCredential : Credential { 
+// Existing attribute for backwards compatibility 
+readonly attribute USVString token;
+
+// New attribute for structured token data
+[RuntimeEnabled=FedCmObjectToken] readonly attribute object? objectToken;
+
+// Existing attributes
```
npm1: Remove this

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+......
+};
+
+
+The `IdentityProviderToken` dictionary is extended to support object tokens:
+
+dictionary IdentityProviderToken {
```
npm1, line 50: ditto ```

explorations/fedcm_object_token_spec.md
npm1: ditto ```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+
+The `IdentityCredential` interface is extended with a new `objectToken` attribute:
+
+[ Exposed=Window, SecureContext, RuntimeEnabled=FedCm ]
+interface IdentityCredential : Credential { 
+// Existing attribute for backwards compatibility 
+readonly attribute USVString token;
+
+// New attribute for structured token data
+[RuntimeEnabled=FedCmObjectToken] readonly attribute object? objectToken;
```
npm1: It is an interesting design discussion whether this should be a new field, as proposed here, or put everything in token as a union type https://webidl.spec.whatwg.org/#idl-union. Did you consider this? If so, can you add it to 'alternatives considered'?

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+### 4.1. Browser Behavior
+
+1. When processing the identity assertion endpoint response:
+   - If `token` is present in the response, set `IdentityCredential.token` to its value
+   - If `object_token` is present in the response, set `IdentityCredential.objectToken` to the parsed object
+   - If both are present, both attributes should be set accordingly
+   - If neither is present but no error is reported, treat as an invalid response
+
+2. For backwards compatibility:
+   - When `object_token` is present but `token` is not, the browser MUST generate a string representation of the object and set it as `IdentityCredential.token`
```
npm1, line 88: So both `token` and `object_token` would be set to the same? Is this needed, or should the caller be expected to handle `object_token` if the IDP is returning that?

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+   - This string representation MUST be valid JSON of the object
+
+3. The browser MUST NOT expose the `objectToken` attribute when the `FedCmObjectToken` feature flag is disabled
```
npm1: Remove this?

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+   - A `token` property containing a string value
+   - An `object_token` property containing a JSON object
+   - Both properties (where `token` might be a string representation of `object_token` or a different value)
+
+2. If providing an `object_token`:
+   - The value MUST be a valid JSON object
+   - The object SHOULD follow standard token structures (e.g., JWT claims where applicable)
+   - The object MAY contain nested objects and arrays
+
+3. Content Type Requirements:
```
npm1: This is unchanged? Why are you adding this here

explorations/fedcm_object_token_spec.md
npm1: Why should it be allowed for IDP to return both? Backwards compatibility or?

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+   - The response MUST use a JSON-compatible Content-Type (e.g., `application/json`)
+
+### 4.3. Relying Party Behavior
+
+1. RPs SHOULD check for the existence of `objectToken` before falling back to `token`
+2. RPs SHOULD implement appropriate feature detection:
+
+const supportsObjectTokens = 'objectToken' in IdentityCredential.prototype;
```
npm1, line 120: ditto ```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+
+## 5. Implementation Guidelines
+
+### 5.1. Feature Detection
+
+if ('objectToken' in IdentityCredential.prototype) {
```
npm1, line 136: Repetitive with what you just wrote.

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,160 @@
...
+### 5.1. Feature Detection
+
+if ('objectToken' in IdentityCredential.prototype) { 
+// Object tokens are supported 
+} else { 
+// Fall back to string tokens 
+}
+
+
+### 5.2. Handling Both Token Types
```
npm1: Hard to tell what is different from the examples section

explorations/fedcm_object_token_spec.md
pottis: I didn’t quite follow the comment. Could you please elaborate or rephrase it for clarity?

explorations/fedcm_object_token_spec.md
pottis: Yes for Backwards compatibility

explorations/fedcm_object_token_spec.md
pottis: I understand this may appear repetitive; however, I am providing this explanation in accordance with the structure of the section.

explorations/fedcm_object_token_spec.md
pottis: This section may appear repetitive; if deemed unnecessary, I am open to removing it.

explorations/fedcm_object_token_spec.md
pottis: This has been included for the sake of completeness; however, I am open to removing it if it is deemed unnecessary.

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,252 @@
+ FedCM Object Token Enhancement Proposal
```
npm1:
```suggestion
# FedCM Object Token Enhancement Proposal
```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,252 @@
...
+
+## 3. API Changes
+
+### 3.1. IDL Changes
+
+The `IdentityCredential` interface is extended with a new `objectToken` attribute:
+
+```
+[ Exposed=Window, SecureContext, RuntimeEnabled=FedCm ]
+interface IdentityCredential : Credential {
```
npm1:
```suggestion
partial interface IdentityCredential : Credential { 
```

explorations/fedcm_object_token_spec.md
npm1: Added it as a suggestion.

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,252 @@
...
+```
+
+## 3. API Changes
+
+### 3.1. IDL Changes
+
+The `IdentityCredential` interface is extended with a new `objectToken` attribute:
+
+```
+[ Exposed=Window, SecureContext, RuntimeEnabled=FedCm ]
```
npm1:
```suggestion
[ Exposed=Window, SecureContext ]
```
Remove the Chrome-specific part.

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,252 @@
...
+interface IdentityCredential : Credential { 
+// Existing attribute for backwards compatibility 
+readonly attribute USVString token;
+
+// New attribute for structured token data
+[RuntimeEnabled=FedCmObjectToken] readonly attribute object? objectToken;
```
npm1:
```suggestion
readonly attribute object? objectToken;
```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,252 @@
...
+};
+```
+
+The `IdentityProviderToken` dictionary is extended to support object tokens:
+
+```
+dictionary IdentityProviderToken { 
+required USVString token;
```
npm1:
```suggestion
USVString token; 
```
Now token becomes optional, no?

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,252 @@
...
+dictionary IdentityProviderToken { 
+required USVString token; 
+any object_token; 
+};
+```
+
+As an alternative to adding a new objectToken attribute, we could enhance the existing token attribute to use a union type that can represent either a string or an object:
+
+```
+[ Exposed=Window, SecureContext, RuntimeEnabled=FedCm ]
```
npm1:
```suggestion
[ Exposed=Window, SecureContext ]
```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,252 @@
...
+interface IdentityCredential : Credential {
+    // Use a union type instead of separate attributes
+    [RuntimeEnabled=FedCmObjectToken] readonly attribute (USVString or object) token;
```
npm1:
```suggestion
    readonly attribute (USVString or object) token;
```

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,252 @@
...
+    
+    // Existing attributes remain unchanged
```
npm1: Use partial interface above and you can remove the rest of this idl stuff...

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,252 @@
...
+    [CallWith=ScriptState, RaisesException, MeasureAs=FedCmDisconnect] 
+    static Promise<undefined> disconnect(IdentityCredentialDisconnectOptions options);
+};
+```
+
+The `IdentityProviderToken` dictionary changes:
+
+```
+dictionary IdentityProviderToken {
+  required (USVString or object) token;
```
npm1: I think this is better. Why is this not the chosen option?

togamid: Allowing the token to be either a JSON Object or a string feels risky from a security standpoint. Expecting something to be a string that is actually a more complex datatype is a classic vulnerability, and while I can't think of a specific attack right now, JavaScript is convoluted enough that there is probably some use-case where it is a security problem.

explorations/fedcm_object_token_spec.md
```diff
@@ -0,0 +1,252 @@
+ FedCM Object Token Enhancement Proposal
+
+## 1. Introduction
+
+The Federated Credential Management (FedCM) API currently requires Identity Providers (IdPs) to return tokens as USVString values through the identity assertion endpoint. This requires serializing complex data structures to JSON strings, which Relying Parties (RPs) must then parse before use. This proposal extends the FedCM API to allow IdPs to return structured objects directly as tokens, improving developer experience while maintaining backward compatibility.
```
togamid, line 5: 1. The response is not necessarily a JSON object. IdPs might implement "SAML over FedCM" and return XML responses, or encrypt their tokens and return them base64 encoded.
2. If it is a JSON object, IdPs still need to serialize the response to JSON strings for the transfer to the browser. This effectively only removes two quotes and <10 lines of code

togamid (commented):
Some comments on this proposal:
1. The proposal assumes that most or all responses are JSON. That is not necessarily the case, as IdPs might encrypt their tokens or respond with other formats like XML. Imo these are valid use-cases that should be protected and benefit from a string response.
2. The proposal further increases the complexity of FedCM. FedCM is already a complex spec, and adding what is effectively a third kind of response to the identity_assertion_endpoint only increases this. It also increases the complexityy of browser implementations that need to deal with these responses. Imo this is not justified by the effort needed to implement the JSOn functionality on the SP and IdP sides. If both implement it in JavaScript, the effort replaced by the proposal amounts to two function calls. This even demonstrated in the "Example" section, where the response is also sent in the token property with minimal differences compared to sending it through the proposed new property.In the example, the decision about which property to use takes up more space than the serilization/parsing the string for the case without this proposal.
3. The token is a response that comes from a third party that is not necessarily trusted by the SP and might even be unknown to it with the IdP registration API. Returning it as a string safely quarantines it and leaves the SP to deal with it with the appropriate amount of caution. When returned as an object, the potential for wrong assumptions, bugs and vulnerabilities is higher as a malicious IdP has access to more obscure JavaScript behaviour.
4. For anything beyond simple customization of a website (e. g. adding a name), the SP probably needs to send the token to a serverside endpoint for validation anyway. This requires its serialization. Keeping it as a string instead of parsing it to JSON and back to a string again is both safer and less bug prone. The SP website can then parse the token once it has been verified and validated by its own backend.

In my opinion, the proposal adds unnecessary complexity to FedCM in an area where it isn't needed. It breaks with FedCMs agnostic approach to the response token while still leaving its structure undefined, imposing restrictions while not providing significant functionality. Instead of restricting the response token, in my opinion more effort should be invested into actually defining spec variations/extensions that define what the token should contain. Attempts like the [OAuth FedCM profile](https://github.com/aaronpk/oauth-fedcm-profile) exist, but are not defined enough to be a real specification and are quickly out of date due to the speed of change of FedCM.

hlflanagan:
Discussed on the [July 29](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-07-29-FedCM-notes.md) call.

Issue #582, opened 2024-05-13
FedCM spec: IdP should validate nonce to prevent CSRF?


obfuscoder:
In the current draft of the FedCM spec (https://fedidcg.github.io/FedCM/), section 3.5. Identity assertion endpoint contains the following:

`
NOTE: An [IDP](https://fedidcg.github.io/FedCM/#idp) should validate the nonce, if present, to prevent CSRF-style attacks.
`

There is some confusion amongst implementers, what should be validated here?

The nonce is optional and it is the RP's responsibility to define unique nonce values when initiating the FedCM call and compare the returned nonce with the provided one.

samuelgoto:
I'm thinking that this should be done as part of #536 in userland (e.g. an OAuth Profile layer), rather than in browserland in the FedCM spec per se. 

@timcappalli, wdyt?

achimschloss:
Agreed, if nonce was not already a top level parameter and removing it would be a breaking change it should be userland via https://github.com/fedidcg/FedCM/issues/556 - I guess it would be fair to just remove the note

aaronpk:
the comment in the sample code in the spec says this:

>    // Assume there is a method returning a random number. Store the value in a variable which can
>    // later be used to check against the value in the token returned.

This assumes the client understands how to parse the token returned, which I thought was not an assumption the FedCM spec wanted to make. I would recommend removing the `nonce` parameter entirely in favor of moving all of that kind of thing into a different layer as described in #536.

If there is anything FedCM needs to do at the FedCM API layer to mitigate this kind of injection/replay attacks, it should provide the other half of the validation story as well.

bc-pi:
@aaronpk's perspective on layering https://github.com/fedidcg/FedCM/issues/582#issuecomment-2108243833 makes a lot of sense.

cbiesinger:
In principle I agree with moving nonce to `params`. However, we have shipped it since the initial version of FedCM and so I am not sure we can remove it now for backwards compatibility reasons.

obfuscoder:
As changing API may be a bit tricky and needs more thinking through, could you at least remove the NOTE I mentioned as @achimschloss suggested?

samuelgoto:
> As changing API may be a bit tricky and needs more thinking through, could you at least remove the NOTE I mentioned as @achimschloss suggested?

Yeah, that sounds reasonable. Care to send a PR?

We'll get to this at some point, but if you could put together a PR, you may beat us to it.

[...]

PR #762, opened 2025-07-15
FedCM nonce Parameter Deprecation Specification

pottis:
The Federated Credential Management (FedCM) API currently allows Identity Providers (IdPs) to specify a `nonce` parameter as a top-level field in the provider configuration object. This specification proposes deprecating this top-level parameter and moving it to the `params` object, which is the intended location for all IdP-specific parameters. This change will improve API consistency while maintaining backward compatibility during a transition period.

### DIFF
--- a/explorations/nonce_deprecate.md
+++ b/explorations/nonce_deprecate.md
@@ -105,8 +105,8 @@ The browser should modify the token request generation code to extract the `nonc
 
 Relying Parties should update their code to use the new parameter location:
 
-// Before 
 ```js
+// Before
 navigator.credentials.get({
   identity: {
     providers: [{
@@ -124,8 +124,8 @@ navigator.credentials.get({
 });
 ```
 
-// After 
 ```js
+// After
 navigator.credentials.get({
   identity: {
     providers: [{
@@ -146,8 +146,8 @@ navigator.credentials.get({
 
 ### 6.1. Relying Party Implementation with New Format
 
-// Request with nonce in params object 
 ```js
+// Request with nonce in params object 
 let credential = await navigator.credentials.get({
   identity: {
     providers: [{


explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+        "response_type": "id_token",
+        "scope": "photos:read photos:write",
+      }
+    },
+  }
+  // If possible, return without prompting the user, if not possible
+  // prompt the user.
+  mediation: "optional",
+});
+
```
TallTed:
~~~suggestion
```js
let {token} = await navigator.credentials.get({
  identity: {
    providers: [{
      clientId: "1234",
      nonce: "234234",    // Top-level parameter 
      loginHint: "previous@user.com",
      configURL: "https://idp.example/fedcm.json",
      // A string with parameters that need to be passed from the
      // RP to the IdP but that don't really play any role with
      // the browser.
      params: {
        "IDP_SPECIFIC_PARAM": "1",
        "foo": "BAR",
        "ETC": "MOAR",
        "response_type": "id_token",
        "scope": "photos:read photos:write",
      }
    },
  }
  // If possible, return without prompting the user; if not possible,
  // prompt the user.
  mediation: "optional",
});
```
~~~

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+{ 
+  clientId: "1234", 
+  nonce: "234234",  // Top-level parameter 
+  configURL: "https://idp.example/fedcm.json", 
+  params: { 
+   "response_type": "id_token", 
+   "scope": "photos:read"
+  }
+}
+
```
TallTed:
~~~suggestion
```js
{ 
  clientId: "1234", 
  nonce: "234234",  // Top-level parameter 
  configURL: "https://idp.example/fedcm.json", 
  params: { 
   "response_type": "id_token", 
   "scope": "photos:read"
  }
}
```
~~~

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+{ 
+  clientId: "1234", 
+  configURL: "https://idp.example/fedcm.json", 
+  params: { 
+   "nonce": "234234",  // // Moved into params
+   "response_type": "id_token", 
+   "scope": "photos:read"
+  }
+}
+
```
TallTed:
~~~suggestion
```js
{ 
  clientId: "1234", 
  configURL: "https://idp.example/fedcm.json", 
  params: { 
   "nonce": "234234",  // // Moved into params
   "response_type": "id_token", 
   "scope": "photos:read"
  }
}
```
~~~

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+
+## 4. Behavioral Requirements
+
+### 4.1. Browser Behavior
+
+1. During a transition period, the browser MUST:
+   - Check for `nonce` in the `params` object first
```
TallTed:
```suggestion
   - First, check for `nonce` in the `params` object
```

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+      loginHint: "previous@user.com",
+      configURL: "https://idp.example/fedcm.json",
+      params: {  
+	   nonce: "234234",   // New location 
+	   /* ... */ 
+	  } }] }
+    },
+  }
+  mediation: "optional",
+});
```
TallTed:
~~~suggestion
#### Before 
```js
navigator.credentials.get({
  identity: {
    providers: [{
      clientId: "1234",
      nonce: "234234",    // // Old location
      loginHint: "previous@user.com",
      configURL: "https://idp.example/fedcm.json",
      // A string with parameters that need to be passed from the
      // RP to the IdP but that don't really play any role with
      // the browser.
      params: { /* ... */ } }] }
    },
  }
  mediation: "optional",
});
```
#### After 
```js
navigator.credentials.get({
  identity: {
    providers: [{
      clientId: "1234",
      loginHint: "previous@user.com",
      configURL: "https://idp.example/fedcm.json",
      params: {  
	   nonce: "234234",   // New location 
	   /* ... */ 
	  } }] }
    },
  }
  mediation: "optional",
});
```
~~~

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+      loginHint: "previous@user.com",
+      configURL: "https://idp.example/fedcm.json",
+      params: {
+        nonce: generateSecureNonce(),
+       /* ... */ 
+      }
+    },
+  }
+  mediation: "optional",
+});
```
TallTed, line 164: #### Request with nonce in params object 
~~~suggestion
```js
let credential = await navigator.credentials.get({
  identity: {
    providers: [{
      clientId: "1234",
      loginHint: "previous@user.com",
      configURL: "https://idp.example/fedcm.json",
      params: {
        nonce: generateSecureNonce(),
       /* ... */ 
      }
    },
  }
  mediation: "optional",
});
```
~~~

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+
+### 6.2. Identity Provider Implementation (Unchanged)
+
+The IdP implementation remains unchanged as the browser will continue to pass the nonce parameter to the token endpoint in the same format.
```
TallTed:
```suggestion
The IdP implementation remains unchanged as the browser will continue to pass the `nonce` parameter to the token endpoint in the same format.
```

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
+# FedCM Nonce Parameter Deprecation Specification
```
npm1: Remove specification here and elsewhere. This is a proposal, not a 'specification'.

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+
+## 1. Introduction
+
+The Federated Credential Management (FedCM) API currently allows Identity Providers (IdPs) to specify a `nonce` parameter as a top-level field in the provider configuration object. This specification proposes deprecating this top-level parameter and moving it to the `params` object, which is the intended location for all IdP-specific parameters. This change will improve API consistency while maintaining backward compatibility during a transition period.
+
+## 2. Background
+
+Currently, when making a FedCM API call, the `nonce` parameter is specified as a direct property of the provider object:
+
```
npm1, line 10: Add ```js for the code part.

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+The `nonce` value is used when constructing token requests to the IdP's assertion endpoint to prevent replay attacks. However, this approach creates inconsistency in the API design, as other IdP-specific parameters are grouped in the `params` object.
+
+## 3. API Changes
+
+### 3.1. Parameter Location Changes
+
+The `nonce` parameter should be moved from the top level of the identity provider configuration to the `params` object:
+
+#### Current usage:
+{
```
npm1, line 47: ditto ```

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+  configURL: "https://idp.example/fedcm.json", 
+  params: { 
+   "response_type": "id_token", 
+   "scope": "photos:read"
+  }
+}
+
+
+#### New usage:
+{
```
npm1, line 60: ditto

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+
+### 4.1. Browser Behavior
+
+1. During a transition period, the browser MUST:
+   - Check for `nonce` in the `params` object first
+   - Fall back to the top-level `nonce` parameter if not found in `params`
+   - Show a deprecation warning in developer tools when the top-level parameter is used
+
+2. After the transition period, the browser:
+   - MAY reject requests that use the top-level `nonce` parameter
```
npm1, line 81: I don't think this is what would happen. Parameters that are not members of the dict are just ignored.

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+   - SHOULD continue to show deprecation warnings until the parameter is fully removed
+
+3. When constructing token requests to the IdP:
+   - Use the `nonce` from `params` if available
```
npm1, line 85: This doesn't make sense. Params is arbitrary, so you can always define nonce on it.

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+### 4.3. Relying Party Requirements
+
+1. Relying Parties SHOULD update their code to place the `nonce` parameter in the `params` object
+2. During the transition period, RPs can continue to use the top-level parameter, though they will receive deprecation warnings
+
+## 5. Implementation Guidelines
+
+### 5.1. Browser Implementation
+
+The browser should modify the token request generation code to extract the `nonce` from the `params` object when available:
```
npm1:
```suggestion
The browser should modify the token request generation code to extract the `nonce` from the `params` object when available.
```

npm1: But actually I think this is just wrong, the browser would not need to modify anything since params is already sent to the IDP. The difficult part is removing support to send the nonce from outside of params.

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+
+### 5.2. Relying Party Migration
+
+Relying Parties should update their code to use the new parameter location:
+
+// Before
```
npm1: ditto ```js

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+    },
+  }
+  mediation: "optional",
+});
+ 
+## 6. Examples
+
+### 6.1. Relying Party Implementation with New Format
+
+// Request with nonce in params object
```
npm1, line 150: ditto

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,195 @@
...
+    },
+  }
+  mediation: "optional",
+});
+
+### 6.2. Identity Provider Implementation (Unchanged)
+
+The IdP implementation remains unchanged as the browser will continue to pass the nonce parameter to the token endpoint in the same format.
+
+## 7. Migration Path
```
npm1, line 171: This seems repetitive with a previous section.

npm1 (commented):
Thanks for the writeup! A lot of this seems written from the Chromium perspective, such as the plan to mitigate the backwards incompatible change (since only Chromium has shipped FedCM). I suppose it doesn't harm to document the plan, but this should probably be made more concise as the proposed spec change here is very simple. In fact, I would have probably directly sent a PR to modify the spec, which this PR is not doing.

explorations/nonce_deprecate.md
TallTed: see https://github.com/w3c-fedid/FedCM/pull/762/files#r2207947429

explorations/nonce_deprecate.md
TallTed: see https://github.com/w3c-fedid/FedCM/pull/762/files#r2207953703

explorations/nonce_deprecate.md
TallTed: see https://github.com/w3c-fedid/FedCM/pull/762/files#r2207954557

explorations/nonce_deprecate.md
TallTed: see https://github.com/w3c-fedid/FedCM/pull/762/files#r2207964917

explorations/nonce_deprecate.md
TallTed: see https://github.com/w3c-fedid/FedCM/pull/762/files#r2207966028

explorations/nonce_deprecate.md
pottis: Although the content may appear repetitive, it is presented under distinct headers/sections. If you feel the repetition should be minimized, please let me know which section you'd prefer to remove.

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,204 @@
...
+  }
+}
+```
+
+## 4. Behavioral Requirements
+
+### 4.1. Browser Behavior
+
+1. During a transition period, the browser MUST:
+   - First, check for `nonce` in the `params` object
```
npm1, line 76: This should not be needed, right? params should be passed where nonce is passed. No need to 'check' for it

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,204 @@
...
+
+### 5.1. Browser Implementation
+
+The browser should modify the token request generation code to extract the `nonce` from the `params` object when available.
+
+### 5.2. Relying Party Migration
+
+Relying Parties should update their code to use the new parameter location:
+
+// Before
```
npm1:
```suggestion
Before:
```
Either write it as prose or move inside the js tag

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,204 @@
...
+      // RP to the IdP but that don't really play any role with
+      // the browser.
+      params: { /* ... */ } }] }
+    },
+  }
+  mediation: "optional",
+});
+```
+
+// After
```
npm1: ditto

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,204 @@
...
+  }
+  mediation: "optional",
+});
+```
+ 
+## 6. Examples
+
+### 6.1. Relying Party Implementation with New Format
+
+// Request with nonce in params object
```
npm1: ditto

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Phase 1: Support Both Formats with Deprecation Warning
+
+- Browser supports both parameter locations
+- Top-level usage triggers a warning in developer tools
+- Recommended format is documented in browser developer documentation
+
+### Phase 2: Elevate Warning to Error
+
+- Warning is upgraded to an error-level message
```
npm1, line 181: Generally, having a non-existent type in a dict does not lead to an error, so should this actually be an error?

togamid:
My understanding of the "params" property was that its content is deliberately arbitrary to allow protocols building on FedCM like [FedCM for OAuth](https://github.com/aaronpk/oauth-fedcm-profile) by aaronpk to define it. Moving the nonce parameter into "params" (or even parsing "params" in the browser implementation to extract the nonce) would change that. 
In my opinion, the "nonce" option doesn't hurt that much as it is now. If it causes trouble, it might be better to just remove it entirely and specify that protocols that care about replay attacks need to define a nonce as part of the "params" option without actually defining how exactly they should do it.

aaronpk:
If i remember the discussions correctly, the intent was to deprecate the `nonce` parameter entirely, as not every protocol that leverages FedCM will use a parameter called `nonce`. If you are using a protocol that needs `nonce`, then you can put a `nonce` field in `params`.

https://github.com/w3c-fedid/FedCM/issues/582#issuecomment-2108243833

explorations/nonce_deprecate.md
```diff
@@ -0,0 +1,204 @@
...
+   - Show a deprecation warning in developer tools when the top-level parameter is used
+
+2. After the transition period, the browser:
+   - MAY reject requests that use the top-level `nonce` parameter
+   - SHOULD continue to show deprecation warnings until the parameter is fully removed
+
+3. When constructing token requests to the IdP:
+   - Use the `nonce` from `params` if available
+   - Use the top-level `nonce` as fallback during transition
+   - Pass the `nonce` value to the assertion endpoint as before
```
cbiesinger, line 87: The purpose of the change is that the browser does not need to know about nonce at all. So, an RP would pass the nonce to the IDP using params, it would be sent like any other param, and the browser gets out of sending the toplevel nonce parameter that it does today.

npm1:
Closing this since the spec change is done.

PR #771, opened 2025-08-07
FedCM Object Token Enhancement

pottis:
This change extends the FedCM API to allow IdPs to return structured objects directly as tokens.

***
<a href="https://pr-preview.s3.amazonaws.com/pottis/FedCM/pull/771.html" title="Last updated on Aug 13, 2025, 4:05 AM UTC (1a2d61b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/771/b8e3ce8...pottis:1a2d61b.html" title="Last updated on Aug 13, 2025, 4:05 AM UTC (1a2d61b)">Diff</a>


spec/index.bs
```diff
@@ -1517,8 +1531,18 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
+            1. Let |tokenString| be null.
+            1. If Type(|tokenValue|) is String, set |tokenString| to |tokenValue|.
+            1. Otherwise, if Type(|tokenValue|) is Object:
+            1. Let |jsonString| be the result of [=serializing a JavaScript value to a JSON string=]
+                given |tokenValue|.
+            1. If serialization threw an exception, set |credential| to a the result of
+                [=create an IdentityCredentialError=] with {} and |globalObject|, and return.
+            1. Set |tokenString| to |jsonString|.
+            1. Otherwise, set |credential| to a the result of [=create an IdentityCredentialError=]
+                with {} and |globalObject|, and return.
```
TallTed: I am sure that the current nesting is incorrect, but I'm not sure my suggested nesting is correct. Especially, I'm not sure about the last Otherwise, and the `Set |tokenString| to |jsonString|` preceding it.
```suggestion
            1. Otherwise, if Type(|tokenValue|) is Object:
               1. Let |jsonString| be the result of [=serializing a JavaScript value to a JSON string=]
                   given |tokenValue|.
               1. If serialization threw an exception, set |credential| to a the result of
                   [=create an IdentityCredentialError=] with {} and |globalObject|, and return.
               1. Set |tokenString| to |jsonString|.
               1. Otherwise, set |credential| to a the result of [=create an IdentityCredentialError=]
                   with {} and |globalObject|, and return.
```

spec/index.bs
```diff
@@ -164,6 +164,19 @@ could be implemented.
...
         }
       });
+      
+      // Handle the token - it could be a string or serialized JSON
+      let token = credential.token;
+      try {
+        // Try to parse as JSON first
+        let tokenData = JSON.parse(token);
+        // Handle structured token (e.g., OAuth response)
+        console.log("Access token:", tokenData.access_token);
```
npm1: Maybe add as comment? I dont think we do console.logs as examples elsewhere.

spec/index.bs
```diff
@@ -465,6 +478,7 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
     :   <b>{{IdentityCredential/token}}</b>
     ::  The {{IdentityCredential/token}}'s attribute getter returns the value it is set to.
         It represents the minted {{IdentityAssertionResponse/token}} provided by the [=IDP=].
+        The token can be either a string (for traditional tokens like JWTs) or a serialized JSON object containing structured authentication data. When the [=IDP=] returns a JSON object as the token, it will be serialized to a string before being made available through this attribute. [=RP=]'s should parse the token value to determine its format and handle it accordingly.
```
npm1: Is this how any works? That an object is serialized to a string? I dont think so?

spec/index.bs
```diff
@@ -1517,8 +1531,18 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
                 could set it to "server_error", and if the [=response/status=] is 503, it could set
                 it to "temporarily_unavailable".
-        1. Otherwise, if {{IdentityAssertionResponse/token}} was specified, let |tokenString|
-            be |token|'s {{IdentityAssertionResponse/token}}.
+        1. Otherwise, if {{IdentityAssertionResponse/token}} was specified:
+            1. Let |tokenValue| be |token|'s {{IdentityAssertionResponse/token}}.
+            1. Let |tokenString| be null.
+            1. If Type(|tokenValue|) is String, set |tokenString| to |tokenValue|.
+            1. Otherwise, if Type(|tokenValue|) is Object:
+            1. Let |jsonString| be the result of [=serializing a JavaScript value to a JSON string=]
```
npm1: Isn't the point of changing the value to allow passing it through? Why are you serializing it?

spec/index.bs
```diff
@@ -2450,7 +2475,7 @@ Portability is left entirely up to [=IDPs=], who can choose
 between a variety of different mechanisms to accomplish it
 (e.g. [OIDC's Account Porting](https://openid.net/specs/openid-connect-account-porting-1_0.html)).
 
-For example:
+For example (string token):
```
```suggestion
Example (string token):
```

spec/index.bs
```diff
@@ -2460,6 +2485,16 @@ For example:
 ```
 </div>
 
+For example (JSON object token):
```
```suggestion
Example (JSON object token):
```

spec/index.bs
```diff
@@ -2924,6 +2959,26 @@ distinguishing browser-initiated FedCM flows from arbitrary requests.
...
+
+**Recommendations**
+
+1. [=RP=]'s should implement robust token parsing logic that can differentiate between traditional
+    string tokens (like JWTs) and serialized JSON objects.
+2. [=RP=]'s should validate tokens according to the format agreed upon with their [=IDP=] partners.
+3. [=IDP=]'s should clearly document the token format they return to help [=RP=]'s implement proper
+    validation.
+4. When returning JSON object tokens, [=IDP=]'s should ensure the structure is consistent and 
+    well-documented.
```
TallTed: Apostrophe-S (`'s`) typically indicates ownership, not plurality. I believe the `.bs` processor will handle plurals within the `[= =]` frames (so, e.g., `[=RP=]'s` should become `[=RPs=]`). If not, the apostrophes should still be removed (e.g., `[=RP=]'s` should become `[=RP]s`).
```suggestion
1. [=RPs=] should implement robust token parsing logic that can differentiate between traditional
    string tokens (like JWTs) and serialized JSON objects.
2. [=RPs=] should validate tokens according to the format agreed upon with their [=IDP=] partners.
3. [=IDPs=] should clearly document the token format they return to help [=RPs=] implement proper
    validation.
4. When returning JSON object tokens, [=IDPs=] should ensure the structure is consistent and 
    well-documented.
```

spec/index.bs
```diff
@@ -3473,6 +3528,19 @@ These schemes have not been adopted by this specification.
...
+### Token Content Privacy ### {#token-content-privacy}
+<!-- ============================================================ -->
+
+The FedCM API treats token content as opaque data, regardless of whether it's a string or JSON object.
+This design ensures:
+
+1. The user agent does not inspect or validate token contents, maintaining the privacy of the
+    authentication data.
+2. The token format (string vs. JSON) does not affect the privacy properties of the API.
+3. [=IDP=]'s have flexibility in token design without compromising user privacy.
```
```suggestion
3. [=IDPs=] have flexibility in token design without compromising user privacy.
```

spec/index.bs
```diff
@@ -1517,8 +1531,18 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
                 |response|'s [=response/status=]. For example, if the [=response/status=] is 500, it
                 could set it to "server_error", and if the [=response/status=] is 503, it could set
```
```suggestion
                |response|'s [=response/status=]. For example, if the [=response/status=] is `500`, it
                could set it to "server_error", and if the [=response/status=] is `503`, it could set
```

spec/index.bs
```suggestion
        The token can be either a string (for traditional tokens like JWTs) or a serialized JSON object containing structured authentication data. When the [=IDP=] returns a JSON object as the token, it will be serialized to a string before being made available through this attribute. [=RPs=] should parse the token value to determine its format and handle it accordingly.
```

spec/index.bs
```diff
@@ -2924,6 +2959,26 @@ distinguishing browser-initiated FedCM flows from arbitrary requests.
 [=IDP=] and [=RP=] endpoints are implemented correctly and do not contain vulnerabilities that could 
 allow an attacker to bypass the FedCM flow or extract sensitive data.
 
+<!-- ============================================================ -->
+### Token Format and Validation ### {#token-format-validation}
+<!-- ============================================================ -->
+
+**Consideration**
+
+With the support for both string and JSON object tokens, [=RP=]s must be prepared to handle different
```
```suggestion
With the support for both string and JSON object tokens, [=RPs=] must be prepared to handle different
```

spec/index.bs
pottis: Addressed. Thanks.

spec/index.bs
pottis: I saw in spec in using **[=IDP=]'s** so got confused as to which format to use.

spec/index.bs
pottis: Since this line remains unchanged, would you still prefer it to be included in this PR?

spec/index.bs
pottis: [same as ](https://github.com/w3c-fedid/FedCM/pull/771#discussion_r2262074983)

spec/index.bs
pottis: Addressed. Thanks.

spec/index.bs
pottis: Addressed. Thanks.

spec/index.bs
pottis: Corrected now.

spec/index.bs
pottis: Modified the steps.

spec/index.bs
pottis: Addressed. Thanks

spec/index.bs
pottis: Addressed. Thanks.

spec/index.bs
pottis: Addressed. Thanks.

mooreds:
@hlflanagan @wseltzer should we have a discussion about this proposal? Has it already happened at a meeting I missed?

spec/index.bs
```diff
@@ -465,6 +473,17 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
...
+        The token attribute can return either:
+        - A string value (e.g., traditional tokens like JWTs)
+        - A JavaScript object (e.g., structured authentication data with access_token, refresh_token, etc.)
+        
+        The type returned depends on what the [=IDP=] provides in its assertion endpoint response.
+        [=RPs=] MUST check the type of the token before processing it.
+        
+        <div class="note">
+        Breaking Change: Previous versions of this specification only supported string tokens.
+        [=RPs=] updating to support this version MUST implement logic to handle both types.
```
cbiesinger: No, that is not right. This is only the case if the IDP changes to return non-string tokens, and I would not expect many IDPs to do that, because they return JWTs today and they don't lend themselves to JSON format very well.

We also don't really add notes like that to the spec for other breaking changes we made

spec/index.bs
```diff
@@ -164,6 +164,14 @@ could be implemented.
           }]
         }
       });
+      
+      // Handle the token - it could be a string or serialized JSON
+      let token = credential.token;
+      if (typeof token === 'object' && token !== null) {
+        // Handle structured token (e.g., OAuth response)
```
cbiesinger, line 171: I am not familiar enough with OAuth, hopefully someone else can confirm that it typically uses objects

spec/index.bs
```diff
@@ -1517,8 +1536,14 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
                 |response|'s [=response/status=]. For example, if the [=response/status=] is 500, it
                 could set it to "server_error", and if the [=response/status=] is 503, it could set
                 it to "temporarily_unavailable".
-        1. Otherwise, if {{IdentityAssertionResponse/token}} was specified, let |tokenString|
-            be |token|'s {{IdentityAssertionResponse/token}}.
+        1. Otherwise, if {{IdentityAssertionResponse/token}} was specified:
+            1. Let |token| be the result of extracting the token from the response.
+            1. If |token| is a JSON object in the response:
```
cbiesinger: This should be specified more precisely.

If you wanted to keep the structure that you wrote, it should be something like `If |token| is an [=ordered map=]`.

But it would be simpler to just not limit token to maps and strings and just replace lines 1541..1545 with `Set |credential|.{{IdentityCredential/token}} to |token|`

I am not sure it's worth saying `The [=user agent=] MUST preserve the object structure.` since that follows from the spec text anyway

I'm fine with `The [=user agent=] MUST NOT modify or validate the token content.` just to be explicit, although I'm not sure this is strictly necessary either.

spec/index.bs
```diff
@@ -2425,7 +2450,15 @@ Every {{IdentityAssertionResponse}} is expected to have members with the followi
 
 <dl dfn-type="dict-member" dfn-for="IdentityAssertionResponse">
     :   <dfn>token</dfn>
-    ::  The resulting token.
+    ::  The resulting token. The value can be either:
```
cbiesinger: With my suggestion above, this value can be anything, not just string or object

spec/index.bs
```diff
@@ -2425,7 +2450,15 @@ Every {{IdentityAssertionResponse}} is expected to have members with the followi
...
+        - A string (e.g., a JWT or other opaque token format)
+        - A JSON object containing structured authentication data
+        
+        The [=user agent=] MUST NOT serialize object tokens to strings. The exact
+        structure provided by the [=IDP=] MUST be preserved and passed to the [=RP=].
+        
+        See the examples at the end of [[#idp-api-id-assertion-endpoint]] for both string
```
cbiesinger, line 2440: I don't think this paragraph adds value, since the algorithm is clear about what the user agent needs to do

spec/index.bs
```diff
@@ -2924,6 +2967,26 @@ distinguishing browser-initiated FedCM flows from arbitrary requests.
...
+**Consideration**
+
+With the support for both string and JSON object tokens, [=RPs=] must be prepared to handle different
+token formats. The token content remains opaque to the user agent, whether it's a string or a 
+serialized JSON object.
+
+**Recommendations**
+
+1. [=RPs=] should implement robust token parsing logic that can differentiate between traditional
+    string tokens (like JWTs) and serialized JSON objects.
```
cbiesinger: remove "serialized", since they are passed as object and not serialized. though with the second point this is not really needed.

spec/index.bs
```diff
@@ -2924,6 +2967,26 @@ distinguishing browser-initiated FedCM flows from arbitrary requests.
...
+2. [=RPs=] should validate tokens according to the format agreed upon with their [=IDP=] partners.
```
cbiesinger: Typically in markdown you use `1.` in each item so that you don't have to renumber the items in the source when adding or removing an item

spec/index.bs
```diff
@@ -465,6 +473,17 @@ This specification introduces a new type of {{Credential}}, called an {{Identity
     :   <b>{{IdentityCredential/token}}</b>
     ::  The {{IdentityCredential/token}}'s attribute getter returns the value it is set to.
         It represents the minted {{IdentityAssertionResponse/token}} provided by the [=IDP=].
+        The token attribute can return either:
```
npm1: I think we can also remove this. You could replace your note below with these two options for token, but `any` can be more than a string or an object

spec/index.bs
```diff
@@ -2425,7 +2450,15 @@ Every {{IdentityAssertionResponse}} is expected to have members with the followi
 
 <dl dfn-type="dict-member" dfn-for="IdentityAssertionResponse">
     :   <dfn>token</dfn>
-    ::  The resulting token.
+    ::  The resulting token. The value can be either:
+        - A string (e.g., a JWT or other opaque token format)
+        - A JSON object containing structured authentication data
+        
+        The [=user agent=] MUST NOT serialize object tokens to strings. The exact
```
npm1: Why is this in the description of token? It should be clear from the algorithm instead

spec/index.bs
```diff
@@ -3473,6 +3536,24 @@ These schemes have not been adopted by this specification.
...
 </div>
 
+<!-- ============================================================ -->
+### Token Content Privacy ### {#token-content-privacy}
+<!-- ============================================================ -->
+
+The FedCM API treats token content as opaque data, regardless of type (string or object).
+This design ensures:
+
+1. The [=user agent=] does not inspect or validate token contents, maintaining the privacy of the
```
npm1: This is not ensured by our design. The user agent can indeed inspect the token if it wanted to.

spec/index.bs
```diff
@@ -3473,6 +3536,24 @@ These schemes have not been adopted by this specification.
...
+    authentication data.
+2. The token type (string vs. object) does not affect the privacy properties of the API.
```
npm1: Token type is `any`, as mentioned

spec/index.bs
```diff
@@ -3473,6 +3536,24 @@ These schemes have not been adopted by this specification.
...
+3. [=IDPs=] have flexibility in token design without compromising user privacy.
```
npm1: Unclear what token design has to do with user privacy

spec/index.bs
```diff
@@ -3473,6 +3536,24 @@ These schemes have not been adopted by this specification.
...
+4. Sensitive information in tokens is only accessible to the [=RP=] after user consent.
+5. The [=user agent=] MUST pass tokens to the [=RP=] without modification, preserving the exact
```
npm1: Unclear what this has to do with privacy

spec/index.bs
```diff
@@ -3473,6 +3536,24 @@ These schemes have not been adopted by this specification.
...
+    structure provided by the [=IDP=].
+
+Note: While this version allows object tokens for enhanced functionality, it does not change
```
npm1: Maybe remove? Not clear what 'this version' is. Also redundant with 2 above.

spec/index.bs
pottis: Will remove Breaking change note.

spec/index.bs
pottis: Addressed. Thanks.

spec/index.bs
pottis: Addressed. Thanks.

spec/index.bs
pottis: Addressed. Thanks.

spec/index.bs
pottis: changed to **any** instead specifying string or json object.

spec/index.bs
pottis: Removed these lines.

spec/index.bs
pottis: Removed these lines.

spec/index.bs
pottis: removed those lines.

spec/index.bs
pottis: changed to any. Thanks.

spec/index.bs
pottis: Removed. Thanks.

spec/index.bs
pottis: Identity Providers (IDPs) retain full autonomy over token structure and content. They may adopt formats such as strings (e.g., JWTs) or JSON objects, and embed any required data—user identifiers, scopes, expiration details, custom claims, or multiple token types—without constraints imposed by the browser.

spec/index.bs
pottis: Unaware of this flexibility. Considering privacy concerns, what is the rationale behind allowing the user agent or browser to inspect token contents? I’m curious to understand the reasoning behind.

spec/index.bs
pottis: removed point 4 and 5.

spec/index.bs
```diff
@@ -1517,8 +1527,10 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
                 |response|'s [=response/status=]. For example, if the [=response/status=] is 500, it
                 could set it to "server_error", and if the [=response/status=] is 503, it could set
                 it to "temporarily_unavailable".
-        1. Otherwise, if {{IdentityAssertionResponse/token}} was specified, let |tokenString|
-            be |token|'s {{IdentityAssertionResponse/token}}.
+        1. Otherwise, if {{IdentityAssertionResponse/token}} was specified:
+            1. Let |token| be the result of extracting the token from the response.
+            1. Set |credential|.{{IdentityCredential/token}} to |token|.
+            1. The [=user agent=] MUST NOT modify or validate the token content.
```
npm1: Maybe add this sentence to the previous bullet since this in itself is not a step?

spec/index.bs
```diff
@@ -1517,8 +1527,10 @@ To <dfn>fetch an identity assertion</dfn> given a {{USVString}}
...
         1. Otherwise, if {{IdentityAssertionResponse/continue_on}} was specified, run these steps
```
npm1, line 1532: The algorithmic changes here are incomplete. I still see `tokenString` which should no longer be a thing. And wherever we receive it, like from continuation API, it should now allow receiving `any` from there as well. So those algorithms also need to be modified. In particular, resolve() needs to be modified

spec/index.bs
```diff
@@ -2425,7 +2437,9 @@ Every {{IdentityAssertionResponse}} is expected to have members with the followi
 
 <dl dfn-type="dict-member" dfn-for="IdentityAssertionResponse">
     :   <dfn>token</dfn>
-    ::  The resulting token.
+    ::  The resulting token. The value can be any.
```
npm1:
```suggestion
    ::  The resulting token. The value can be of any type.
```

Is this what you mean?

spec/index.bs
npm1: What I am saying is that the design does not ensure that the user agent cannot inspect the token, so this bullet point is incorrect.

spec/index.bs
TallTed: Note that both lines have changes, wrapping the codes in codefences ``` ` ```, which the `.bs` processor will typically transform into `<code> </code>` wrappers.

spec/index.bs
domfarolino: You'd want `{{any}}`, if you're referring to the IDL type that you've changed `USVString` to.

spec/index.bs
pottis: Addressed. Thanks.

spec/index.bs
pottis: Addressed. Thanks.

spec/index.bs
pottis: Updated to use tokenValue in algorithm

spec/index.bs
pottis: corrected the type.

spec/index.bs
pottis: Update the line accordingly.

spec/index.bs
pottis: Addressed. Thanks.

spec/index.bs
```diff
@@ -3473,6 +3517,19 @@ These schemes have not been adopted by this specification.
   </pre>
 </div>
 
+<!-- ============================================================ -->
+### Token Content Privacy ### {#token-content-privacy}
```
cbiesinger, line 3521: I am not sure this section adds much value

npm1 (commented):
Can you change resolve too? It shows as `DOMString token` still.

spec/index.bs
aaronpk: This corresponds to the access token response in OAuth, which is defined as a JSON object: https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.4

Issue #725, opened 2025-05-13
Support showing iframe origins in the UI, when they are third-party


cbiesinger:
We would like to revive the proposal at https://github.com/w3c-fedid/FedCM/issues/449#issuecomment-1515631336

There are two ways to use iframes with fedcm:

1.  The iframe might be an "implementation detail" of the website, e.g. a `rp-static.example` domain to isolate RP cookies from an IDP-provided SDK
2. The iframe might be a genuine third-party embedded into the top-level website to provide some kind of utility. For example, a book editing website might embed a photo editor, which might want you to log in to access previously saved files

In the first case, the expectation is that the iframe origin will not be shown because that origin is not meaningful to the user; it is conceptually first-party even if it is not same-site.

In the second case, the iframe site is important to the user to make an informed decision and to understand why they are asked to log in.

Because both cases are cross-site, the user agent needs additional information to decide which case the iframe falls into. The proposal I linked at the top solves this by letting the IDP tell the user agent (in the client metadata endpoint) whether the two origins are part of the same client.

hlflanagan:
Discussed in [May 20th](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-05-20-FedCM-notes.md) WG/CG call

bvandersloot-mozilla:
To tackle 2, based on our discussion in the meeting: I see two options the IdP has. 1. They have a business relation with the caller. If they misbehave, they can take action. 2. We could add an additional bit to the token request signifying the user interface shown (as we do for autoreauthn).

yi-gu:
@bvandersloot-mozilla could you clarify a bit more on the downside of the proposal?

The alternative feels like more burden on the RP side, e.g. figuring out whether to show itself (may not be straightforward), calling the API properly.

> We could add an additional bit to the token request signifying the user interface shown (as we do for autoreauthn).

The UX wouldn't be ideal because if case of failure (could be an oversight by the iframe) , there would be no way back for users to sign in. However, with the proposal, this could be avoided.

hlflanagan:
Discussed in [FedID CG/WG call, 24 June 2025](https://github.com/w3c-fedid/meetings/blob/main/2025/2025-06-24-FedCM-notes.md)

cbiesinger:
@bvandersloot-mozilla here is another use case for a client metadata endpoint independent of privacy policy and terms of service: #758

cbiesinger:
[...]

PR #774, opened 2025-08-26
[FedCM] Specify third party iframe display support

cbiesinger:
Fixed: #725


cbiesinger:
I think this is ready for review now

spec/index.bs
```diff
@@ -1628,6 +1628,10 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
             does not contain the fields required by the IDP. This is to enable user agents that
             do not support showing a permission prompt.
 
+        1. If |metadata| is not null or failure, and |metadata|.
+            {{IdentityProviderClientMetadata/client_matches_top_frame_origin}} is `false`, this dialog
+            MUST show the [=host/registrable domain=] of |globalObject|'s [=associated Document=]'s
+            [=Document/origin=] to the user.
```
TallTed: Eliminating the space that will otherwise appear between `|metadata|.` and `{{` —
```suggestion
        1. If |metadata| is not null or failure, and 
            |metadata|.{{IdentityProviderClientMetadata/client_matches_top_frame_origin}}
            is `false`, this dialog
            MUST show the [=host/registrable domain=] of |globalObject|'s 
            [=associated Document=]'s [=Document/origin=] to the user.
```

TallTed (changes requested):

spec/index.bs
```diff
@@ -1667,6 +1671,17 @@ an {{IdentityProviderRequestOptions}} |provider|, run the following steps. This
     1. Let |clientMetadataUrl| be the result of [=computing the manifest URL=] given |provider|,
         |config|["{{IdentityProviderAPIConfig/client_metadata_endpoint}}"], and |globalObject|.
     1. If |clientMetadataUrl| is failure, return failure.
+    1. Let |embedderOrigin| be the |globalObject|'s [=Window/navigable=]'s
+        [=navigable/top-level traversable=]'s [=navigable/active document=]'s [=Document/origin=].
+    1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
+    1. Set the [=URL-query string=] of |clientMetadataUrl| to the [=string/concatenation=] of the
+        following strings:
+        1. `"client_id="`
```
npm1, line 1695: I guess this seems fine but why is this one needed?

spec/index.bs
```diff
@@ -1709,6 +1724,7 @@ an {{IdentityProviderRequestOptions}} |provider|, run the following steps. This
 dictionary IdentityProviderClientMetadata {
   USVString privacy_policy_url;
   USVString terms_of_service_url;
+  boolean client_matches_top_frame_origin = true;
```
npm1: This is strongly discouraged. https://w3ctag.github.io/design-principles/#optional-parameters

spec/index.bs
cbiesinger: Thanks, commited a variant of that.

spec/index.bs
cbiesinger: This was missing from the spec. We do send it, and we should send it, because IDPs need it to look up the correct PP/TOS.

Note that the example in https://w3c-fedid.github.io/FedCM/#idp-api-client-id-metadata-endpoint already contains the client ID.

spec/index.bs
cbiesinger: OK. What's your alternative proposal?

I'm not supposed to use a negative in optional arguments either (https://w3ctag.github.io/design-principles/#naming-optional-parameters) so...?

cbiesinger: maybe: `client_is_third_party_to_top_frame_origin`?

npm1: Yea that's one option. Or just make the default set to false (although I see arguably this might be a bigger change given Chromium's existing deployment)

cbiesinger:
@pottis could you review this PR as well?

spec/index.bs
```diff
@@ -1667,6 +1671,17 @@ an {{IdentityProviderRequestOptions}} |provider|, run the following steps. This
...
         |config|["{{IdentityProviderAPIConfig/client_metadata_endpoint}}"], and |globalObject|.
     1. If |clientMetadataUrl| is failure, return failure.
+    1. Let |embedderOrigin| be the |globalObject|'s [=Window/navigable=]'s
+        [=navigable/top-level traversable=]'s [=navigable/active document=]'s [=Document/origin=].
+    1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
+    1. Set the [=URL-query string=] of |clientMetadataUrl| to the [=string/concatenation=] of the
+        following strings:
+        1. `"client_id="`
+        1. |provider|.{{IdentityProviderConfig/clientId}}
+        1. If the [=host/registrable domain=] of |rpOrigin| is different from the
```
npm1: I think this should use not [=same site=] instead.

spec/index.bs
```diff
@@ -1628,6 +1628,10 @@ an {{IdentityProviderAPIConfig}} |config|, an {{IdentityProviderRequestOptions}}
             does not contain the fields required by the IDP. This is to enable user agents that
             do not support showing a permission prompt.
 
+        1. If |metadata| is not null or failure, and
```
npm1, line 1647: I think this along with the IdentityProviderBranding mention above are not placed correctly here since there are other UIs that could be shown. Instead they could go in the "Build UI" step (or perhaps they need to be pointed out in both places?)

spec/index.bs

spec/index.bs
cbiesinger: Hmm so the problem is client metadata is an extension.

I think you're right that it needs mention of this though. Can I add extensions inline or do they need a separate section?

npm1: Right. Yea we could add a link to wherever in the core part you want to add it and then add the extension right after that algorithm, linking to where it would occur. I think we've done this already in other places

cbiesinger: Done, please take another look.

spec/index.bs
```diff
@@ -1122,9 +1136,11 @@ An extension may use the following instead of the [=create identity credential/s
...
                         algorithm, but we do not want to show an extra dialog in this case.
                 1. Otherwise:
-                    1. Set |permission| to the result of running the [=request permission to sign-up=]
-                        algorithm with |selectedAccount|, the relevant |config|, the relevant |provider|,
-                        and |globalObject|.
+                    1. Let |providerOrigin| be the [=/origin=] of the relevant
+                        |provider|'s {{IdentityProviderConfig/configURL}}.
+                    1. Set |permission| to the result of running the [=request permission to
+                        sign-up=] algorithm with |selectedAccount|, the relevant |provider|,
+                        |clientMetadataMap|[|providerOrigin|], and |globalObject|.
```
npm1: The map was set with configURL as key and now it is using origin? Which one is it

spec/index.bs
```diff
@@ -1080,6 +1081,19 @@ When asked to <dfn for="create identity credential">fetch accounts</dfn> given a
 <div class="issue" heading="extension">
 An extension may use the following instead of the [=create identity credential/show accounts=] step, where
```
npm1, line 1083: Maybe wrap these in div algorithm to get things like variable highlighting?

npm1 (approved):

spec/index.bs
cbiesinger: ?
it was set as (line 1091):
```
            |rpOrigin|, set |clientMetadataMap|[|providerOrigin|] to the result of running [=fetch the client metadata=] with
```

But I actually changed it to configURL because that is a bit simpler.

spec/index.bs
cbiesinger: Done although that wasn't new. Fixed some bikeshed warnings that were triggered by this.

PR #786, opened 2025-09-26
Update to explainer for Structured data support

pottis:
Refactored explainer in removing IDL references and added examples for rejected solutions

### DIFF
--- a/explorations/structured_data_support.md
+++ b/explorations/structured_data_support.md
@@ -30,84 +30,169 @@ Allow the `token` field to accept any valid JSON type (`any`) instead of restric
 
 #### 1. ID Assertion Endpoint Response
 
-The `token` field type changes from `string` to `any`, allowing structured JSON objects.
-
-#### 2. IdentityProvider.resolve() Parameter Type
-
-**Before:**
-```typescript
-interface IdentityProvider {
-  resolve(token: string): Promise<IdentityCredential>;
+**Current (String only):**
+```json
+{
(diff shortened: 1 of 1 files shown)


explorations/structured_data_support.md
```diff
@@ -30,84 +30,169 @@ Allow the `token` field to accept any valid JSON type (`any`) instead of restric
...
+```javascript
+let tokenData;
+if (typeof credential.token === 'string') {
+  tokenData = JSON.parse(credential.token);
+} else {
+  tokenData = credential.token;
+}
+```
+
+### Why Not Support Both Formats Simultaneously?
```
npm1: How is this different from the one saying separate fields?

explorations/structured_data_support.md
```diff
@@ -30,84 +30,169 @@ Allow the `token` field to accept any valid JSON type (`any`) instead of restric
...
-- **Eliminates precedence ambiguity**: Clear which token value is authoritative
-- **Reduces security risks**: Fewer code paths mean fewer vulnerabilities
-- **Simplifies validation**: Single validation path reduces error opportunities
-- **Improves developer experience**: Cleaner, more predictable API behavior
+**Benefits:**
+- Clear semantics with single authoritative value
+- Simpler validation and fewer security risks
+- Better developer experience
 
 ### Rejected Alternatives Summary
```
npm1: Im not sure this summary is still accurate? Do we need it?

pottis: removed

explorations/structured_data_support.md
pottis: removed

PR #787, opened 2025-10-03
Update service-workers mode to all for credentialed requests

### LINKED ISSUE #80: Mediation oriented API too limited - Bearer credentials
I was reviewing the mediation-oriented API from WebID as a successor to the permissions-oriented API. I work for Microsoft in the Identity space. I observed that the mediation-oriented API is limited to Bearer authentication - "the browser will then issue a credentialed GET request." Bearer tokens are increasingly seen as a legacy technology in the Identity space and some Identity services already support successor technologies like [DPoP](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-dpop). WebID should avoid constraining auth to such a legacy technology. However, it is a reasonable default.

One option might be [service workers](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API). If WebID permitted IDPs to return a "service_worker" entry as part of of their well-known metadata, and then 
have the WebID code effectively trigger `ServiceWorkerContainer.register()` for the URI. Then, the service worker could customize the subsequent account list and token fetch request and responses. In this manner, WebID could default to Bearer credentials, but still support customization, as to avoid over-constraining the Identity space and becoming a blocker to future security features.

will-bartlett:
Resolves #80 . Service workers enable powerful customizations for the identity provider (IdP). They must be restricted to the credentialed endpoints to maintain FedCM's privacy posture.

***
<a href="https://pr-preview.s3.amazonaws.com/will-bartlett/FedCM/pull/787.html" title="Last updated on Oct 4, 2025, 6:22 PM UTC (9c706d2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/787/1bfbf91...will-bartlett:9c706d2.html" title="Last updated on Oct 4, 2025, 6:22 PM UTC (9c706d2)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -164,7 +164,7 @@ could be implemented.
           }]
         }
       });
-
+      
       // Handle the token - it could be a string or JSON object
       let token = credential.token;
       if (typeof token === 'object' && token !== null) {
@@ -769,7 +769,7 @@ dictionary IdentityProviderRequestOptions : IdentityProviderConfig {
         interested in, or "any" if the [=RP=] wants any account associated with at least one domain
         hint. If provided, the user agent will not show accounts which do not match the domain hint
         value.
-
+        
     Note: "nonce" is to be passed within {{IdentityProviderRequestOptions/params}}.
 </dl>
 
@@ -887,7 +887,7 @@ the exception immediately.
             * Show an identity provider selector and only proceed once
                   the user confirms their desire to use an identity provider.
             * Provide a button in the URL bar and wait for user input
-                  before continuing.
+                  before continuing. 
             * Use previous user behavior to determine not to show any
                   UI to the user and silently reject the request.
         </div>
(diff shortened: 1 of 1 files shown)


will-bartlett:
Now whitespace change free

will-bartlett:
@npm1 sounds good, let me know how that goes

EmLauber:
@npm1 any updates on the privacy discussion?

will-bartlett:
I will add an explanatory note here, to make this more clear.

npm1:
Expanding security/privacy does seem applicable, but I agree that this doesnt introduce new UI or autosignin considerations

cbiesinger:
This change is incorrect. It would let the RP's service worker intercept the requests, which is not what we want.

cbiesinger:
> @pottis none of those changes are proposed. Where did all that come from?
> 
> I think you may have fundamentally misunderstood this PR. This PR allows the **Identity Provider** to set a service worker. This is the normal way fetch works - it is the origin of _destination_ of the fetch request whose service worker is run as part of fetch. It is also not the intention that service workers require silent / no UI modes.
> 
> Your comments seem to relate to how a **Relying Party** could use install a service worker to invoke and or customize FedCM. That's a wholly separate topic.

I understand that this is your *intention* but that is not what the PR does, because that's not how service workers normally work.

Because this is not how service workers normally work, this may require a monkey patch to the service worker spec, or perhaps describing more informally how this is supposed to work...

Essentially, this is a form of https://developer.chrome.com/blog/foreign-fetch

CC @domfarolino

will-bartlett:
> I understand that this is your _intention_ but that is not what the PR does, because that's not how service workers normally work.

Thanks. You are right. @EmLauber fyi. I think I confused myself when reading about loginUrl. There, the FedCM spec emphasizes that the flow is a top-level traversable, i.e. it is not foreign fetch, but for the non-interactive requests, you are absolutely correct.

Perhaps there's a better way to factor this than as a service worker (and thus no need to monkey patch foreign fetch into service worker spec). The FedCM config file could simply point to some sort of javascript file (a web worker?), and the procedures for the various algorithms could optionally call the web worker, if supplied in the config file. I will try to find some time to play around with that factoring.

domfarolino:
Great to know, thanks @samuelgoto. Has there been any outreach to other browsers involved in FedCM discussions? cc @bvandersloot-mozilla

PR #802, opened 2026-01-22
FedCM Service Worker Interception

pottis:


explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+
+## Proposed Solution
+
+Enable IDPs to register Service Workers that can intercept FedCM requests to specific endpoints, while maintaining security by bypassing Service Workers for configuration and metadata endpoints.
+
+### How It Works
+
+#### 1. Destination-Based Interception (Standard SW Behavior)
+
+When the browser makes a FedCM request from `rp.com` to `idp.example`:
```
npm1: Which fetch are you talking about? The ID assertion I think? That is the only one for which the properties are as below

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+- Initiator: https://rp.com
+- Mode: cors
+- Credentials: include
+
+Service Worker Routing:
+- Storage Key: StorageKey(idp.example, idp.example)  ← Based on destination
+- Result: idp.example's Service Worker intercepts the request
+```
+
+**This is standard Service Worker behavior** as defined by the [W3C Service Worker specification](https://w3c.github.io/ServiceWorker/): Just like `<img src="https://cdn.example/logo.png">` is intercepted by `cdn.example`'s Service Worker (not the page's SW), FedCM requests to `idp.example` are intercepted by `idp.example`'s Service Worker.
```
npm1: This paragraph and the example are incorrect. Are you not in the email thread where this was discussed at length?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+          }), {
+            headers: { 'Content-Type': 'application/json' }
+          });
+        })
+    );
+  }
+});
+```
+
+**Result**: User can authenticate even offline, enabling access to cached content and limited functionality.
```
npm1: I guess this presuposes the uncredentialed fetches are cached? Eg if the config/well-known fetches fail, we dont even get to the credentialed ones

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+
+### Performance Optimization via Caching
+
+**Scenario**: Popular IDP serves millions of authentication requests daily. Account information rarely changes but is fetched on every sign-in.
+
+**Without Service Worker**:
+- Every authentication requires a network round-trip to fetch accounts
```
npm1: Cant credentialed requests also be cached? Why does this need service workers?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+    );
+  }
+});
+```
+
+**Result**: 99.9% authentication availability even during partial outages.
+
+### Progressive Web App (PWA) Integration
+
+**Scenario**: IDP provides a PWA that users install. The PWA should work offline and provide seamless authentication.
```
npm1: If the PWA is an IDP provided one, how does it use FedCM? I'm not sure I follow this example.

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+    );
+  }
+});
+```
+
+**Result**: Consistent authentication experience between PWA usage and third-party RP authentication.
+
+### Custom Token Generation Logic
+
+**Scenario**: Enterprise IDP needs to generate different token types based on RP requirements, apply rate limiting, or implement fraud detection.
```
npm1: This use case I also don't follow. This logic can just live in the server side? Is it a bad thing for it to live there?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+    );
+  }
+});
+```
+
+**Result**: Reduced server load, faster token generation, improved security through client-side checks.
+
+### Pre-fetching and Background Sync
+
+**Scenario**: IDP wants to pre-fetch account information and credentials before user initiates sign-in.
```
npm1: This is interesting, although might be limited due to lifetime of the service worker, right?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+## Privacy and Security Considerations
+
+### Privacy Impact
+
+**IDP Sees All RP Requests (Acceptable)**
+
+The IDP's Service Worker runs in a single instance and sees requests from all RPs:
+
+```javascript
+// Single IDP Service Worker instance sees:
```
npm1: Isn't this missing the accounts requests for the example to be complete?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+- This matches the current FedCM model where IDP servers see all RP requests
+- No worse than traditional OAuth where IDP servers log all authentication attempts
+- Service Workers have same privacy characteristics as IDP backend servers
+- User consent is still required before any tokens are issued
+
+### Security Impact
+
+**Origin Isolation (Maintained)**
+
+Service Workers can **only** intercept requests to their own origin. This is enforced by the browser's Service Worker infrastructure:
```
npm1: This is not true.

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+
+**Key point**: RPs don't need to know or care whether the IDP uses Service Workers. The FedCM API contract remains unchanged.
+
+## Design Decisions and Alternatives Considered
+
+### Why Selective Endpoint Interception?
+
+**Rejected Alternative**: Allow Service Workers for all endpoints.
+
+**Decision Rationale**: Configuration endpoints must remain trustworthy and unmodified to maintain security guarantees. Only authentication-specific endpoints benefit from Service Worker capabilities without compromising trust relationships.
```
npm1: IIRC the reason was that it would trivially allow the IDP to correlate which user is visiting which websites by looking at the client-md fetches vs the accounts fetches

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+
+### Why Destination-Based Interception?
+
+**Decision Rationale**: This follows standard Service Worker behavior. Just like `<img src="https://cdn.example/logo.png">` is intercepted by `cdn.example`'s Service Worker, FedCM requests to `idp.example` are intercepted by `idp.example`'s Service Worker. No new concepts are introduced to the platform.
```
npm1: False?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+
+### Why Not Communicate RP Identity via Service Worker APIs?
```
npm1: I mean, the RP identity is going to be communicated in the ID token fetch, no? I'm not sure what this means.

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+
+**Decision Rationale**: The RP's identity flows through the FedCM protocol itself (via `client_id` in request bodies), not through Service Worker client APIs. This maintains consistency with existing OAuth/OIDC patterns and avoids introducing new cross-origin communication channels.
+
+### Rejected Alternatives Summary
```
npm1: Do we really need this?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+
+1. **Allow Service Workers for All Endpoints**: Rejected - Configuration endpoints must remain trustworthy and unmodified
+2. **Use WebTransport or WebSockets**: Rejected - Doesn't provide offline support or caching capabilities
+3. **Browser-Managed Cache**: Rejected - Less flexible than Service Worker, harder to customize, no offline logic support
+4. **Require Service Worker to Signal Intent**: Rejected - Adds complexity. Standard Service Worker registration is sufficient
+
+## Implementation Considerations
+
+### Browser Implementation
```
npm1: Probably doesnt belong in explainer

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+https://github.com/w3c-fedid/FedCM/issues/80
+
+## Summary
+
+This document describes how Service Workers can intercept and handle Federated Credential Management (FedCM) network requests, enabling Identity Providers (IDPs) to provide offline support, caching, and custom authentication logic while maintaining security and privacy guarantees.
+
+## Problem Statement
+
+Identity Providers need to:
+- Provide authentication services even when the user is offline or has poor connectivity
```
samuelgoto: How would this provide offline authentication services, if all of the un-credentialed requests (e.g. config, well-known, metadata, etc) are done via HTTP rather than the Service Worker?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+✅ idp.example's SW can intercept: https://idp.example/token
+❌ idp.example's SW cannot intercept: https://other-idp.com/token
+❌ rp.com's SW cannot intercept: https://idp.example/token
+```
+
+**Security guarantee**: No cross-origin interception is possible.
+
+**Configuration Endpoint Protection**
+
+Critical configuration endpoints (`/.well-known`, `/config.json`, `/client_metadata`) **bypass Service Workers entirely**:
```
samuelgoto: Isn't this more of a privacy consideration than a security consideration? That is, why is it that we bypass the well-known, the config and the client_metadata endpoints? Is it because of security (and if so, why?)? or is it because the SW can have internal state and hence can know who the user is (and hence could make all of the uncredentialed requests effectively credentialed)?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,680 @@
...
+**Security guarantee**: Service Workers cannot bypass SameSite cookie protections.
+
+## Examples
+
+### IDP Service Worker Registration
+
+```javascript
+// On https://idp.example - register Service Worker
+// Per https://w3c.github.io/ServiceWorker/#navigator-service-worker-register
+if ('serviceWorker' in navigator) {
```
samuelgoto, line 379: When would you expect this to happen? 

Doesn't this presuppose that you have navigated to the IdP ahead of time, before navigating to the RP? If so, what happens when you don't (e.g. in active mode)?

explorations/service_worker_interception.md
samuelgoto: Right, this seems way more of  a privacy consideration than a security one, right?

explorations/service_worker_interception.md
pottis: removed.

explorations/service_worker_interception.md
pottis: Removed.

explorations/service_worker_interception.md
pottis: Yes

explorations/service_worker_interception.md
pottis: Service Workers are [origin-based, not page-based](https://w3c.github.io/ServiceWorker/#service-worker-concept). Per [§ 2.1](https://w3c.github.io/ServiceWorker/#service-worker-concept), "a service worker executes in the registering service worker client's origin", and per [§ 2.3](https://w3c.github.io/ServiceWorker/#service-worker-registration-concept), registrations are tied to a specific origin and scope URL.

explorations/service_worker_interception.md
pottis: Provided if config, well-known, metadata gets cached by the user-agent, though FedCM spec doesnt mention on these end-points caching.

explorations/service_worker_interception.md
pottis: Yes. Offline capabilities via the service worker are available only when uncredentialed requests are cached by the browser.

explorations/service_worker_interception.md
npm1: Fix text?

explorations/service_worker_interception.md
npm1: But even if those uncredentialed fetches are cached, if the user is offline then those fetches will fail because they are not handled by the service worker? So I don't think this use case is satisfied with this proposal

explorations/service_worker_interception.md
npm1: Yes but the service worker's client's origin is the browsing context's origin, not the resource's origin. https://w3c.github.io/ServiceWorker/#control-and-use-window-client

explorations/service_worker_interception.md
pottis: First-time FedCM flow (no prior IDP visit):
1. User on rp.com clicks "Sign in with Example IDP"
2. Browser fetches idp.example/accounts → No SW registered → Direct to network
3. Browser fetches idp.example/token → No SW registered → Direct to network
4. Authentication succeeds, but no caching/fallback benefits

Subsequent FedCM flows (after user has visited IDP):
1. User on different-rp.com clicks "Sign in with Example IDP"
2. Browser fetches idp.example/accounts → SW intercepts → Can serve cached response
3. Browser fetches idp.example/token → SW intercepts → Can apply custom logic

explorations/service_worker_interception.md
pottis: Adddressed

explorations/service_worker_interception.md
pottis: Removed this section as it stands confusing. Intent was to Help developers avoid the event.clientId pitfall when implementing IDP Service Workers.
For IDP to identity dont rely on event.clientId rather use client_id from POST body..

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+
+## Status
+
+This explainer documents a **proposed change** to the FedCM specification. The current FedCM spec sets `service-workers mode: "none"` for all requests. This proposal changes select endpoints to use `service-workers mode: "all"`, enabling IDP Service Worker interception.
+
+**Proposed Spec Changes:**
+
+| Endpoint | Current Spec | Proposed |
+|----------|-------------|----------|
+| `/accounts` | `service-workers mode: "none"` | `service-workers mode: "all"` |
```
cbiesinger, line 22: Hm, technically the proposal is to allow the *IDP*'s service worker to intercept the requests, whereas service-workers mode would traditionally let the RP intercept it.

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+- Apply client-side rate limiting and fraud detection
+- Pre-fetch and actively manage authentication resources
+
+**Key Question**: Can credentialed requests already be cached?
+
+**Yes** - Standard HTTP caching works for credentialed FedCM requests (accounts, token endpoints) when `Cache-Control` headers allow. Browsers cache responses based on these directives.
+
+**Why Service Workers?** Service Workers add **programmable control** beyond passive HTTP caching:
+
+**Critical limitation of HTTP caching**: Standard browser HTTP cache is **passive** - it won't serve cached responses when the network fails. The cache only works when network succeeds and `Cache-Control` headers allow reuse.
```
cbiesinger: This is a bit unclear. Is this specifically talking about when the cache is expired and the network request is made to revalidate the cached content, but the network is not available and you want to reuse the content anyway?

If the cached content is not expired yet, I don't think this description is correct.

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+
+**Service Worker capabilities**:
+- **Active fallback logic**: Return cached responses when `fetch()` fails (HTTP cache can't do this)
+- **Custom strategies**: Stale-while-revalidate, try backup servers, conditional logic based on error type
+- **Response modification**: Add metadata like `{offline: true, limited_scope: true}` to responses
```
cbiesinger: But the RP does not have access to the response and the browser would not understand such metadata. Is this specifically for modifying the token?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+- **Active management**: Background sync, pre-fetching, intelligent cache invalidation
```
cbiesinger: Some of this seems to require an installed PWA (if it relies on periodic background sync), not sure how valuable this is in practice

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+**Key Question**: Can credentialed requests already be cached?
+
+**Yes** - Standard HTTP caching works for credentialed FedCM requests (accounts, token endpoints) when `Cache-Control` headers allow. Browsers cache responses based on these directives.
+
+**Why Service Workers?** Service Workers add **programmable control** beyond passive HTTP caching:
+
+**Critical limitation of HTTP caching**: Standard browser HTTP cache is **passive** - it won't serve cached responses when the network fails. The cache only works when network succeeds and `Cache-Control` headers allow reuse.
+
+**Service Worker capabilities**:
+- **Active fallback logic**: Return cached responses when `fetch()` fails (HTTP cache can't do this)
```
cbiesinger, line 64: Can you elaborate how this works for the token endpoint? As far as I am aware, token responses are typically specific to an RP and also rely on a nonce. How will this work with a missing network?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+| Endpoint | Current Spec | Proposed |
+|----------|-------------|----------|
+| `/accounts` | `service-workers mode: "none"` | `service-workers mode: "all"` |
+| `/token` (id_assertion_endpoint) | `service-workers mode: "none"` | `service-workers mode: "all"` |
+| `/disconnect` | `service-workers mode: "none"` | `service-workers mode: "all"` |
+| `/.well-known/web-identity` | `service-workers mode: "none"` | No change |
+| `/config.json` | `service-workers mode: "none"` | No change |
+| `/client_metadata` | `service-workers mode: "none"` | No change |
+
+## Problem Statement
```
cbiesinger, line 31: I'm curious, is all of this based on specific IDP feedback, or is it speculation for things that might be useful?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+- Mode: no-cors
+- Credentials: include
+- Origin: opaque (privacy-preserving)
+- Destination: webidentity
+- Service-Workers Mode: "all" (proposed change)
+- Client: null
+
+Service Worker Matching:
+- With service-workers mode: "all" and client: null, the browser
+  matches against Service Worker registrations for the request URL's origin
```
cbiesinger, line 94: I had a hard time navigating the spec but could you point to the part of the spec where this happens?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+```
+
+**How this works**: When a FedCM request has `service-workers mode: "all"`, the browser checks if a Service Worker is registered for the request URL's origin (`idp.example`). If found, the SW receives a `FetchEvent`. This is similar to how the browser would handle any fetch to that origin when a SW is registered.
+
+The same mechanism applies to other SW-enabled endpoints (`/token`, `/disconnect`), each with their specific request properties as defined in the FedCM specification.
+
+**Important**: FedCM requests are browser-initiated with `client: null`. This means:
+- The request has no associated document/window client
+- `FetchEvent.clientId` will be empty
+- The RP's identity comes from the POST body (`client_id` parameter), not from SW APIs
```
cbiesinger: This is only true for the token and disconnect endpoints, for what it's worth

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+
+Per the [W3C Service Worker specification § 2.1](https://w3c.github.io/ServiceWorker/#service-worker-concept), Service Workers execute within a specific origin. Only `idp.example`'s Service Worker can intercept requests to `idp.example`.
```
cbiesinger: Hm, I am not sure what the spec is referring to with that line, but in general a service worker can intercept requests made to multiple origins, so this sentence is misleading.

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+| `/token` | Allowed | `"all"` | Token generation - can use custom logic |
+| `/disconnect` | Allowed | `"all"` | Account management - graceful error handling |
+| `/.well-known/web-identity` | Blocked | `"none"` | IDP discovery - privacy protection |
+| `/config.json` | Blocked | `"none"` | Configuration - privacy protection |
+| `/client_metadata` | Blocked | `"none"` | RP metadata - privacy protection |
+| `/picture` | Blocked | `"none"` | Profile images - not authentication-critical |
+
+**Why are configuration endpoints protected?** See [Privacy Considerations](#privacy-considerations) below.
+
+> **Developer Note**: FedCM requests are browser-initiated with `client: null`, so `FetchEvent.clientId` will be empty. To identify the RP in your Service Worker, read the `client_id` parameter from the POST body of `/token` or `/disconnect` requests—this is standard OAuth/OIDC behavior, unchanged by this proposal.
```
cbiesinger: Perhaps mention that the accounts endpoint intentionally does not allow identifying the RP?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+
+## Benefits
+
+### Graceful Degradation During IDP Outages
+
+**Scenario**: IDP experiences temporary server issues or undergoes maintenance.
+
+**Without Service Worker**:
+- All authentication requests fail immediately
```
cbiesinger: Only if the response is not cached or has expired

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+- Users see error messages
+- RPs cannot function
+- No fallback mechanism
+
+**With Service Worker**:
+```javascript
+// IDP's Service Worker provides network resilience
```
cbiesinger: is it useful to only handle the accounts endpoint if the token request will just fail once the user clicks?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+    );
+  }
+});
+```
+
+**Result**: Improved authentication availability during partial outages.
+
+### Performance Optimization via Caching
+
+**Scenario**: Popular IDP serves millions of authentication requests daily. Account information rarely changes but is fetched on every sign-in.
```
cbiesinger: Since this can be addressed with regular cache control headers, it would be better to explain the value-add of a service worker (in this case, apparently stale-while-revalidate, although I am not sure that requires a service worker either)

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+```
+
+**Benefits**:
+- Instant account display for returning users
+- Reduced IDP server load
+- Better user experience with faster UI
+
+### Custom Token Handling Logic
+
+**Scenario**: Enterprise IDP needs to apply rate limiting or implement fraud detection.
```
cbiesinger: I don't think fraud detection can rely on client-side code.

Is the expectation that a single client might possibly send enough requests to be a problem for a server? And this can't be handled server-side?

Did an IDP ask for either of these?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+**Result**: Reduced server load, faster fraud detection, improved security through client-side checks.
+
+### Pre-fetching and Background Sync
+
+**Scenario**: IDP wants to pre-fetch account information before user initiates sign-in.
+
+**With Service Worker**:
+```javascript
+// Background sync for account updates
+// Per https://w3c.github.io/ServiceWorker/#sync-event
```
cbiesinger: That link is broken fwiw

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+
+**Result**: Reduced server load, faster fraud detection, improved security through client-side checks.
+
+### Pre-fetching and Background Sync
+
+**Scenario**: IDP wants to pre-fetch account information before user initiates sign-in.
+
+**With Service Worker**:
+```javascript
+// Background sync for account updates
```
cbiesinger: This example would be better if it also showed where the sync is requested, not just the event listener.

In particular:
- If this is using regular background sync, I am not sure how this is useful in this context
- If this uses periodic background sync, this only works if the IDP has an installed PWA, which seems rare

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+
+**Key Question**: Can credentialed requests already be cached?
+
+**Yes** - Standard HTTP caching works for credentialed FedCM requests (accounts, token endpoints) when `Cache-Control` headers allow. Browsers cache responses based on these directives.
+
+**Why Service Workers?** Service Workers add **programmable control** beyond passive HTTP caching:
+
+**Critical limitation of HTTP caching**: Standard browser HTTP cache is **passive** - it won't serve cached responses when the network fails. The cache only works when network succeeds and `Cache-Control` headers allow reuse.
+
+**Service Worker capabilities**:
```
cbiesinger, line 62: I'm surprised that authentication isn't mentioned at all, since that was the initial reason for asking for this feature in https://github.com/w3c-fedid/FedCM/issues/80 (DPoP in that case)

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+
+## Limitations
+
+### No True Offline Authentication
+
+**Important**: This proposal does **not** enable complete offline authentication. Here's why:
+
+1. Configuration endpoints (`/.well-known/web-identity`, `/config.json`) still bypass Service Workers
+2. The FedCM flow requires successful config fetch **before** reaching `/accounts` or `/token`
+3. If the network is offline during config fetch, the entire flow fails
```
cbiesinger: Only if caching are not used or the response is expired

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+- Graceful error handling
+
+**What SW does NOT enable**:
+- Complete offline authentication
+- Bypassing the config/well-known fetch requirement
+
+### Service Worker Must Be Pre-Registered
+
+For the IDP's Service Worker to intercept FedCM requests:
+1. User must have **previously** visited `https://idp.example`
```
cbiesinger, line 224: Maybe mention that this is not an issue because the user must have visited the IDP to log in anyway

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+2. The SW must be **installed and activated**
+3. Only then can subsequent FedCM requests be intercepted
+
+This means:
+- First-time FedCM flows (user never visited IDP) will not benefit from SW caching
```
cbiesinger: If the user never visited the IDP, the user is not logged in to the IDP (by definition...) and thus the FedCM request will fail anyway. Right?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+- Works best for IDPs with direct user relationships (Google, Microsoft, enterprise IDPs)
+
+## Privacy Considerations
+
+### Why Configuration Endpoints Are Protected
+
+Configuration endpoints (`/.well-known`, `/config.json`, `/client_metadata`) bypass Service Workers for **privacy** reasons, not just security:
```
cbiesinger: What do you mean with "not *just* security"? What is the security reason for this?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+Configuration endpoints are fetched with privacy-preserving properties:
+- `credentials mode: "omit"` - no cookies sent
+- `origin: opaque` - RP origin not revealed
+- `referrer policy: "no-referrer"` - no referrer sent
+
+This prevents the IDP from learning which RP is requesting the configuration.
+
+**Example attack**:
+```javascript
+// Hypothetical malicious SW (if config interception were allowed)
```
cbiesinger: It seems an easier attack would be:
Intercept the client metadata request. In the handler, make a new request to the IDP with credentials included, and also include the RP from the client metadta request.

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+- This matches the current FedCM model where IDP servers see all RP requests
+- No worse than traditional OAuth where IDP servers log all authentication attempts
+- Service Workers have the same visibility as IDP backend servers
+- User consent is still required before any tokens are issued
+
+## Security Considerations
+
+### Origin Isolation
+
+For FedCM requests (which have `client: null`), the browser matches against Service Worker registrations using the **request URL's origin**. Since FedCM requests to `idp.example` target that origin, only `idp.example`'s registered Service Worker can intercept them.
```
cbiesinger: It would be good to link to the part of the spec that says that

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+
+```
+:  [=request/redirect mode=]
+:: "error"
+```
+
+This means:
+- If the server returns a redirect, the request fails
+- Service Workers cannot redirect FedCM requests to different URLs
+- Token theft via redirect attacks is prevented
```
cbiesinger, line 326: I mean, that is technically correct, but the IDP has access to the token already and no incentive to steal it.

If the service worker did want to steal it, it could just make a separate request in the token endpoint fetch and send the token to some other server, right?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+### Request Origin Handling
+
+FedCM requests use different `origin` values depending on the endpoint:
+
+| Endpoint | Origin Value | Privacy Implication |
+|----------|-------------|---------------------|
+| `/.well-known` | Opaque origin | IDP cannot identify requesting RP |
+| `/config.json` | Opaque origin | IDP cannot identify requesting RP |
+| `/accounts` | Opaque origin | IDP cannot identify requesting RP |
+| `/token` | RP's document origin | IDP knows which RP is requesting (required for protocol) |
```
cbiesinger: What does "required for protocol" mean and why is that only the case for the token endpoint and not disconnect or client_metadata?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+| `/disconnect` | RP's document origin | IDP knows which RP is requesting |
+| `/client_metadata` | RP's document origin | IDP knows which RP is requesting |
+
+### Sec-Fetch-Dest Header
```
cbiesinger, line 341: That reminds me, we need to update the note at https://fetch.spec.whatwg.org/#ref-for-destination-type now that webidentity is exposed to JavaScript (in the service worker)

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+
+FedCM requests include `Sec-Fetch-Dest: webidentity`, a [forbidden request-header](https://fetch.spec.whatwg.org/#forbidden-header-name) that cannot be set by JavaScript:
```
cbiesinger: This would probably be clearer if you add something like "including service workers" at the end of this sentence because otherwise it is not clear to me why this section exists in this document

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+credentials mode: "include"     // Permission to try sending cookies
+site_for_cookies: (empty)       // Cross-site request
+
+// Browser enforces SameSite filtering:
+// - SameSite=Strict cookies -> Blocked
+// - SameSite=Lax cookies    -> Blocked
+// - SameSite=None; Secure   -> Allowed
+```
+
+**Security guarantee**: Service Workers cannot bypass SameSite cookie protections.
```
cbiesinger: hm, I am not certain this is correct. A service worker is registered on the IDP's origin, wouldn't a service worker's fetches be considered same-site and therefore include even "Strict" cookies?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,641 @@
...
+- SW matching uses the request URL's origin
+- Only the IDP's registered SW can intercept
+
+This ensures the IDP controls interception of requests to its own endpoints, which is the appropriate trust model.
+
+## Implementation Considerations
+
+### Browser Implementation
+
+1. **Endpoint Classification**: Browser classifies each FedCM request by endpoint type
```
cbiesinger: this entire section seems to describe standard service worker behavior, not sure it is necessary in this document?

explorations/service_worker_interception.md
pottis: Added new section to differentiate IDP vs RP Interception.

explorations/service_worker_interception.md
pottis: Theses are all concerns picked from https://github.com/w3c-fedid/FedCM/issues/80 discusssions

explorations/service_worker_interception.md
pottis: Added a table with possible scenario where SW fits.

explorations/service_worker_interception.md
pottis: Yes, added authenitcation as SW capability.

explorations/service_worker_interception.md
pottis: Yes, it doesnt offer much benefit for token endpoint as token is bound to that specific nonce.

explorations/service_worker_interception.md
pottis: removed this point.

explorations/service_worker_interception.md
pottis: removed this point.

explorations/service_worker_interception.md
pottis: Added link to Service Worker Spec

explorations/service_worker_interception.md
pottis: Yes, eloborated the line.

explorations/service_worker_interception.md
pottis: updated the line.

explorations/service_worker_interception.md
pottis: updated.

explorations/service_worker_interception.md
pottis: Yes, updated the line.

explorations/service_worker_interception.md
pottis: Added a note on the scenario.

explorations/service_worker_interception.md
pottis: removing this sectin as graceful degradation section already covers this.

explorations/service_worker_interception.md
pottis: removed thios section.

explorations/service_worker_interception.md
pottis: removed this section here and also from SW capability.

explorations/service_worker_interception.md
pottis: removed this section.

explorations/service_worker_interception.md
pottis: Updated. Thanks.

explorations/service_worker_interception.md
pottis: Updated. Thanks.

explorations/service_worker_interception.md
pottis: Yes.

explorations/service_worker_interception.md
pottis: removed that phrase. Thanks.

explorations/service_worker_interception.md
pottis: Considered this. Thanks.

explorations/service_worker_interception.md
pottis: Added reference to spec. Thanks.

explorations/service_worker_interception.md
pottis: Removed this section.

explorations/service_worker_interception.md
pottis: removed this section.

explorations/service_worker_interception.md
pottis: Updated in note section.

explorations/service_worker_interception.md
pottis: removed

explorations/service_worker_interception.md
pottis: added in context of service worker.

explorations/service_worker_interception.md
pottis: removed

explorations/service_worker_interception.md
pottis: addressed the concern.

explorations/service_worker_interception.md
pottis: Added a note.

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,560 @@
...
+**Critical limitation of HTTP caching**: Standard browser HTTP cache handles fresh (non-expired) content correctly, but has a gap with stale content during network failures:
+
+| Scenario | HTTP Cache Behavior | Service Worker Can Help? |
+|----------|---------------------|--------------------------|
+| Fresh cache (not expired) | ✅ Serves from cache | Not needed |
+| Stale cache + network available | ✅ Revalidates with server | Not needed |
+| Stale cache + network fails | ❌ Returns network error | ✅ Can serve stale content |
+| No cache + network fails | ❌ Returns network error | ✅ Can return graceful error |
+
+The key gap is the third scenario: when cached content has expired and the browser attempts revalidation, but the network is unavailable, the browser returns a network error rather than serving the stale content. Service Workers enable "stale-if-error" behavior that HTTP caching alone doesn't provide.
```
npm1, line 58: This makes sense for standalone requests, but FedCM requires several requests. How does this help if the uncredentialed fetches are performed first, and they would fail if the network is unavailable?

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,560 @@
...
+
+The same mechanism applies to other SW-enabled endpoints (`/token`, `/disconnect`), each with their specific request properties as defined in the FedCM specification.
+
+**Important**: FedCM requests are browser-initiated with `client: null`. This means:
+- The request has no associated document/window client
+- `FetchEvent.clientId` will be empty
+- For `/token` and `/disconnect` (POST requests), the RP's identity comes from the POST body (`client_id` parameter), not from SW APIs
+- For `/accounts` (GET request), the RP's identity is not available - the IDP intentionally doesn't know which RP is requesting until user consent
+
+**Why does the IDP's SW intercept (not the RP's)?** In general, a Service Worker can intercept cross-origin requests made by pages it controls. However, FedCM requests have `client: null` - there's no associated page, so there's no "page's active SW" to intercept. The browser falls back to matching based on the request URL's origin, which is why `idp.example`'s SW intercepts requests to `idp.example`.
```
npm1: I mean we can certainly spec this behavior but I don't think we can claim it is currently what would happen? I think only navigations would be intercepted by the resource's origin's service worker, and FedCM fetches are not navigations.

explorations/service_worker_interception.md
```diff
@@ -0,0 +1,560 @@
...
+### No True Offline Authentication
+
+**Important**: This proposal does **not** enable complete offline authentication. Here's why:
+
+1. Configuration endpoints (`/.well-known/web-identity`, `/config.json`) still bypass Service Workers
+2. The FedCM flow requires successful config fetch **before** reaching `/accounts` or `/token`
+3. If the network is offline during config fetch and the response is not cached or has expired, the entire flow fails
+
+**What SW enables**:
+- Fallback during **partial** outages (auth servers down, but network available)
```
npm1, line 213: Makes me wonder if enabling it on the config and well-known would improve this use-case. The one that is stricly off-limits is the client md fetch but that one failing does not fail the fedcm flow.

explorations/service_worker_interception.md
```diff
@@ -26,28 +26,41 @@ This explainer documents a **proposed change** to the FedCM specification. The c
 | `/config.json` | `service-workers mode: "none"` | No change |
 | `/client_metadata` | `service-workers mode: "none"` | No change |
 
+> **Important Distinction - IDP vs RP Interception**: Traditional `service-workers mode: "all"` allows the **page's** Service Worker (the RP's SW) to intercept subresource fetches. However, FedCM requests are browser-initiated with `client: null` (no document context), so there is no "page SW" to intercept them. Instead, the browser matches against Service Worker registrations using the **request URL's origin** (the IDP). This means the **IDP's** Service Worker intercepts requests to IDP endpoints—not the RP's.
```
TallTed:
```suggestion
> **Important Distinction - IDP vs RP Interception**: Traditional `service-workers mode: "all"` allows the **page's** Service Worker (the RP's SW) to intercept subresource fetches. However, FedCM requests are browser-initiated with `client: null` (no document context), so there is no "page SW" to intercept them. Instead, the browser matches against Service Worker registrations using the **request URL's origin** (the IDP). This means the **IDP's** — not the RP's — Service Worker intercepts requests to IDP endpoints.
```

explorations/service_worker_interception.md
```diff
@@ -77,19 +90,24 @@ Request Properties:
...
+**How this works**: When a FedCM request has `service-workers mode: "all"` and `client: null`, the browser matches against Service Worker registrations using the request URL's origin (`idp.example`). If a SW is registered for that origin, it receives a `FetchEvent`. This is the standard SW matching behavior for requests without an associated document client.
 
 The same mechanism applies to other SW-enabled endpoints (`/token`, `/disconnect`), each with their specific request properties as defined in the FedCM specification.
 
 **Important**: FedCM requests are browser-initiated with `client: null`. This means:
 - The request has no associated document/window client
 - `FetchEvent.clientId` will be empty
-- The RP's identity comes from the POST body (`client_id` parameter), not from SW APIs
+- For `/token` and `/disconnect` (POST requests), the RP's identity comes from the POST body (`client_id` parameter), not from SW APIs
+- For `/accounts` (GET request), the RP's identity is not available - the IDP intentionally doesn't know which RP is requesting until user consent
```
TallTed:
```suggestion
- For `/accounts` (GET request), the RP's identity is not available — the IDP intentionally doesn't know which RP is requesting until user consent
```

explorations/service_worker_interception.md
```diff
@@ -77,19 +90,24 @@ Request Properties:
...
+
+**Why does the IDP's SW intercept (not the RP's)?** In general, a Service Worker can intercept cross-origin requests made by pages it controls. However, FedCM requests have `client: null` - there's no associated page, so there's no "page's active SW" to intercept. The browser falls back to matching based on the request URL's origin, which is why `idp.example`'s SW intercepts requests to `idp.example`.
```
TallTed:
```suggestion
**Why does the IDP's SW intercept (not the RP's)?** In general, a Service Worker can intercept cross-origin requests made by pages it controls. However, FedCM requests have `client: null` — there's no associated page, so there's no "page's active SW" to intercept. The browser falls back to matching based on the request URL's origin, which is why `idp.example`'s SW intercepts requests to `idp.example`.
```

explorations/service_worker_interception.md
```diff
@@ -77,19 +90,24 @@ Request Properties:
...
 
-Per the [W3C Service Worker specification § 2.1](https://w3c.github.io/ServiceWorker/#service-worker-concept), Service Workers execute within a specific origin. Only `idp.example`'s Service Worker can intercept requests to `idp.example`.
+> **Note**: If the RP has its own Service Worker registered, it will **not** intercept FedCM requests. The RP's SW only intercepts requests made by pages it controls. Since FedCM requests are browser-initiated (`client: null`), they bypass the RP's SW entirely - there is no conflict between RP and IDP Service Workers.
```
TallTed, line 114:
```suggestion
> **Note**: If the RP has its own Service Worker registered, it will **not** intercept FedCM requests. The RP's SW only intercepts requests made by pages it controls. Since FedCM requests are browser-initiated (`client: null`), they bypass the RP's SW entirely — there is no conflict between RP and IDP Service Workers.
```

explorations/service_worker_interception.md
```diff
@@ -107,7 +125,7 @@ Not all endpoints allow Service Worker interception. This is enforced via `servi
 
 **Why are configuration endpoints protected?** See [Privacy Considerations](#privacy-considerations) below.
 
-> **Developer Note**: FedCM requests are browser-initiated with `client: null`, so `FetchEvent.clientId` will be empty. To identify the RP in your Service Worker, read the `client_id` parameter from the POST body of `/token` or `/disconnect` requests—this is standard OAuth/OIDC behavior, unchanged by this proposal.
+> **Developer Note**: FedCM requests are browser-initiated with `client: null`, so `FetchEvent.clientId` will be empty. To identify the RP in your Service Worker, read the `client_id` parameter from the POST body of `/token` or `/disconnect` requests—this is standard OAuth/OIDC behavior, unchanged by this proposal. Note that the `/accounts` endpoint (GET request) intentionally does not include RP identity—this is a privacy protection so the IDP cannot track which RPs a user visits before the user grants consent.
```
TallTed:
```suggestion
> **Developer Note**: FedCM requests are browser-initiated with `client: null`, so `FetchEvent.clientId` will be empty. To identify the RP in your Service Worker, read the `client_id` parameter from the POST body of `/token` or `/disconnect` requests — this is standard OAuth/OIDC behavior, unchanged by this proposal. Note that the `/accounts` endpoint (GET request) intentionally does not include RP identity — this is a privacy protection so the IDP cannot track which RPs a user visits before the user grants consent.
```

explorations/service_worker_interception.md
pottis: Addressed. Thanks.

explorations/service_worker_interception.md
pottis: Addressed. Thanks.

explorations/service_worker_interception.md
pottis: Addressed. Thanks.

explorations/service_worker_interception.md
pottis: Addressed. Thanks.

explorations/service_worker_interception.md
pottis: Adderessed. Thanks.

explorations/service_worker_interception.md
pottis: Added **spec-coordination-note**. Thanks.

explorations/service_worker_interception.md
pottis: Good point. Allowing SW on config/well-known (no RP identity) while blocking client_metadata would enable true offline resilience without privacy risk.

explorations/service_worker_interception.md
pottis: Added a caveat.

TallTed (approved):

Issue #809, opened 2026-02-02
Use of apex domain is hard for organizations with different services owning different subdomains


will-bartlett:
The algorithm for rootUrl contains:

> 5. Let rootUrl be a new [URL](https://url.spec.whatwg.org/#concept-url).
> 
> 6. Set rootUrl’s [scheme](https://url.spec.whatwg.org/#concept-url-scheme) to configUrl’s [scheme](https://url.spec.whatwg.org/#concept-url-scheme).
> 
> 7. Set rootUrl’s [host](https://url.spec.whatwg.org/#concept-url-host) to configUrl’s [host](https://url.spec.whatwg.org/#concept-url-host)’s [registrable domain](https://url.spec.whatwg.org/#host-registrable-domain).
> 
> 8. Set rootUrl’s [path](https://url.spec.whatwg.org/#concept-url-path) to the [list](https://infra.spec.whatwg.org/#list) «".well-known", "web-identity"».

This organization of domains is not ideal for some organizations (such as my own). There are two problems.

First, this algorithm results in an apex domain most of the time (except for subdomains of domain on the public suffix list). Apex domains have additional complexity for DNS - see https://www.namesilo.com/blog/en/dns/aliasaname-at-the-root-making-apex-domains-work-with-modern-cdns for one explanation (first result from search engine with query "why apex domains are hard dns cname"). In particular, they do not support CNAMEs (because they must hold NS records), and so must use either ALIAS records (which degrade geo-affinity) or A/AAAA records directly (which are not how modern cloud services typically manage DNS to support failover and autoscale). 

Second, it's extremely common when an organization (let's say Contoso) owns a domain (let's say contoso.com) for different subdomains to have different purposes, and point to different web services. For example, login.contoso.com might CNAME to a white label authentication service (like Entra B2C), while support.contoso.com might CNAME to a white label ticketing system, while blog.contoso.com might CNAME to a Substack organization, etc. In such setups, the apex domain (contoso.com) is often completely unrouted, or routed to service different than the authentication service. It makes FedCM deployment harder for two services (the service that owns the login subdomain and the service that owns the apex domain) to make changes.

It seems to be that step 7 could be modified as follows:

[...]

PR #821, opened 2026-03-25
Allow IDPs to delegate well-known file hosting via DNS TXT record

will-bartlett:
Adds a _fedcm DNS TXT record mechanism that allows IDPs to indicate which subdomain should host the well-known file, instead of requiring it at the registrable domain (apex). This addresses operational challenges for organizations where the apex domain is operated by a different service than the authentication service.

- Add 'resolve the well-known host' algorithm that queries _fedcm.<domain>
- Distinguish DNS 'no record' (fallback to apex) from DNS errors (hard fail)
- Extract well-known fetch into reusable 'fetch a well-known resource' algorithm
- Update well-known file documentation with DNS examples
- Add Deployment Considerations section explaining motivation

Addresses: https://github.com/w3c-fedid/FedCM/issues/809

***
<a href="https://pr-preview.s3.amazonaws.com/will-bartlett/FedCM/pull/821.html" title="Last updated on Apr 13, 2026, 9:34 PM UTC (6d6d42f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/821/b3383dc...will-bartlett:6d6d42f.html" title="Last updated on Apr 13, 2026, 9:34 PM UTC (6d6d42f)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -1148,6 +1148,96 @@ An extension may use the following instead of the [=create identity credential/s
 </div>
 </div>
 
+<!-- ============================================================ -->
+### Resolve the well-known host ### {#resolve-well-known-host}
+<!-- ============================================================ -->
+
+The <a>resolve the well-known host</a> algorithm determines the host from which the [=well-known
+file=] should be fetched. It does so by querying a DNS TXT record at a well-known subdomain, which
+allows [=IDPs=] to delegate [=well-known file=] hosting to a subdomain other than the
+[=host/registrable domain=].
+
+<div algorithm="resolve the well-known host">
+To <dfn>resolve the well-known host</dfn> given a [=host=] |registrableDomain|, run the following
+steps. This returns a [=host=] or failure.
+    1. Let |queryName| be the [=string/concatenation=] of «"_fedcm.", |registrableDomain|».
+    1. Let |result| be the result of resolving |queryName| as a DNS TXT record type.
+    1. If the resolution returned a response indicating that no records exist (e.g., NXDOMAIN or
+        an empty answer section), return |registrableDomain|.
+    1. If the resolution failed due to a network or DNS infrastructure error (e.g., SERVFAIL,
+        timeout, or connection failure), return failure.
+    1. Let |records| be |result|.
+    1. For each |record| in |records|:
+        1. Let |params| be the result of parsing |record|'s text content as a set of
+            key=value pairs separated by semicolons (";"), where keys and values are
+            stripped of leading and trailing [=ASCII whitespace=].
+        1. If |params| contains a key "sub" whose value is not empty:
+            1. Let |sub| be the value associated with "sub".
+            1. Return the [=string/concatenation=] of «|sub|, ".", |registrableDomain|».
+    1. Return |registrableDomain|.
+
+    Note: The DNS TXT record format follows the convention used by other DNS-based discovery
+        mechanisms (e.g., DMARC's `_dmarc` records). An [=IDP=] whose authentication service
+        runs on `login.idp.example` can set a TXT record at `_fedcm.idp.example` with value
+        `sub=login`, causing the user agent to fetch the [=well-known file=] from
+        `login.idp.example` instead of `idp.example`. If no `_fedcm` record exists, the
+        [=host/registrable domain=] is used directly. If DNS resolution fails due to a
+        network error, the algorithm returns failure rather than silently falling back,
+        to avoid intermittent behavior.
+
+</div>
+
+<!-- ============================================================ -->
+### Fetch a well-known resource ### {#fetch-well-known-resource}
+<!-- ============================================================ -->
+
+<div algorithm="fetch a well-known resource">
+To <dfn>fetch a well-known resource</dfn> given a [=/URL=] |url| and |globalObject|,
+run the following steps. This returns an {{IdentityProviderWellKnown}} or failure.
+    1. Let |result| be null.
+    1. Let |wellKnownRequest| be a new [=/request=] as follows:
+
+        :  [=request/URL=]
+        :: |url|
+        :  [=request/client=]
+        :: null
+        :  [=request/service-workers mode=]
+        :: "none"
+        :  [=request/destination=]
+        :: "webidentity"
+        :  [=request/origin=]
+        :: a unique [=opaque origin=]
+        :  [=request/header list=]
+        :: a [=list=] containing a single [=header=] with [=header/name=] set to `Accept` and
+            [=header/value=] set to `application/json`
+        :  [=request/referrer policy=]
+        :: "no-referrer"
+        :  [=request/credentials mode=]
+        :: "omit"
+        :  [=request/mode=]
+        :: "no-cors"
+
+        Issue: The spec is yet to be updated so that all <a spec=fetch for=/>requests</a> are created
+        with [=request/mode=] set to "user-agent-no-cors". See the relevant
+        [pull request](https://github.com/whatwg/fetch/pull/1533) for details.
+
+    1. [=Fetch request=] with |wellKnownRequest| and |globalObject|, and with <var ignore>processResponseConsumeBody</var>
+        set to the following steps, given a <a spec=fetch for=/>response</a> |response| and |responseBody|:
+        1. Let |json| be the result of [=extract the JSON fetch response=] from |response| and
+            |responseBody|.
+        1. Set |result| to the result of [=converted to an IDL value|converting=] |json|
+            to an {{IdentityProviderWellKnown}}.
+        1. If one of the previous two steps threw an exception, or if the
+            [=list/size=] of |result|["{{IdentityProviderWellKnown/provider_urls}}"] is
+            greater than 1, set |result| to failure.
+
+            Issue: [relax](https://github.com/fedidcg/FedCM/issues/333) the size of the
+            provider_urls array.
+
+    1. Wait for |result| to be set.
+    1. Return |result|.
+
+</div>
 
 <!-- ============================================================ -->
 ### Fetch the config file ### {#fetch-config-file}
@@ -1166,58 +1256,26 @@ or failure.
     1. Run a [[!CSP]] check with a [[CSP#directive-connect-src|connect-src]] directive on the URL
         passed as |configUrl|. If it fails, return failure.
     1. If |configUrl| is not a [=potentially trustworthy URL=], return failure.
+    1. Let |registrableDomain| be |configUrl|'s [=url/host=]'s [=host/registrable domain=].
+    1. Let |wellKnownHost| be the result of [=resolve the well-known host=] given
+        |registrableDomain|.
+    1. If |wellKnownHost| is failure, return failure.
     1. Let |rootUrl| be a new [=/URL=].
     1. Set |rootUrl|'s [=url/scheme=] to |configUrl|'s [=url/scheme=].
-    1. Set |rootUrl|'s [=url/host=] to |configUrl|'s [=url/host=]'s [=host/registrable domain=].
+    1. Set |rootUrl|'s [=url/host=] to |wellKnownHost|.
     1. Set |rootUrl|'s [=url/path=] to the <a>list</a> «".well-known", "web-identity"».
     1. Let |config|, |wellKnown|, |accounts_url|, and |login_url| be null.
     1. Let |skipWellKnown| be false.
     1. Let |rpOrigin| be |globalObject|'s [=associated Document=]'s [=Document/origin=].
-    1. If |rpOrigin| is not an [=opaque origin=], and |rootUrl|'s [=url/host=] is equal
-        to |rpOrigin|'s [=host/registrable domain=], and |rootUrl|'s [=url/scheme=] is
+    1. If |rpOrigin| is not an [=opaque origin=], and |registrableDomain| is equal
+        to |rpOrigin|'s [=host/registrable domain=], and |configUrl|'s [=url/scheme=] is
         equal to |rpOrigin|'s [=origin/scheme=], set |skipWellKnown| to true.
 
         Note: Because domain cookies are valid across an entire site, there is no privacy
             benefit from doing the well-known check if the RP and IDP are in the same site.
     1. Otherwise:
-        1. Let |wellKnownRequest| be a new [=/request=] as follows:
-
-            :  [=request/URL=]
-            :: |rootUrl|
-            :  [=request/client=]
-            :: null
-            :  [=request/service-workers mode=]
-            :: "none"
-            :  [=request/destination=]
-            :: "webidentity"
-            :  [=request/origin=]
-            :: a unique [=opaque origin=]
-            :  [=request/header list=]
-            :: a [=list=] containing a single [=header=] with [=header/name=] set to `Accept` and
-                [=header/value=] set to `application/json`
-            :  [=request/referrer policy=]
-            :: "no-referrer"
-            :  [=request/credentials mode=]
-            :: "omit"
-            :  [=request/mode=]
-            :: "no-cors"
-
-            Issue: The spec is yet to be updated so that all <a spec=fetch for=/>requests</a> are created
-            with [=request/mode=] set to "user-agent-no-cors". See the relevant
-            [pull request](https://github.com/whatwg/fetch/pull/1533) for details.
-
-        1. [=Fetch request=] with |wellKnownRequest| and |globalObject|, and with <var ignore>processResponseConsumeBody</var>
-            set to the following steps, given a <a spec=fetch for=/>response</a> |response| and |responseBody|:
-            1. Let |json| be the result of [=extract the JSON fetch response=] from |response| and
-                |responseBody|.
-            1. Set |wellKnown| to the result of [=converted to an IDL value|converting=] |json|
-                to an {{IdentityProviderWellKnown}}.
-            1. If one of the previous two steps threw an exception, or if the
-                [=list/size=] of |wellKnown|["{{IdentityProviderWellKnown/provider_urls}}"] is
-                greater than 1, set |wellKnown| to failure.
-
-                Issue: [relax](https://github.com/fedidcg/FedCM/issues/333) the size of the
-                provider_urls array.
+        1. Set |wellKnown| to the result of [=fetch a well-known resource=] given
+            |rootUrl| and |globalObject|.
 
     1. Let |configRequest| be a new <a spec=fetch for=/>request</a> as follows:
 
@@ -1292,11 +1350,12 @@ the [=fetch the config file/check accounts and login url step=]:
 
 NOTE: a two-tier file system is used in order to prevent the [=IDP=] from easily determining the [=RP=]
 that a user is visiting by encoding the information in the config file path. This issue is solved by
-requiring a [=well-known file=] to be on the root of the [=IDP=]. The config file itself can be anywhere, but
-it will not be used if the user agent does not find it in the [=well-known file=]. This allows the [=IDP=]
-to keep their actual config files on an arbitary path while allowing the user agent to prevent config file
-path manipulation to fingerprint (for instance, by including the RP in the path). See
-[[#manifest-fingerprinting]].
+requiring a [=well-known file=] to be on the [=IDP=]'s [=host/registrable domain=], or on a
+subdomain indicated by a DNS TXT record at `_fedcm` under the [=host/registrable domain=]. The
+config file itself can be anywhere, but it will not be used if the user agent does not find it in
+the [=well-known file=]. This allows the [=IDP=] to keep their actual config files on an arbitary
+path while allowing the user agent to prevent config file path manipulation to fingerprint (for
+instance, by including the RP in the path). See [[#manifest-fingerprinting]].
 
 <xmp class="idl">
 dictionary IdentityProviderWellKnown {
@@ -2004,11 +2063,11 @@ The <dfn>in parallel steps for getUserInfo</dfn> given a {{Promise}} |promise| a
     1. For each |account| in |accountsList|:
         1. If |account|["{{IdentityProviderAccount/approved_clients}}"] is not empty and it does not
             [=list/contain=] |provider|'s {{IdentityProviderConfig/clientId}}, continue.
-        
+
             Note: this allows the [=IDP=] to override whether an account is a returning account.
             This could be useful for instance in cases where the user has disconnected the
             account out of band.
-        
+
         1. If |account| is [=eligible for auto reauthentication=] given |provider| and
             |globalObject|, set |hasAccountEligibleForAutoReauthentication| to true.
     1. If |hasAccountEligibleForAutoReauthentication| is false, [=queue to reject=] |promise| with
@@ -2018,11 +2077,11 @@ The <dfn>in parallel steps for getUserInfo</dfn> given a {{Promise}} |promise| a
         1. [=list/Append=] an {{IdentityUserInfo}} to |userInfoList| with the following values:
 
             :  {{IdentityUserInfo/email}}
-            :: |account|["{{IdentityProviderAccount/email}}"]   
+            :: |account|["{{IdentityProviderAccount/email}}"]
             :  {{IdentityUserInfo/name}}
-            :: |account|["{{IdentityProviderAccount/name}}"]   
+            :: |account|["{{IdentityProviderAccount/name}}"]
             :  {{IdentityUserInfo/givenName}}
-            :: |account|["{{IdentityProviderAccount/given_name}}"]   
+            :: |account|["{{IdentityProviderAccount/given_name}}"]
             :  {{IdentityUserInfo/picture}}
             :: |account|["{{IdentityProviderAccount/picture}}"]
     1. [=Queue a global task=] on the [=DOM manipulation task source=] given |globalObject| to
@@ -2074,7 +2133,7 @@ network requests performed:
 
 NOTE: The browser uses the [=well-known file=] to prevent [[#manifest-fingerprinting]].
 
-The [=IDP=] exposes a <dfn>well-known file</dfn> in a pre-defined location, specifically at the "web-identity" file at the [=IDPs=]'s path ".well-known".
+The [=IDP=] exposes a <dfn>well-known file</dfn> in a pre-defined location, specifically at the "web-identity" file at the path ".well-known". The host from which the [=well-known file=] is fetched is determined by the [=resolve the well-known host=] algorithm: the user agent first checks for a DNS TXT record at `_fedcm` under the [=IDP=]'s [=host/registrable domain=]. If a record is found with a `sub` parameter (e.g., `sub=login`), the [=well-known file=] is fetched from the indicated subdomain (e.g., `login.idp.example`). Otherwise, the [=well-known file=] is fetched from the [=host/registrable domain=] directly (e.g., `idp.example`).
 
 The [=well-known file=] is fetched in the [=fetch the config file=] algorithm:
 
@@ -2083,7 +2142,8 @@ The [=well-known file=] is fetched in the [=fetch the config file=] algorithm:
 (c) **without** revealing the [=RP=] in the <a http-header>Origin</a> or
     [[RFC9110#field.referer|Referer]] headers.
 
-For example:
+For example, if the [=IDP=]'s [=host/registrable domain=] is `idp.example` and no `_fedcm` DNS TXT
+record is present, the user agent fetches the [=well-known file=] from the [=host/registrable domain=]:
 
 <div class=example>
 ```http
@@ -2094,12 +2154,28 @@ Sec-Fetch-Dest: webidentity
 ```
 </div>
 
+Alternatively, if the DNS TXT record `_fedcm.idp.example` contains `sub=login`, the user agent
+fetches the [=well-known file=] from the indicated subdomain instead:
+
+<div class=example>
+```http
+GET /.well-known/web-identity HTTP/1.1
+Host: login.idp.example
+Accept: application/json
+Sec-Fetch-Dest: webidentity
+```
+</div>
+
+This allows [=IDPs=] whose [=host/registrable domain=] (apex domain) is operated by a different
+service to delegate [=well-known file=] hosting to a subdomain via a DNS record, without requiring
+changes to the apex domain's web server.
+
 The file is parsed expecting a {{IdentityProviderWellKnown}} JSON object.
 
 The {{IdentityProviderWellKnown}} JSON object has the following semantics:
 
 <dl dfn-type="dict-member" dfn-for="IdentityProviderWellKnown">
-    :   <dfn>provider_urls</dfn> 
+    :   <dfn>provider_urls</dfn>
     ::  A list containing exactly one URL pointing to [[#idp-api-config-file]].
     :   <dfn>accounts_endpoint</dfn>
     ::  A URL that points to the same location as the {{IdentityProviderAPIConfig/accounts_endpoint}} in [[#idp-api-config-file]]s.
@@ -2467,9 +2543,9 @@ Every {{IdentityAssertionResponse}} is expected to have members with the followi
 
 <dl dfn-type="dict-member" dfn-for="IdentityAssertionResponse">
     :   <dfn>token</dfn>
-    ::  The resulting token. The value is of type <a href="https://webidl.spec.whatwg.org/#idl-any">`any`</a>, allowing 
+    ::  The resulting token. The value is of type <a href="https://webidl.spec.whatwg.org/#idl-any">`any`</a>, allowing
         [=IDPs=] to return tokens in various formats (string, object, etc.).
-        See the examples at the end of [[#idp-api-id-assertion-endpoint]] for both string 
+        See the examples at the end of [[#idp-api-id-assertion-endpoint]] for both string
         and object token formats.
     :   <dfn>continue_on</dfn>
     ::  A URL that the user agent will open in a popup to finish the authentication process.
@@ -2951,9 +3027,9 @@ The design of FedCM relies on several key security assumptions:
 ### Trusted User Agent ### {#trusted-user-agent}
 <!-- ============================================================ -->
 
-The user agent (i.e., the browser) is assumed to be a trusted entity. It is expected to faithfully 
-enforce same-origin policies, execute CSP and CORS checks, and present a secure, non-forgeable UI 
-that users can trust, and to not contains or executes malicious third parties scripts. The browser 
+The user agent (i.e., the browser) is assumed to be a trusted entity. It is expected to faithfully
+enforce same-origin policies, execute CSP and CORS checks, and present a secure, non-forgeable UI
+that users can trust, and to not contains or executes malicious third parties scripts. The browser
 is responsible for mediating the flow and preventing unauthorized access to credentials.
 
 <!-- ============================================================ -->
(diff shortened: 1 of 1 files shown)


spec/index.bs
```diff
@@ -1148,6 +1148,96 @@ An extension may use the following instead of the [=create identity credential/s
...
+
+The <a>resolve the well-known host</a> algorithm determines the host from which the [=well-known
+file=] should be fetched. It does so by querying a DNS TXT record at a well-known subdomain, which
+allows [=IDPs=] to delegate [=well-known file=] hosting to a subdomain other than the
+[=host/registrable domain=].
+
+<div algorithm="resolve the well-known host">
+To <dfn>resolve the well-known host</dfn> given a [=host=] |registrableDomain|, run the following
+steps. This returns a [=host=] or failure.
+    1. Let |queryName| be the [=string/concatenation=] of «"_fedcm.", |registrableDomain|».
```
samuelgoto:
```suggestion
    1. Let |queryName| be the [=string/concatenation=] of «"_web-identity.", |registrableDomain|».
```

For consistency with the `.well-known/web-identity`  file?

samuelgoto: Clarification question: there can be only ONE `_web-identity.domain.com` per registerable domain? or were you hoping you'd be able to get `_web-identity.subdomain1.domain.com` and `_web-identity.subdomain2.domain.com`?

If I understand this algorithm and the definition of the `[=host/registrable domain=]` definition, it would seem to be it is the former (just ONE per registerable domain).

Do I understand that correctly?

spec/index.bs
```diff
@@ -1148,6 +1148,96 @@ An extension may use the following instead of the [=create identity credential/s
...
+        1. If |params| contains a key "sub" whose value is not empty:
+            1. Let |sub| be the value associated with "sub".
+            1. Return the [=string/concatenation=] of «|sub|, ".", |registrableDomain|».
+    1. Return |registrableDomain|.
+
+    Note: The DNS TXT record format follows the convention used by other DNS-based discovery
+        mechanisms (e.g., DMARC's `_dmarc` records). An [=IDP=] whose authentication service
+        runs on `login.idp.example` can set a TXT record at `_fedcm.idp.example` with value
+        `sub=login`, causing the user agent to fetch the [=well-known file=] from
+        `login.idp.example` instead of `idp.example`. If no `_fedcm` record exists, the
```
samuelgoto: Ah, interesting. Can you move this note to the top? It explains a bunch and I think it addresses all of my comments and concerns!!

spec/index.bs
```diff
@@ -1148,6 +1148,96 @@ An extension may use the following instead of the [=create identity credential/s
...
+    1. For each |record| in |records|:
+        1. Let |params| be the result of parsing |record|'s text content as a set of
+            key=value pairs separated by semicolons (";"), where keys and values are
+            stripped of leading and trailing [=ASCII whitespace=].
+        1. If |params| contains a key "sub" whose value is not empty:
+            1. Let |sub| be the value associated with "sub".
+            1. Return the [=string/concatenation=] of «|sub|, ".", |registrableDomain|».
+    1. Return |registrableDomain|.
+
+    Note: The DNS TXT record format follows the convention used by other DNS-based discovery
```
samuelgoto: I think this mechanism works, but have you considered having a `url` instead of a `sub` in the DNS TXT record? That way, the IdP could just use ANY url, rather than a specific .well-known file one.

On a related note, I'm wondering if we could use CNAME rather than TXT to just perform a redirect?

Like CNAME points _web-identity.domain.com to login.domain.com.

Sorry, just thinking out loud.

spec/index.bs
```diff
@@ -1166,58 +1256,26 @@ or failure.
     1. Run a [[!CSP]] check with a [[CSP#directive-connect-src|connect-src]] directive on the URL
         passed as |configUrl|. If it fails, return failure.
     1. If |configUrl| is not a [=potentially trustworthy URL=], return failure.
+    1. Let |registrableDomain| be |configUrl|'s [=url/host=]'s [=host/registrable domain=].
+    1. Let |wellKnownHost| be the result of [=resolve the well-known host=] given
```
samuelgoto: Wouldn't this be making the browser be making an extra DNS requests for every well-known fetch we make today?

Meaning, for all of the IdPs today that have no issue hosting a .well-known file in their APEX domain, do we want them to pay the price of *also* having support the extra (declaration or load of) DNS TXT record?

samuelgoto (commented):
Good stuff @will-bartlett ! Clarification questions!!!

spec/index.bs
will-bartlett: Yes, just one. My understanding is that having only one is a privacy requirement, detailed in the "manifest fingerprinting" section

samuelgoto: Yeah, that's a big privacy requirement, I was just wondering if you were proposing we relax that, but it doesn't seem to be the case. Thanks for the clarification, resolving!!

spec/index.bs
will-bartlett: I think CNAME typically causes another lookup. A/AAAA request for foo.example, gets CNAME result pointing to bar.example, causes immediate A/AAAA request for bar.example. CNAME is like an HTTP 302 - it's not a final result - it's a "look over here". Also the TXT pattern is common in other specs (DMARC, SVBC, others) - so I'd rather follow the well-trodden path.

Regarding `url` vs `sub` - no strong opinion. I think the rest of the spec makes assumptions about all the endpoints being hosted on the same domain, so I see limited value here - just changing path, not changing everything - and I guess I generally like .well-known as a pattern. So, no strong opinion, but slight preference for `sub` over `url`.

spec/index.bs
will-bartlett: If the IdP isn't hosted at the apex, it's the same.

Today:
 - A/AAAA for idp.foo.example
 - A/AAAA for foo.example

Tomorrow:
 - A/AAAA for idp.foo.example
 - TXT for foo.example

Assuming the TXT points to the idp subdomain - but I think that's the normal case.

If the IdP is hosted at the apex, it's an additional DNS lookup.

Today:
 - A/AAAA for idp.example

Tomorrow:
 - A/AAAA for idp.example
 - TXT for idp.example

I guess I'm not too worried about this. DNS is well-cached and fast, the two DNS requests occur in parallel, and most idps aren't hosted and so would already require two in status quo.

spec/index.bs
will-bartlett: tentatively resolving, but feel free to reopen if you disagree.

spec/index.bs
will-bartlett: Added a small note about NXDOMAIN being cached, so no repetitive lookups on the apex if the TXT record is missing.

spec/index.bs
```diff
@@ -2985,19 +3069,19 @@ allow an attacker to bypass the FedCM flow or extract sensitive data.
...
 
-The FedCM API supports flexible token formats, allowing [=IDPs=] to return tokens in various forms. 
-[=RPs=] must be prepared to handle different token structures as agreed upon with their [=IDP=] 
+The FedCM API supports flexible token formats, allowing [=IDPs=] to return tokens in various forms.
+[=RPs=] must be prepared to handle different token structures as agreed upon with their [=IDP=]
 partners. The token content remains opaque to the user agent.
 
 **Recommendations**
 
-1. [=RPs=] should implement robust token parsing logic that can handle the specific format agreed
```
samuelgoto, line 2994: Can you remove some of these "formatting" changes from this PR so that it is easier to review? There are many LOC here that aren't very substantive, but makes it hard to tell the difference between what's important and what's not ...

spec/index.bs
```diff
@@ -2074,7 +2141,7 @@ network requests performed:
 
 NOTE: The browser uses the [=well-known file=] to prevent [[#manifest-fingerprinting]].
 
-The [=IDP=] exposes a <dfn>well-known file</dfn> in a pre-defined location, specifically at the "web-identity" file at the [=IDPs=]'s path ".well-known".
+The [=IDP=] exposes a <dfn>well-known file</dfn> in a pre-defined location, specifically at the "web-identity" file at the path ".well-known". The host from which the [=well-known file=] is fetched is determined by the [=resolve the well-known host=] algorithm: the user agent first checks for a DNS TXT record at `_web-identity` under the [=IDP=]'s [=host/registrable domain=]. If a record is found with a `sub` parameter (e.g., `sub=login`), the [=well-known file=] is fetched from the indicated subdomain (e.g., `login.idp.example`). Otherwise, the [=well-known file=] is fetched from the [=host/registrable domain=] directly (e.g., `idp.example`).
```
samuelgoto: Like I mentioned earlier, I'm a bit worried about this part:

> the user agent first checks for a DNS TXT record at `_web-identity` under

Because that would create an extra DNS TXT record call to all of the existing FedCM requests (e.g. it would slow down all of the existing requests, because it inserts an extra one).

I'm concerned about the latency regression, but also a deployment requirement regression.

Can't we make this request happen **after** the browser gets a 404 on the .well-known file, as a fallback?

will-bartlett: Not sure about 404. In some deployments (microsoftonline.com) the apex won't resolve (NXDOMAIN) so technically no HTTP status code is present. In others (reddit.com) there is a generic placeholder page on unknown paths (which comes with an HTTP 200) or a redirect (HTTP 302) to a generic placeholder page hosted at a fixed URL.

I also don't want to pessimize the subdomain case, as that seems to be the more common case compared to the apex among the working group members.

I thought about making two requests in parallel and attempting to parse both as web identity metadata or parsing whichever one has the appropriate content type, which would probably handle all those cases. This could probably be made to work, and I can send out another PR with it if you think it's more likely to be accepted than a DNS-based solution.

> deployment requirement regression

Can you clarify, here? Websites are vanishingly unlikely to have _web-identity DNS records today, so I don't think there should be a compatibility problem. Sites can continue to use the apex by not setting an TXT record.

> latency regression

I think this is speculative and may be incorrect for IDPs hosted in subdomains. If TXT points to the same subdomain as the config file, the reduction in TLS handshakes from 2 to 1 may more than compensate for the additional link in the dependency chain. I'd be interested in reviewing data if we can find an IDP and Browser willing to run the experiment.

samuelgoto: > Not sure about 404.

It doesn't have to be a `404`. I'm suggesting we use the DNS route as a "fallback": if the browser fails to fetch the `.well-known` file, for any reason (including for a 404), it fallbacks to query DNS to see if there is **another** location that it can load from.

Does that make sense?

> deployment requirement regression
>> Can you clarify, here? 

I intended to say that it would put a burden on the DNS servers today if we sent an **extra** DNS request (to `_web-identity`) every time a FedCM request was made.

Currently, 3% of all chrome page loads make a FedCM request, so I just wanted to raise that this could be an extra load in the DNS infrastructure that we should probably take into account.

> latency regression
>> I think this is speculative 

The fact that we are making 1 extra request BEFORE we make any the requests that we make today isn't speculative: we should expect **some** latency regression, right? You could argue that it might be small, but we'd expect **some**, right?

> I'd be interested in reviewing data if we can find an IDP and Browser willing to run the experiment.

@pottis any chance you can try to put something together?

ThisIsMissEm: > It doesn't have to be a `404`. I'm suggesting we use the DNS route as a "fallback": if the browser fails to fetch the `.well-known` file, for any reason (including for a 404), it fallbacks to query DNS to see if there is **another** location that it can load from.

I would strongly recommend doing these in parallel, if possible.

samuelgoto: > I would strongly recommend doing these in parallel, if possible.

Yeah, doing these in parallel would get the best of both worlds, I think. 

The only drawback, I believe is the deployment requirement regression I mentioned here:

https://github.com/w3c-fedid/FedCM/pull/821#discussion_r3012351310

We would be making a **lot** of DNS requests to the IdP's DNS servers, and I'm not sure if that's a problem or not (maybe it is not?).

will-bartlett:
Ok. I've updated the PR now to have all the options and marked my preferences. All my preferences are fairly weak however, so I'm happy to align with whatever we conclude when we meet. 

> also gives you TLS encryption of the values

I wasn't sure how important this was, given that the values must match the values from the config file, which is received over TLS.

> we currently make FedCM requests for 3% of chrome page loads, which is quite a lot

Wow! That **is** quite a lot! 1) Maybe I don't have my facts right. 2) I was thinking about IDPs more than RPs - I expect that huge/complex IDPs adopt FedCM sooner than small ones, but also have a greater capacity to deal with revisions.

> I was actually hoping that we'd get rid of the configURL at some point and only use the .well-known files, because that makes them easier to be discovered (e.g. given a domain, you can figure out how to use it as an IdP).

I don't know how to do N=1 for Microsoft Entra. Microsoft lets tenants [brand](https://learn.microsoft.com/en-us/entra/fundamentals/how-to-customize-branding) the login screen, all within the single Entra domain (login.microsoftonline.com). What the spec says right now works for us, because we can have per-customer branding in config files (/customerid/fedcm/config.json) and common endpoints in well-known files.

[RFC8414](https://datatracker.ietf.org/doc/html/rfc8414) does an interesting thing with well-known, where it puts subpaths in the well known. It says that configuration files should be located at idp.example/.well-known/oauth-authorization-server/customerid/v2.0 in Microsoft's setup. FedCM could do something like this for config files - idp.example/.well-known/web-identity/customerid, i.e. switch from configUrl to configDomain + configPath, or require that configUrl's path start with .well-known/web-identity - but that doesn't really seem to solve your discovery goals, if configPath is not auto-discoverable.

hlflanagan:
Discussed on the [7 April 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-04-07-FedCM-notes.md#allow-idps-to-delegate-well-known-file-hosting-via-dns-txt-record-821) call. @hlflanagan will take the architectural question to the TAG. The WG/CG are strongly advised to review this PR and indicate their preferred option so we can resolve the .well-known question.

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
 </div>
 </div>
 
+<!-- ============================================================ -->
+### Well-known discovery mechanism ### {#well-known-discovery}
+<!-- ============================================================ -->
+
+Issue: This section presents options for how the user agent discovers the [=well-known file=]
```
npm1: Inappropriate use of 'Issue'

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+The following options describe alternative mechanisms for obtaining [=well-known file=] data
+from a source other than the [=host/registrable domain=]. Each algorithm takes a [=host=]
+|registrableDomain|, a |scheme|, and a |globalObject|, and returns an
+{{IdentityProviderWellKnown}} or failure.
+
+<!-- BEGIN OPTION 1A (PREFERRED) -->
+
+##### Option 1A: SVCB record with embedded data (Preferred) ##### {#option-1a}
+
+In this option, the user agent queries a DNS [[RFC9460]] SVCB record at
```
npm1: Exposing DNS record resolution (SVCB or TXT) directly in a web specification is a layering violation. There is no standard primitive in the WHATWG/W3C ecosystem for "resolve a DNS record." By specifying this, you are forcing the user agent to integrate a custom network-layer API into the web platform's processing model.

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+_web-identity.idp.example.  SVCB  1 login.idp.example. accounts-endpoint="/accounts" login-url="/login"
+```
+
+This produces `https://login.idp.example/accounts` and `https://login.idp.example/login`.
+
+<!-- END OPTION 1A -->
+
+<!-- BEGIN OPTION 1B -->
+
+##### Option 1B: SVCB record pointing to subdomain ##### {#option-1b}
```
npm1: ditto layering violation.

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+    resolvers, so repeated lookups are fast. However, SVCB records are not yet universally
+    supported by all DNS registrars. The specific SvcParamKeys would need to be registered
+    with IANA per [[RFC9460]].
+
+<div algorithm="fetch well-known via SVCB data">
+To <dfn>fetch well-known via SVCB data</dfn> given a [=host=] |registrableDomain|, a |scheme|,
+and |globalObject|, run the following steps. This returns an {{IdentityProviderWellKnown}} or
+failure.
+    1. Let |queryName| be the [=string/concatenation=] of «"_web-identity.", |registrableDomain|».
+    1. Let |result| be the result of resolving |queryName| as a DNS SVCB record type.
```
npm1: Handwavy. This should at the very least link to something? And the thing is, I'm not sure there's a reasonable thing that can be linked here. This would require something new in Infra or Fetch.

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+    fetch. It provides infrastructure benefits of SVCB (e.g., ALPN hints, ECH) but
+    requires both a DNS query and an HTTP round-trip. Like Option 1A, SVCB support
+    among DNS registrars is not yet universal.
+
+<div algorithm="fetch well-known via SVCB host">
+To <dfn>fetch well-known via SVCB host</dfn> given a [=host=] |registrableDomain|, a |scheme|,
+and |globalObject|, run the following steps. This returns an {{IdentityProviderWellKnown}} or
+failure.
+    1. Let |queryName| be the [=string/concatenation=] of «"_web-identity.", |registrableDomain|».
+    1. Let |result| be the result of resolving |queryName| as a DNS SVCB record type.
```
npm1: ditto handwavy

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+_web-identity.idp.example.  SVCB  1 login.idp.example.
+```
+
+The user agent fetches `https://login.idp.example/.well-known/web-identity`.
+
+<!-- END OPTION 1B -->
+
+<!-- BEGIN OPTION 1C -->
+
+##### Option 1C: Fixed `web-identity.` subdomain ##### {#option-1c}
```
npm1: This seems like the only option that does not introduce a massive layering violation, although it does force a rigid network topology

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+    1. If the [=list/size=] of |records| is not 1, return failure.
+    1. Let |record| be |records|[0].
+    1. Let |target| be |record|'s TargetName.
+    1. If |target| is empty or |target| is ".", set |target| to |registrableDomain|.
+    1. If |target|'s [=host/registrable domain=] is not equal to |registrableDomain|,
+        return failure.
+    1. Let |params| be |record|'s SvcParams.
+    1. If |params| contains both an `accounts-endpoint` key and a `login-url`
+        key, and neither value is empty:
+        1. Let |accountsUrl| be a new [=/URL=] with [=url/scheme=] set to |scheme|,
```
npm1: This should use the URL parser. Manually setting properties on a new URL object bypasses standard validation and normalization steps.

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+web-identity.idp.example.  CNAME  login.idp.example.
+```
+
+The user agent fetches `https://web-identity.idp.example/.well-known/web-identity`.
+
+<!-- END OPTION 1C -->
+
+<!-- BEGIN OPTION 1D -->
+
+##### Option 1D: TXT record pointing to subdomain ##### {#option-1d}
```
npm1: ditto layering violation

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+    conventions established by DMARC (`_dmarc`), DKIM (`_domainkey`), and MTA-STS
+    (`_mta-sts`). However, it requires both a DNS query and an HTTP round-trip, and
+    TXT records lack the structured semantics of SVCB.
+
+<div algorithm="fetch well-known via TXT host">
+To <dfn>fetch well-known via TXT host</dfn> given a [=host=] |registrableDomain|,
+a |scheme|, and |globalObject|, run the following steps. This returns an
+{{IdentityProviderWellKnown}} or failure.
+    1. Let |queryName| be the [=string/concatenation=] of «"_web-identity.", |registrableDomain|».
+    1. Let |result| be the result of resolving |queryName| as a DNS TXT record type.
```
npm1: ditto

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+    1. If the resolution returned a response indicating that no records exist (e.g., NXDOMAIN
+        or an empty answer section), return failure.
+    1. If the resolution failed due to a network or DNS infrastructure error (e.g., SERVFAIL,
+        timeout, or connection failure), return failure.
+    1. Let |records| be |result|.
+    1. For each |record| in |records|:
+        1. Let |params| be the result of parsing |record|'s text content as a set of
+            key=value pairs separated by semicolons (";"), where keys and values are
```
npm1: This should probably use something in infra, not try to parse here

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+    1. Return the result of [=fetch a well-known resource=] given |rootUrl| and
+        |globalObject|.
+
+</div>
+
+<!-- END OPTION 2B -->
+
+<!-- BEGIN OPTION 2C (PREFERRED) -->
+
+##### Option 2C: Both in parallel, first success wins (Preferred) ##### {#option-2c}
```
npm1: Race condition, not acceptable

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+    1. If |newResult| is not null and |newResult| is not failure, return |newResult|.
+    1. Return failure.
+
+</div>
+
+<!-- END OPTION 2C -->
+
+<!-- BEGIN OPTION 2D -->
+
+##### Option 2D: Both in parallel, results must match ##### {#option-2d}
```
npm1: Not backwards compatible

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+    1. Return the result of the chosen Dimension 1 algorithm given |registrableDomain|,
+        |scheme|, and |globalObject|.
+
+</div>
+
+<!-- END OPTION 2A -->
+
+<!-- BEGIN OPTION 2B -->
+
+##### Option 2B: New mechanism first, existing as fallback ##### {#option-2b}
```
npm1: I dont think this is necessarily backwards compatible, as the new mechanism may time out. Also, slows down all existing IDPs.

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+fetch at the [=host/registrable domain=] (the "existing method") with the new discovery
+mechanism from [[#discovery-mechanism-options]] (the "new method").
+
+In all options below, |rootUrl| is the [=/URL=] for the [=well-known file=] at the
+[=host/registrable domain=], and the "chosen Dimension 1 algorithm" refers to whichever
+option is selected from [[#discovery-mechanism-options]].
+
+<!-- BEGIN OPTION 2A -->
+
+##### Option 2A: Existing fetch first, new as fallback ##### {#option-2a}
```
npm1: Unclear if this would actually work reasonably well for IDPs that cannot modify the existing location of well-known. Eg if the current fetch would timeout for them, this would not work. It would only work if the current fetch would always quickly fail.

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+|registrableDomain|, a |scheme|, and |globalObject|, run the following steps. This returns an
+{{IdentityProviderWellKnown}} or failure.
+    1. Let |existingResult| and |newResult| be null.
+    1. [=In parallel=]:
+        1. Set |existingResult| to the result of [=fetch a well-known resource=] given
+            |rootUrl| and |globalObject|.
+    1. [=In parallel=]:
+        1. Set |newResult| to the result of the chosen Dimension 1 algorithm given
+            |registrableDomain|, |scheme|, and |globalObject|.
+    1. Wait until |existingResult| is set and is not failure, or |newResult| is set and is
```
npm1: This is convoluted. Typically it would be preferred to just post a task from the in parallel work once the expected work is done and the new step can be processed from main

spec/index.bs
```diff
@@ -1148,6 +1148,437 @@ An extension may use the following instead of the [=create identity credential/s
...
+        not failure, or both are set.
+    1. If |existingResult| is not null and |existingResult| is not failure, return
```
npm1: If existingResult succeeds first, what happens to the pending newResult network request? I think you are leaking fetches

npm1 (changes requested):

spec/index.bs
will-bartlett: In practice, the user agent has already done this integration to support RFC 9460 (SVCB / HTTPS RRs). It's a layering violation only on paper, not in practice. Still, I've included option 1C in case folks think this conceptual layering violation is unsuitable.

spec/index.bs
will-bartlett: Can you give any guidance as to what that new thing might look like? As I mention in another comment, this is not _really_ new for browsers, it's just new for browser specifications.

spec/index.bs
will-bartlett: Do we normally do memory management in web specs? :) It's disposed when the last reference to it goes out of scope.

spec/index.bs
will-bartlett: Why do you say that? It seems backward compatible to me.

spec/index.bs
will-bartlett: Why do you say that? Happy eyeballs is a strategic race condition in web specs already. This is the same - racing because both methods are equivalent

will-bartlett:
@npm1 - the ask from the most recent notes (4/7) is for participants to express their preferences for how to solve this problem. I understand from your comments that for the first decision point, you prefer option C, as it avoids a conceptual layering violation. However, you left negative feedback about all options for the second decision point. Can you clarify your preference of the available options, or suggest a new option? Alternatively, if you'd prefer no option, i.e. if you think this problem is not worth solving, you can share that as well.

spec/index.bs
will-bartlett: A time out is just a type of failure, so it's not clear to me why this would cause a compatibility problem. Can you explain?

Ultimately, we have to choose between:
- pessimizing the majority of IDPs (which don't own the apex)
- pessimizing IDPs that already integrated with FedCM / requiring updates from them to restore performance
- race condition

I feel that the pessimizing the majority is the easiest to rule out, and many folks expressed doubts about inconsistent behavior / race conditions - which sort of points to this option.

spec/index.bs
npm1: Oh no, this is very much an in practice layering violation too, as @cbiesinger found out when implementing a similar DNS query for EVP.

spec/index.bs
npm1: Sure, I can elaborate. I strongly oppose this option. Racing network requests is a known anti-pattern in web standardization. Spinning up redundant DNS resolutions, TCP connections, and TLS handshakes for every FedCM discovery wastes client resources, drains mobile batteries, and unnecessarily increases server load. The W3C TAG explicitly advises against API designs that consume excessive user agent resources or rely on non-deterministic outcomes. In addition to this, this would introduce a privacy issue because an IDP could actually respond differently to the two well-known endpoints, thus essentially providing a fingerprinting bit, which we've worked really hard to avoid for many years.

spec/index.bs
npm1: Not sure what your definition of backwards compatible is. To me, is means: if an IDP does not change anything, will this change break them? This change WOULD break them, unless they implemented, unnecessarily, the new endpoint.

spec/index.bs
npm1: Because timeout means it would take ~forever for even the well-known to resolve? Timeout is a deal breaker here.

spec/index.bs
npm1: You must introduce an AbortController/AbortSignal internal to the algorithm to cancel the losing request, otherwise, you are forcing the user agent to waste bandwidth and battery.

npm1:
> @npm1 - the ask from the most recent notes (4/7) is for participants to express their preferences for how to solve this problem. I understand from your comments that for the first decision point, you prefer option C, as it avoids a conceptual layering violation. However, you left negative feedback about all options for the second decision point. Can you clarify your preference of the available options, or suggest a new option? Alternatively, if you'd prefer no option, i.e. if you think this problem is not worth solving, you can share that as well.

The ask, both in the call and in the comment which I'm quoting below, is to read and review the PR and express preferences. My preference would be 1C and 2A (or fetching both new and existing at the same time but giving priority to the existing, which I think is not any of the options: 2E?), but subject to this actually resolving the use cases, which is why I left a comment there on why it might not. Also, you called me out as being an editor and not reviewing in a timeline fashion, so not sure why you are not happy about my review?

> Discussed on the [7 April 2026](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-04-07-FedCM-notes.md#allow-idps-to-delegate-well-known-file-hosting-via-dns-txt-record-821) call. @hlflanagan will take the architectural question to the TAG. The WG/CG are strongly advised to review this PR and indicate their preferred option so we can resolve the .well-known question.

spec/index.bs
will-bartlett: Can you please share details? These sort of allusions to unspecified problems are not as helpful as details

spec/index.bs
will-bartlett: Yes, that's what backwards compatibility means. This change is backwards compatible under that definition.

Can you explain why you think an IDP that does not change anything would be broken?

will-bartlett:
> The ask, both in the call and in the comment which I'm quoting below, is to read and review the PR and express preferences. My preference would be 1C and 2A

Thanks for clarifying. It wasn't clear from your other comments that you preferred 2A.

> Also, you called me out as being an editor and not reviewing in a timeline fashion, so not sure why you are not happy about my review?

I don't believe I called you out specifically in the WG call - that was certainly not my intention. I merely asked all participants to take a look.

I also don't think I expressed any unhappiness with your review - not sure what that it in reference to. To be clear, I appreciate you reviewing the PR. I only asked you to clarify which option you preferred for decision point 2.

spec/index.bs
will-bartlett: To expand - I'd expect an IDP that does not change anything to get an error from the new mechanism and a success from the old mechanism. This algorithm only requires the two results to match when both results are not error. With one error and one success, this algorithm chooses the successful result.

spec/index.bs
will-bartlett: All the options except racing have a small but unlikely possibility of a timeout on the path they don't implement. IDPs that implement the new path can get timeout on the old path and IDPs that implement the old path can get timeout on the new path.

So it comes back to the same choice:
- pessimize (and rarely have timeout problems) for IDPs that don't own apex
- pessimize (and rarely have timeout problems) for IDPs that already integrated until they make updates
- race conditions

will-bartlett:
It occurs to me that there's perhaps another option for compatibility, 2E, which is just to have the relying party specify.

```
      let token = await navigator.credentials.get({
        identity: {
          providers: [{
            configURL: "https://idp.example/manifest.json",
            clientId: "123",
            privacyMode: "apex-http"
          }]
        }
      });
```

```
      let token = await navigator.credentials.get({
        identity: {
          providers: [{
            configURL: "https://idp.example/manifest.json",
            clientId: "123",
            privacyMode: "apex-dns"
          }]
        }
      });
```

That avoids the timeout or race condition choice, but muddles the API with an additional parameter.

Maybe 2F is to put privacyMode in manifest.json.

It adds a bit of data - the IdP can categorize RPs into one of two buckets based on which mode they use. But I'm not sure how much of a problem that really is - two buckets doesn't seem to be a big deal.

will-bartlett:
A DNSOP participant (Ben Schwartz) has another suggestion (let's call it 2G) which is to specify FedCM to only send origin-scoped cookies (i.e. those with _Host) prefix. This could be on the manifest as a "privacyMode":"originScopedCookies". This would allow hosting the well-known file on the same origin rather than on the registrable domain. However, it is a whole new mode for sending cookies, which brings its own complexities.

spec/index.bs
npm1: Well that's why I tagged Christian, I don't know the details

spec/index.bs
npm1: Ah, I thought this would be fetching both and only accepting when both match. The proposed behavior is super strange IMO. And that would still require waiting for both results, potentially resulting in a timeout for an existing IDP

npm1:
> A DNSOP participant (Ben Schwartz) has another suggestion (let's call it 2G) which is to specify FedCM to only send origin-scoped cookies (i.e. those with _Host) prefix. This could be on the manifest as a "privacyMode":"originScopedCookies". This would allow hosting the well-known file on the same origin rather than on the registrable domain. However, it is a whole new mode for sending cookies, which brings its own complexities.

The manifest and well-known fetches are uncredentialed.

will-bartlett:
> > A DNSOP participant (Ben Schwartz) has another suggestion (let's call it 2G) which is to specify FedCM to only send origin-scoped cookies (i.e. those with _Host) prefix. This could be on the manifest as a "privacyMode":"originScopedCookies". This would allow hosting the well-known file on the same origin rather than on the registrable domain. However, it is a whole new mode for sending cookies, which brings its own complexities.
> 
> The manifest and well-known fetches are uncredentialed.

I know. I mean that the manifest could indicate that the accountUrl and loginUrl fetches only send origin scoped cookies (cookies whose 'host-only' flag is true, in the language of [6252bis](https://www.ietf.org/archive/id/draft-ietf-httpbis-rfc6265bis-22.html#name-retrieval-model)). But I'm thinking this is probably too complex. We'd have to update 6252bis to have multiple retrieval modes and then update fetch to have a new credential modes which correspond to the new retrieval modes. And I don't think we'll convince the 6525bis folks - cookies are hard enough without modes.

spec/index.bs
TallTed: Perhaps you might suggest a better word for the use.

bemasc:
[Here](https://mailarchive.ietf.org/arch/msg/dnsop/cms8ZUg-SRsk9Z94OrqkbrQl5GI/)'s that message.

I don't claim to understand the whole flow here, but I doubt that new DNS records are the right answer.

Regarding origin-scoped cookies, another option would be to disable cookies entirely and keep the login state in LocalStorage (etc.), which already has the desired origin-scoping behavior.

will-bartlett:
Ok. I've updated this PR substantially. It now no longer display options, but focuses exclusively on the DNS option. It also includes that the new result is preferred over the old (matching @samuelgoto 's preference) and a note that browsers can make them in parallel to preserve performance in the short term.

I will send out a second PR with the other option (HTTP to fixed subdomain). This is my preferred option at this point. The only substantial con is the layering issues with web specs depending on DNS, which seems to be mostly a theory issue rather than a practical one.

> Regarding origin-scoped cookies, another option would be to disable cookies entirely and keep the login state in LocalStorage (etc.), which already has the desired origin-scoping behavior.

Needs to be credentials under fetch's definition, which basically means cookies, mtls, or basic/digest auth.

samuelgoto:
> I will send out a second PR with the other option (HTTP to fixed subdomain). This is my preferred option at this point. 

Would that be Option (C) described [here](https://mailarchive.ietf.org/arch/msg/dnsop/aLACo0YpxJezsvlXZipp9aL0mFs/)? If so, yeah, that's my preferred option too.

```
Option C: Fixed subdomain — No DNS lookup; always fetch from web-identity.<registrable-domain>, which the IDP points (e.g., via CNAME) to their infrastructure.
```

If so, should we include that in this PR rather than a second one?

spec/index.bs
```diff
@@ -1148,13 +1148,49 @@ An extension may use the following instead of the [=create identity credential/s
 </div>
 </div>
 
+<!-- ============================================================ -->
+### Well-known discovery via DNS TXT hash ### {#well-known-discovery}
+<!-- ============================================================ -->
+
+As an alternative to hosting the [=well-known file=] at the [=host/registrable domain=],
+an [=IDP=] may publish a DNS TXT record at `_web-identity.<registrable domain>` containing
```
samuelgoto, line 1156: > an [=IDP=] may publish a DNS TXT record at `_web-identity.<registrable domain>` containing

can we use Option (C) here instead?

https://github.com/w3c-fedid/FedCM/pull/821#issuecomment-4239805368

will-bartlett:
Here's that other PR: https://github.com/w3c-fedid/FedCM/pull/823

will-bartlett:
> This is my preferred option at this point.

I meant "This [PR (that uses DNS)] is my preferred option" - but I'm ok with either.

hlflanagan:
Discussed during [21 April 2026 meeting](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-04-21-FedCM-notes.md#aob). WG/CG encouraged to review how WebAuthn handles this. Will continue discussion at the next call (5 May 2026).

hlflanagan:
Discussed during [5 May 2026 meeting](https://github.com/w3c-fedid/meetings/blob/main/2026/2026-05-05-FedCM-notes.md#allow-idps-to-delegate-well-known-file-hosting-via-dns-txt-record-821)

Issue #833, opened 2026-05-20
Remaining items to resolve in IDP Service Worker design


domfarolino:
I think we had a productive call today in https://www.w3.org/events/meetings/b2f840e0-5593-4164-908e-dfaf26e8b3b7/20260519T080000/, discussing #815, thanks for all that contributed. We resolved that neither a request's [destination](https://fetch.spec.whatwg.org/#concept-request-destination) nor its eventual `Sec-Fetch-Dest` need to be "preserved" by the IDP's service worker that intercepts FedCM's client-less requests that get modified (with something like DPoP) and sent to the IDP's servers.

Reviewing https://github.com/w3c-fedid/FedCM/blob/main/explorations/identity_handler.md after the fact, I think ["Approach C"](https://github.com/w3c-fedid/FedCM/blob/main/explorations/identity_handler.md#how-it-works) laid out in that explainer is on the right track, although I want to make a few corrections, and use it as a jumping off point to seed the next call's agenda.

The section [introducing "Approach C"](https://github.com/w3c-fedid/FedCM/blob/main/explorations/identity_handler.md#approach-c--dedicated-event-modeled-on-payment-handler-current-direction) describes a new event type `IdentityRequestEvent` that is *not* a `FetchEvent`, somewhat [following the Payment Handler API spec](https://w3c.github.io/web-based-payment-handler/#handling-a-paymentrequestevent). It says:

> Being a different event entirely sidesteps the "is this foreign fetch in disguise?" critique against reusing FetchEvent.

While we may want a new event type, I think this statement is probably not accurate. I'll admit that foreign fetch was before my time, but I think the concern was with double-key cache privacy causing lack of utility, since client-having requests couldn't naturally invoke cross-origin service workers in a double-key-cached world, and bypassing that would be bad. We bypass that here not by merely using a non-`FetchEvent` type, but by requesting things from the UA in a special blessed FedCM flow, *not* with the RP's [client](https://fetch.spec.whatwg.org/#concept-request-client), and manually selecting the service worker to invoke. So I suspect the real reason that we "sidestep the 'is this foreign fetch in disguise?' critique" is the fact that our request is sent by a trusted path in the browser, not accessible to ordinary web content/scripts, so we're not providing a generic double-key-cache bypass for ordinary web requests like foreign fetch might've if we didn't rip it out.

# Event type

[...]

PR #815, opened 2026-02-28
Enabling IDP Interception in FedCM Request

pottis:
The current FedCM specification defines a **browser-mediated** flow where the browser makes **direct HTTP requests** to IDP endpoints. This PR implements a foundational capability allowing Identity Providers (IDPs) to intercept and handle FedCM (Federated Credential Management) requests through Service Workers. This feature unlocks advanced security and operational patterns that were previously impossible in the FedCM ecosystem.

Having this feature, it will address some of the key-issue from enterprise identity providers.
**Key issues:** 

- **Multi-domain failover**: Route to backup IDP when primary fails
- **Geographic/sovereign routing**: Send requests to jurisdiction-specific servers
- **Token caching during outages**: Graceful degradation when IDP temporarily unavailable
- **DPoP (Request Signing)**: Prevent token theft with cryptographic proof-of-possession
- **Legacy system integration**: Wrap existing identity services with FedCM interface

Links:
Issue #80

Explainer : https://github.com/pottis/FedCM/blob/4499cd7593e8f39b588ff3a6cca90e7ffd68beb8/explorations/identity_handler.md

***
<a href="https://pr-preview.s3.amazonaws.com/pottis/FedCM/pull/815.html" title="Last updated on May 15, 2026, 3:23 PM UTC (b03568e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/815/b3383dc...pottis:b03568e.html" title="Last updated on May 15, 2026, 3:23 PM UTC (b03568e)">Diff</a>

PR #842, opened 2026-07-28
Add initial Identity Handler SW Registration Tie In.

Brandr0id:
This change starts to update and split the proposed changes in #815. Specifically it focuses on the SW registration aspects and adds the `identity_handler` declaration to the well-known file and the registration lifecycle for it: register on config fetch, unregister when the declaration goes away, and clear alongside the rest of the IDP's storage. Registration is UA managed under a storage key isolated from the IDP's first-party one, which is where #833 looks to have landed on isolation.

Dispatch, the event IDL, and the security and privacy considerations will follow in separate PRs.

For additional context the full proposal [explainer](https://github.com/w3c-fedid/identity-handler/blob/main/identity-handler-explainer.md) may also help.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/pull/842.html" title="Last updated on Aug 1, 2026, 2:01 AM UTC (c7360cb)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c-fedid/FedCM/842/26a12eb...c7360cb.html" title="Last updated on Aug 1, 2026, 2:01 AM UTC (c7360cb)">Diff</a>

### DIFF
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -59,6 +59,13 @@ spec: service-workers; urlPrefix: https://w3c.github.io/ServiceWorker/
         text: register; url: register
         text: unregister; url: unregister
         text: soft update; url: soft-update
+        text: get registration; url: get-registration
+        text: create job; url: create-job
+        text: schedule job; url: schedule-job
+        text: registration map; url: dfn-scope-to-registration-map
+    type: dfn; for: job
+        text: update via cache mode; url: dfn-job-update-via-cache-mode
+        text: referrer; url: job-referrer
 </pre>
 
 <pre class=link-defaults>
@@ -1306,8 +1313,8 @@ or failure.
             |wellKnown|.{{IdentityProviderWellKnown/provider_urls}}[0], and |globalObject|.
         1. If |allowed_config_url| is not [=url/equal=] to |configUrl|, return failure.
     1. If |wellKnown|'s {{IdentityProviderWellKnown/identity_handler}} is present, then
-        [=in parallel=], [=register the identity handler=] given |configUrl|'s [=url/origin=] and
-        |wellKnown|'s {{IdentityProviderWellKnown/identity_handler}}.
+        [=in parallel=], [=register the identity handler=] given |configUrl| and |wellKnown|'s
+        {{IdentityProviderWellKnown/identity_handler}}.
     1. Otherwise, [=in parallel=], [=unregister the identity handler=] given |configUrl|'s
         [=url/origin=].
 
@@ -1316,6 +1323,11 @@ or failure.
         |configUrl| (see [=computing the manifest URL=]), so keying the handler this way guarantees
         its registration scope can cover the endpoints it intercepts, regardless of where the
         [=well-known file=] is hosted.
+
+        Note: These steps are only reached when the [=well-known file=] was fetched successfully, so
+        a network error or an unparsable [=well-known file=] leaves an existing registration
+        unchanged. An [=IDP=] disables interception by serving a [=well-known file=] without an
+        {{IdentityProviderWellKnown/identity_handler}} member, not by making the file unavailable.
     1. Return |config|.
 
 </div>
@@ -1366,10 +1378,10 @@ Each [=/origin=] has an associated <dfn>FedCM identity-handler storage key</dfn>
 [=storage key=] the [=user agent=] constructs for that [=/origin=]'s FedCM identity handler
 [[!STORAGE]]. It MUST be distinct from the [=/origin=]'s first-party [=storage key=], such that:
 
-1. it can only be produced by the [=user agent=]; page script MUST NOT be able to cause a
+1. no Web API surface accepts this [=storage key=] or otherwise allows page script to cause a
     [=service worker registration=] to be stored under it; and
-1. it is derivable from the [=/origin=] alone, so that a later FedCM flow can recompute it to look
-    up or [=unregister=] the [=/service worker=] without any persisted state.
+1. it is derivable from the [=/origin=] alone by the [=user agent=], so that a later FedCM flow can
+    recompute it to look up or [=unregister=] the [=/service worker=] without any persisted state.
 
 Note: This isolation ensures FedCM registrations neither collide with the [=IDP=]'s ordinary
 first-party [=/service workers=] nor are reachable or removable by unrelated code sharing the
@@ -1382,34 +1394,69 @@ worker match path used for client-controlled fetches. Because a credentialed Fed
 controlling client, this ensures the request reaches only a [=/service worker=] the [=user agent=]
 registered for that specific [=IDP=].
 
+The [=user agent=] MUST serialize [=register the identity handler=] and
+[=unregister the identity handler=] per [=/origin=]: at most one of them runs at a time for a given
+[=/origin=], and their effects are observable in the order they were dispatched.
+
+Note: Concurrent FedCM requests for the same [=IDP=] can otherwise observe different
+[=well-known files=] from the HTTP cache and race to opposite conclusions about whether an identity
+handler is declared.
+
 <div algorithm>
-To <dfn>register the identity handler</dfn> given an [=/origin=] |idpOrigin| and an
+To <dfn>register the identity handler</dfn> given a [=/URL=] |configUrl| and an
 {{IdentityHandler}} |handler|, run these steps:
 
     1. Assert: These steps are running [=in parallel=].
+    1. Let |idpOrigin| be |configUrl|'s [=url/origin=].
     1. Let |scriptURL| be the result of running the [=URL parser=] on |handler|'s
-        {{IdentityHandler/service_worker}} with |idpOrigin| as the base URL.
+        {{IdentityHandler/service_worker}} with |configUrl| as the base URL.
     1. If |scriptURL| is failure, return.
     1. If |scriptURL|'s [=url/origin=] is not [=same origin=] with |idpOrigin|, return.
 
-        Note: The declared script MUST be [=same origin=] with the [=config file=]. A
-        cross-origin declaration is ignored and credentialed requests fall back to the network.
-    1. Let |scope| be the [=string=] formed by |scriptURL| with its last path segment removed.
+        Note: A cross-origin declaration is ignored, and credentialed requests fall back to the
+        network. The [=same origin=] requirement is stated normatively on
+        {{IdentityHandler/service_worker}} above.
+    1. Let |scope| be the result of running the [=URL parser=] on "`./`" with |scriptURL| as the
+        base URL.
     1. Let |storageKey| be the [=FedCM identity-handler storage key=] for |idpOrigin|.
-    1. Let |registration| be the result of running <a spec=service-workers>Match Service Worker
-        Registration</a> given |storageKey| and |scope| [[!SERVICE-WORKERS]].
-    1. If |registration| is not null and its [=active worker=]'s script url is |scriptURL|:
+    1. Let |registration| be the result of running [=get registration=] given |storageKey| and
+        |scope| [[!SERVICE-WORKERS]].
+
+        Note: [=Get registration=] matches |scope| exactly. The [=user agent=] derived that
+        [=/scope=] itself, so it does not need the longest-prefix matching that
+        [=match service worker registration=] performs on behalf of a [=/service worker client=].
+    1. If |registration| is not null, |registration|'s [=active worker=] is not null, and
+        |registration|'s [=active worker=]'s [=service worker/script url=] [=url/equals=]
+        |scriptURL|, then:
         1. [=Soft update=] |registration|.
 
-            Note: The currently [=active worker=] handles the request that triggered this
-            invocation while the [=user agent=] checks for a newer version in parallel.
+            Note: The [=user agent=] proceeds with the [=active worker=] that is already
+            registered while it checks for a newer version in parallel, so update discovery never
+            delays a FedCM request.
         1. Return.
-    1. [=Register=] a [=/service worker=] with script URL |scriptURL|, scope |scope|,
-        and [=storage key=] |storageKey|, with its update-via-cache mode set to "`all`", and wait
-        until it has an [=active worker=].
+    1. [=Unregister the identity handler=] given |idpOrigin|.
+
+        Note: This clears a registration left over from an earlier declaration, including one whose
+        script never activated and one whose [=/scope=] no longer matches because the [=IDP=] moved
+        the script. At most one [=service worker registration=] exists under |storageKey|.
+    1. Let |job| be the result of running [=create job=] with *register*, |storageKey|, |scope|,
+        |scriptURL|, null, and null.
+    1. Set |job|'s [=job/update via cache mode=] to "`all`".
 
         Note: Setting update-via-cache to "`all`" lets the [=IDP=] control update discovery through
         the script's `Cache-Control`/`ETag` headers.
+    1. Set |job|'s [=job/referrer=] to |scriptURL|.
+
+        Note: [=Register=] compares |job|'s script URL and [=/scope=] against |job|'s
+        [=job/referrer=]'s [=/origin=], which [=create job=] otherwise derives from a
+        [=/service worker client=]. A FedCM registration has no client, so the [=user agent=] sets
+        the [=job/referrer=] to the declared script URL, which is [=same origin=] with both by
+        construction.
+    1. Invoke [=schedule job=] with |job|.
+
+        Note: The job has no [=/service worker client=] and no promise, so a script that fails to
+        fetch or install is not reported to the [=IDP=]. Credentialed requests fall back to the
+        network until a later FedCM flow retries the registration.
 </div>
 
 <div algorithm>
@@ -1417,7 +1464,16 @@ To <dfn>unregister the identity handler</dfn> given an [=/origin=] |idpOrigin|, 
 
     1. Assert: These steps are running [=in parallel=].
     1. Let |storageKey| be the [=FedCM identity-handler storage key=] for |idpOrigin|.
-    1. [=Unregister=] every [=service worker registration=] whose [=storage key=] is |storageKey|.
+    1. [=list/For each=] (|entryStorageKey|, |entryScope|) of the [=registration map=]'s
+        [=map/keys=]:
+        1. If |entryStorageKey| is not |storageKey|, [=iteration/continue=].
+        1. Let |job| be the result of running [=create job=] with *unregister*, |entryStorageKey|,
+            |entryScope|, null, null, and null.
+        1. Invoke [=schedule job=] with |job|.
+
+    Note: The [=user agent=] iterates the [=registration map=] rather than [=unregister=]ing one
+    known [=/scope=], because the {{IdentityProviderWellKnown/identity_handler}} declaration, and
+    with it the declared [=/scope=], may already be gone by the time these steps run.
 </div>
 
 The [=user agent=] [=unregister the identity handler|unregisters the identity handler=] when the
@@ -1429,9 +1485,10 @@ lifetime bounds how quickly adding or removing the {{IdentityProviderWellKnown/i
 member takes effect. [=IDPs=] are encouraged to serve it with a short `Cache-Control` `max-age` so
 that enabling or disabling the identity handler propagates promptly.
 
-The [=FedCM identity-handler storage key=] is part of the [=IDP=]'s [=/origin=]'s storage: when the
-[=IDP=]'s storage is cleared (for example, via [[CLEAR-SITE-DATA]]), the [=user agent=] MUST also
-remove [=service worker registration|registrations=] stored under it, so no FedCM-specific clearing
+The [=/origin=]'s storage, for the purposes of clearing storage [[!STORAGE]] and of any
+[[CLEAR-SITE-DATA]] operation targeting that [=/origin=], includes any
+[=service worker registration|registrations=] stored under the
+[=FedCM identity-handler storage key=] for that [=/origin=], so no FedCM-specific clearing
 mechanism is required.
 
 <xmp class="idl">
--- a/spec/index.bs
+++ b/spec/index.bs
@@ -54,6 +54,11 @@ spec: login-status; urlPrefix: https://w3c-fedid.github.io/login-status
         text: unknown; url: unknown
         text: get the login status; url: get-the-login-status
         text: set the login status; url: set-the-login-status
+spec: service-workers; urlPrefix: https://w3c.github.io/ServiceWorker/
+    type: dfn
+        text: register; url: register
+        text: unregister; url: unregister
+        text: soft update; url: soft-update
 </pre>
 
 <pre class=link-defaults>
@@ -69,6 +74,10 @@ spec:fetch; type:dfn; for:/; text:response
 spec:css-color-5; type:type; text:<color>
 spec:html; type:dfn; text:allowed to use
 spec:html; type:dfn; for:/; text:same site
+spec:service-workers; type:dfn; for:/; text:service worker
+spec:service-workers; type:dfn; for:/; text:service workers
+spec:service-workers; type:dfn; for:/; text:service worker registration
+spec:storage; type:dfn; text:storage key
 </pre>
 
 <style>
@@ -1296,6 +1305,17 @@ or failure.
         1. Let |allowed_config_url| be the result of [=computing the manifest URL=] with |provider|,
             |wellKnown|.{{IdentityProviderWellKnown/provider_urls}}[0], and |globalObject|.
         1. If |allowed_config_url| is not [=url/equal=] to |configUrl|, return failure.
+    1. If |wellKnown|'s {{IdentityProviderWellKnown/identity_handler}} is present, then
+        [=in parallel=], [=register the identity handler=] given |configUrl|'s [=url/origin=] and
+        |wellKnown|'s {{IdentityProviderWellKnown/identity_handler}}.
+    1. Otherwise, [=in parallel=], [=unregister the identity handler=] given |configUrl|'s
+        [=url/origin=].
+
+        Note: The identity handler is keyed to |configUrl|'s [=/origin=], not to the [=/origin=] the
+        [=well-known file=] was fetched from. Every credentialed endpoint is [=same origin=] with
+        |configUrl| (see [=computing the manifest URL=]), so keying the handler this way guarantees
+        its registration scope can cover the endpoints it intercepts, regardless of where the
+        [=well-known file=] is hosted.
     1. Return |config|.
 
 </div>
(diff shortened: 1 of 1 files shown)


Brandr0id:
@cbiesinger @domfarolino @wanderview @TallTed @yoshisatoyanagisawa @monica-ch
FYI as other reviewers on #815

spec/index.bs
```diff
@@ -1316,11 +1336,114 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+    1. If |scriptURL| is failure, return.
+    1. If |scriptURL|'s [=url/origin=] is not [=same origin=] with |idpOrigin|, return.
+
+        Note: The declared script MUST be [=same origin=] with the [=config file=]. A
+        cross-origin declaration is ignored and credentialed requests fall back to the network.
+    1. Let |scope| be the [=string=] formed by |scriptURL| with its last path segment removed.
+    1. Let |storageKey| be the [=FedCM identity-handler storage key=] for |idpOrigin|.
+    1. Let |registration| be the result of running <a spec=service-workers>Match Service Worker
+        Registration</a> given |storageKey| and |scope| [[!SERVICE-WORKERS]].
+    1. If |registration| is not null and its [=active worker=]'s script url is |scriptURL|:
```
monica-ch: if |registration| is not null and its [=active worker=]'s script url is |scriptURL| — if a prior Register attempt failed to activate (script parse error, install-event rejection), Match Service Worker Registration can still return a registration whose [=active worker=] is null, so accessing .script url deref-fails. That silently pushes us to step 8 and we Register again on every FedCM invocation — an infinite one-attempt-per-call retry loop with no backoff.

> If |registration| is not null and its [=active worker=] is not null and its [=active worker=]'s script url is |scriptURL|:

spec/index.bs
```diff
@@ -1316,11 +1336,114 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+
+Note: There is no separate "enabled" switch. An [=IDP=] disables interception by removing the
+{{IdentityProviderWellKnown/identity_handler}} member from its [=well-known file=], which causes the
+[=user agent=] to [=unregister the identity handler=] and use the direct-network path.
+
+Each [=/origin=] has an associated <dfn>FedCM identity-handler storage key</dfn>: a
+[=storage key=] the [=user agent=] constructs for that [=/origin=]'s FedCM identity handler
+[[!STORAGE]]. It MUST be distinct from the [=/origin=]'s first-party [=storage key=], such that:
+
+1. it can only be produced by the [=user agent=]; page script MUST NOT be able to cause a
```
monica-ch: The two MUST clauses look internally inconsistent:
> 1. "it can only be produced by the [=user agent=]…"
> 2. "it is derivable from the [=/origin=] alone, so that a later FedCM flow can recompute it…"

Page script has the origin, so if the key is deterministically derivable from origin alone, the "produced only by UA" property can't be about unguessability. The property that actually holds — verified against Chromium's blink::StorageKey::CreateWithNonce (browser-process-only) — is an API-surface constraint: no Web API accepts this key or lets a client cause a registration to be stored under it.

Suggestion:

> 1. no Web API surface accepts this [=storage key=] or otherwise allows page script to cause a [=service worker registration=] to be stored under it; and
> 2. it is derivable from the [=/origin=] alone by the [=user agent=], so that a later FedCM flow can recompute it without any persisted state.

spec/index.bs
```diff
@@ -1316,11 +1336,114 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+The [=user agent=] [=unregister the identity handler|unregisters the identity handler=] when the
+[=IDP=] removes the {{IdentityProviderWellKnown/identity_handler}} member from its
+[=well-known file=].
+
+Note: Because the [=well-known file=] is fetched subject to ordinary HTTP caching, its freshness
+lifetime bounds how quickly adding or removing the {{IdentityProviderWellKnown/identity_handler}}
+member takes effect. [=IDPs=] are encouraged to serve it with a short `Cache-Control` `max-age` so
+that enabling or disabling the identity handler propagates promptly.
+
+The [=FedCM identity-handler storage key=] is part of the [=IDP=]'s [=/origin=]'s storage: when the
```
monica-ch: The MUST here is right in spirit, but the mechanics don't connect through the Storage spec. Clear-Site-Data [[CLEAR-SITE-DATA]] delegates to Storage spec's clear the storage over the origin's [=storage key=], and this section defines the FedCM key as distinct from the first-party [=storage key=] — so a plain implementation of clear the storage iterates the first-party key only and never touches FedCM registrations.

Suggestion is to make the inclusion structural rather than a floating MUST:

> The [=/origin=]'s storage, for the purposes of clear the storage [[STORAGE]] and any [[CLEAR-SITE-DATA]] operation targeting this [=/origin=], includes any [=service worker registration|registrations=] stored under the [=FedCM identity-handler storage key=] for this [=/origin=].

That way it flows through by definition instead of relying on every CLEAR-SITE-DATA implementer discovering this section.

spec/index.bs
```diff
@@ -1296,6 +1305,17 @@ or failure.
...
+        |wellKnown|'s {{IdentityProviderWellKnown/identity_handler}}.
+    1. Otherwise, [=in parallel=], [=unregister the identity handler=] given |configUrl|'s
+        [=url/origin=].
+
+        Note: The identity handler is keyed to |configUrl|'s [=/origin=], not to the [=/origin=] the
+        [=well-known file=] was fetched from. Every credentialed endpoint is [=same origin=] with
+        |configUrl| (see [=computing the manifest URL=]), so keying the handler this way guarantees
+        its registration scope can cover the endpoints it intercepts, regardless of where the
+        [=well-known file=] is hosted.
     1. Return |config|.
```
monica-ch, line 1340: IdentityProviderWellKnown.provider_urls is a list and can hold URLs on distinct [=/origins=] under the same eTLD+1. Because registration is keyed to |configUrl|'s origin, each origin gets its own [=FedCM identity-handler storage key=] and its own independent registration — all sourced from the same well-known identity_handler.service_worker value.

> Note: If a [=well-known file=] lists multiple [=/origins=] in {{IdentityProviderWellKnown/provider_urls}}, each [=/origin=] has its own [=FedCM identity-handler storage key=] and its own registration, even though they share a single {{IdentityProviderWellKnown/identity_handler}} declaration.

spec/index.bs
monica-ch, line 1303: The register/unregister branches at lines 1308–1312 run only if the well-known fetch succeeds — the algorithm returns failure earlier ("If |wellKnown| is failure, return failure" at line 1293) on network error or non-2xx. That means an IDP that pulls /.well-known/web-identity to disable the feature during a bad rollout does not get the handler unregistered until a subsequent successful fetch that omits identity_handler.

Two options — either is fine, but pick one and state it:

Explicit: "If the well-known fetch fails, any existing identity handler registration is unchanged." (matches SW soft-update semantics; deliberate)
Or: also call unregister the identity handler on well-known fetch failure (aggressive but risks flapping on transient errors)

spec/index.bs
```diff
@@ -1296,6 +1305,17 @@ or failure.
         1. Let |allowed_config_url| be the result of [=computing the manifest URL=] with |provider|,
             |wellKnown|.{{IdentityProviderWellKnown/provider_urls}}[0], and |globalObject|.
         1. If |allowed_config_url| is not [=url/equal=] to |configUrl|, return failure.
+    1. If |wellKnown|'s {{IdentityProviderWellKnown/identity_handler}} is present, then
```
monica-ch, line 1319: Two overlapping FedCM invocations for the same IDP (different tabs, close in time, one hits an HTTP-cached well-known with identity_handler present, the other with it absent) each schedule register / unregister [=in parallel=] with no coordination. Nothing in the algorithms serializes them per |idpOrigin|, so the terminal registration state depends on which parallel task's last mutation lands.

> Note: [=Register the identity handler=] and [=unregister the identity handler=] invocations for the same |idpOrigin| MUST be serialized: at most one runs at a time per [=/origin=], and their effects are observable in the order they were dispatched.

monica-ch (changes requested):

spec/index.bs
```diff
@@ -1316,11 +1336,114 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+    1. Let |storageKey| be the [=FedCM identity-handler storage key=] for |idpOrigin|.
+    1. Let |registration| be the result of running <a spec=service-workers>Match Service Worker
+        Registration</a> given |storageKey| and |scope| [[!SERVICE-WORKERS]].
+    1. If |registration| is not null and its [=active worker=]'s script url is |scriptURL|:
+        1. [=Soft update=] |registration|.
+
+            Note: The currently [=active worker=] handles the request that triggered this
+            invocation while the [=user agent=] checks for a newer version in parallel.
+        1. Return.
+    1. [=Register=] a [=/service worker=] with script URL |scriptURL|, scope |scope|,
```
monica-ch: (and 1420 for Unregister) [=Register=] per the SW spec takes a job as input, not (scriptURL, scope, storageKey, updateViaCache). Start Register requires a client, promise, and referrer — none exist during well-known processing. This was raised in #815 (comment on L1279) and hasn't been resolved here. Options:

1.. Define this as a spec-level primitive: "install a service worker registration in the scope-to-registration-map with storage key K, script URL S, scope C, update-via-cache 'all'", explicitly bypassing the job queue (analogous to how the Storage spec defines direct bottle-map operations).
2. Manufacture a synthetic job with a resolved promise and a synthetic client bound to idpOrigin, then invoke Register/Start Register.

Option 1 matches the "isolated from Web SW" intent better and is what Payment Handler effectively does via its manifest-driven registration.

spec/index.bs
monica-ch: The register algorithm's guard is active worker's script url is scriptURL. If the IDP publishes a new well-known with a different service_worker value (e.g. /sw-v2.js), the old registration at /sw-v1.js never gets torn down — it just falls off the "match" path. Result: dead registrations accumulate under the isolated storage key forever, invisible to devtools and to Clear-Site-Data unless the origin is fully cleared. The algorithm needs an explicit "unregister any prior registration under this storage key whose script URL differs" step before Register.

Brandr0id: Updated.

spec/index.bs
Brandr0id: Thanks; updated this.

spec/index.bs
Brandr0id: Opt'd to call out a synthetic job here.

spec/index.bs
Brandr0id: added an unregister step + note for this.

spec/index.bs
Brandr0id: Updated to try and make this clearer.

spec/index.bs
Brandr0id: Would fetching the well-known file end up in failure if provider_urls are > 1 currently due to :
>
1. If one of the previous two steps threw an exception, or if the
[=list/size=] of \|wellKnown\|["{{IdentityProviderWellKnown/provider_urls}}"] is
greater than 1, set \|wellKnown\| to failure.

monica-ch: You're right, I missed the size > 1 failure at L1249–1251. So this is moot on today's spec text.

spec/index.bs
```diff
@@ -1296,6 +1312,22 @@ or failure.
         1. Let |allowed_config_url| be the result of [=computing the manifest URL=] with |provider|,
             |wellKnown|.{{IdentityProviderWellKnown/provider_urls}}[0], and |globalObject|.
         1. If |allowed_config_url| is not [=url/equal=] to |configUrl|, return failure.
+    1. If |wellKnown|'s {{IdentityProviderWellKnown/identity_handler}} is present, then
+        [=in parallel=], [=register the identity handler=] given |configUrl| and |wellKnown|'s
```
npm1: What happens if this `navigator.credentials.get` call finishes quickly and then another one is kicked off right after? It's possible that the identity handler might not have finished registering yet. Should this new call wait on pending identity handler registration or proceed with the existing state? If it proceeds, we need to ensure the register method is idempotent.

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+            delays a FedCM request.
+        1. Return.
+    1. [=Unregister the identity handler=] given |idpOrigin|.
+
+        Note: This clears a registration left over from an earlier declaration, including one whose
+        script never activated and one whose [=/scope=] no longer matches because the [=IDP=] moved
+        the script. At most one [=service worker registration=] exists under |storageKey|.
+    1. Let |job| be the result of running [=create job=] with *register*, |storageKey|, |scope|,
+        |scriptURL|, null, and null.
+    1. Set |job|'s [=job/update via cache mode=] to "`all`".
```
monica-ch, line 1464: I traced the synthetic-job pattern through the SW algorithms. Two things:

Null promise + null client is safe on the promise paths. [Reject Job Promise](https://w3c.github.io/ServiceWorker/#reject-job-promise-algorithm) and [Resolve Job Promise](https://w3c.github.io/ServiceWorker/#resolve-job-promise-algorithm) both guard with "If job's client is not null" and skip null-client entries in the equivalent-jobs loop. So the Note at L1457–1459 is algorithmically supported by SW spec, not just intent.

Gap: job's worker type is never set. [Create Job](https://w3c.github.io/ServiceWorker/#create-job-algorithm) doesn't initialize it, but [Register](https://w3c.github.io/ServiceWorker/#register-algorithm) step 5 compares it against newestWorker's type, and [Update](https://w3c.github.io/ServiceWorker/#update-algorithm) branches on it ("classic" / "module"). Unset → undefined behavior.

Suggest adding after L1443: Set |job|'s worker type to "classic" — matches how Payment Handler / Background Fetch specify their internal registrations. If you want module support later, add a type member to IdentityHandler.

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
 path manipulation to fingerprint (for instance, by including the RP in the path). See
 [[#manifest-fingerprinting]].
 
+<!-- ============================================================ -->
+### Register the identity handler ### {#register-identity-handler}
+<!-- ============================================================ -->
+
+An [=IDP=] MAY declare a FedCM-specific [=/service worker=] in its [=well-known file=] using the
+{{IdentityProviderWellKnown/identity_handler}} member. When it is present, the [=user agent=]
+registers and manages this [=/service worker=] itself [[!SERVICE-WORKERS]] and dispatches
```
npm1: The `[[!SERVICE-WORKERS]]` mention seems out of place?

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+credentialed FedCM requests to it.
+The [=IDP=] page does not register the [=/service worker=]: because credentialed FedCM requests can
+happen long after the [=IDP=] page was last visited, the [=user agent=] is responsible for ensuring
+the declared [=/service worker=] is registered and has an [=active worker=] when it is needed.
```
npm1: Not a sw expert but do we need to mention how long the sw is registered for before it becomes inactive and needs to be re-registered? Or is that all automatically handled in sw spec land?

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+
+The {{IdentityHandler}} members have the following semantics:
+
+<dl dfn-type="dict-member" dfn-for="IdentityHandler">
```
npm1: These seem like non-normative descriptions, so they should be encapsulated in a non-normative disclosure (see other places where we have dictionaries described in non-normative sections).

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+    :   <dfn>service_worker</dfn>
+    ::  The [=/service worker=] script URL. It MUST be [=same origin=] with the [=config file=],
```
npm1: Since this will be non-normative, try to avoid using normative terms like "MUST".

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+        from this URL (the script URL with its last path segment removed).
+</dl>
+
+Note: There is no separate "enabled" switch. An [=IDP=] disables interception by removing the
+{{IdentityProviderWellKnown/identity_handler}} member from its [=well-known file=], which causes the
+[=user agent=] to [=unregister the identity handler=] and use the direct-network path.
+
+Each [=/origin=] has an associated <dfn>FedCM identity-handler storage key</dfn>: a
+[=storage key=] the [=user agent=] constructs for that [=/origin=]'s FedCM identity handler
+[[!STORAGE]]. It MUST be distinct from the [=/origin=]'s first-party [=storage key=], such that:
```
npm1: ditto `[[!STORAGE]]` seems out of place.

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+
+1. no Web API surface accepts this [=storage key=] or otherwise allows page script to cause a
```
npm1: nit: capitalize first word? Here and below

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+[=IDP=]'s [=/origin=].
+
+Note: When the [=user agent=] later dispatches a credentialed request to the identity handler, it
+uses this explicit, [=user agent=]-only association between the [=IDP=] and the registration it
+created (keyed by the [=FedCM identity-handler storage key=]), rather than the implicit service
+worker match path used for client-controlled fetches. Because a credentialed FedCM request has no
+controlling client, this ensures the request reaches only a [=/service worker=] the [=user agent=]
+registered for that specific [=IDP=].
+
+The [=user agent=] MUST serialize [=register the identity handler=] and
```
npm1: This needs to be baked into the algorithm itself, not just stated as a requirement.

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+
+    1. Assert: These steps are running [=in parallel=].
+    1. Let |idpOrigin| be |configUrl|'s [=url/origin=].
+    1. Let |scriptURL| be the result of running the [=URL parser=] on |handler|'s
+        {{IdentityHandler/service_worker}} with |configUrl| as the base URL.
+    1. If |scriptURL| is failure, return.
+    1. If |scriptURL|'s [=url/origin=] is not [=same origin=] with |idpOrigin|, return.
+
+        Note: A cross-origin declaration is ignored, and credentialed requests fall back to the
+        network. The [=same origin=] requirement is stated normatively on
```
npm1: This second sentence seems unneeded?

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+        {{IdentityHandler/service_worker}} above.
+    1. Let |scope| be the result of running the [=URL parser=] on "`./`" with |scriptURL| as the
```
npm1, line 1422: Is this not something that we can delegate to be computed in the service worker algorithm?

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+The [=user agent=] MUST serialize [=register the identity handler=] and
+[=unregister the identity handler=] per [=/origin=]: at most one of them runs at a time for a given
+[=/origin=], and their effects are observable in the order they were dispatched.
+
+Note: Concurrent FedCM requests for the same [=IDP=] can otherwise observe different
+[=well-known files=] from the HTTP cache and race to opposite conclusions about whether an identity
+handler is declared.
+
+<div algorithm>
+To <dfn>register the identity handler</dfn> given a [=/URL=] |configUrl| and an
```
npm1, line 1409: Can we avoid calling this `configUrl`? Perhaps `serviceWorkerUrl` or `swUrl`?

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+    1. If |registration| is not null, |registration|'s [=active worker=] is not null, and
+        |registration|'s [=active worker=]'s [=service worker/script url=] [=url/equals=]
+        |scriptURL|, then:
+        1. [=Soft update=] |registration|.
+
+            Note: The [=user agent=] proceeds with the [=active worker=] that is already
+            registered while it checks for a newer version in parallel, so update discovery never
+            delays a FedCM request.
+        1. Return.
+    1. [=Unregister the identity handler=] given |idpOrigin|.
```
npm1, line 1450: Is this only needed when `registration` is not null?

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+</div>
+
+<div algorithm>
+To <dfn>unregister the identity handler</dfn> given an [=/origin=] |idpOrigin|, run these steps:
+
+    1. Assert: These steps are running [=in parallel=].
+    1. Let |storageKey| be the [=FedCM identity-handler storage key=] for |idpOrigin|.
+    1. [=list/For each=] (|entryStorageKey|, |entryScope|) of the [=registration map=]'s
+        [=map/keys=]:
+        1. If |entryStorageKey| is not |storageKey|, [=iteration/continue=].
```
npm1: Is there a service worker method to compare equality of storage keys?

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+        fetch or install is not reported to the [=IDP=]. Credentialed requests fall back to the
+        network until a later FedCM flow retries the registration.
+</div>
+
+<div algorithm>
+To <dfn>unregister the identity handler</dfn> given an [=/origin=] |idpOrigin|, run these steps:
+
+    1. Assert: These steps are running [=in parallel=].
+    1. Let |storageKey| be the [=FedCM identity-handler storage key=] for |idpOrigin|.
+    1. [=list/For each=] (|entryStorageKey|, |entryScope|) of the [=registration map=]'s
```
npm1, line 1488: Is iterating over the registration map while at the same time modifying it safe? Alternatively, we could break when we find one (we only expect one), or do a loop first to gather the ones we find, assert that the size is 1, and schedule the deletion?

spec/index.bs
```diff
@@ -1316,11 +1348,159 @@ to keep their actual config files on an arbitary path while allowing the user ag
...
+        1. Let |job| be the result of running [=create job=] with *unregister*, |entryStorageKey|,
+            |entryScope|, null, null, and null.
+        1. Invoke [=schedule job=] with |job|.
+
+    Note: The [=user agent=] iterates the [=registration map=] rather than [=unregister=]ing one
+    known [=/scope=], because the {{IdentityProviderWellKnown/identity_handler}} declaration, and
+    with it the declared [=/scope=], may already be gone by the time these steps run.
+</div>
+
+The [=user agent=] [=unregister the identity handler|unregisters the identity handler=] when the
```
npm1: Is this meant to be a note? Otherwise it seems weird to be mixing this non-normative text with the algorithms

npm1 (commented):
Neat thanks for writing this up! I guess this is only part of the changes needed, the other part being actually using the registered service worker in the credentialed fetches?

yoshisatoyanagisawa:
Thanks for drafting this.

Since there are still several unresolved threads in the [design doc](https://docs.google.com/document/d/1PgiLzfvE8NsSnxgJYA3WUEld0FguPfPETHDPjUb6wlo/edit?tab=t.0#heading=h.49f4nui9ffg2), I just wanted to clarify if we are moving forward with parallel discussions here.

Separately, speaking as a ServiceWorker editor, I'm concerned about the use of non-exported ServiceWorker algorithms in this PR. Those algorithms are not guaranteed to remain stable in the future, so we should avoid calling them directly. Instead, I'd recommend using the exported Fire Functional Event algorithm, which is intended for this kind of integration.

TallTed:
I have definite desire and intent to review this, but I would prefer to do so after the existing comments are addressed, given their number.

Brandr0id:
> Thanks for drafting this.
> 
> Since there are still several unresolved threads in the [design doc](https://docs.google.com/document/d/1PgiLzfvE8NsSnxgJYA3WUEld0FguPfPETHDPjUb6wlo/edit?tab=t.0#heading=h.49f4nui9ffg2), I just wanted to clarify if we are moving forward with parallel discussions here.
> 
> Separately, speaking as a ServiceWorker editor, I'm concerned about the use of non-exported ServiceWorker algorithms in this PR. Those algorithms are not guaranteed to remain stable in the future, so we should avoid calling them directly. Instead, I'd recommend using the exported Fire Functional Event algorithm, which is intended for this kind of integration.

Thanks for taking a look. Yeah to clarify the design doc is more focused on a Chromium based impl of this and may have specifics related to Chromium and that implementation/architecture. In parallel we'd like to ensure we align on the spec'd behaviour/integrations.

As for the exported methods and functional event. This PR focuses more on the JIT registration of a SW in FedCM flows and I'll post a follow up that focuses on the event firing (and does use Fire Function Event :) ). I wanted to try and split the spec change up a bit so it was a little more discrete size wise for reviewers to look at. For this though I'm not sure we currently have the necessary exported primitives to do JIT registration or unregistration. I think Payment Handler may also benefit from this.

Would you be open to exporting soft-update and adding/exposing something in the SW spec like:

```
<dfn export>Register a Service Worker</dfn>
: Input
:: |storageKey|, a storage key
:: |scriptURL|, a URL
:: |scopeURL|, a URL
:: |workerType|, a worker type
:: |updateViaCache|, an update via cache mode

1. Assert: |scriptURL|'s origin is same origin with |scopeURL|'s origin.
2. If the result of running potentially trustworthy origin with |scriptURL|'s
   origin is Not Trusted, return.
3. Let |job| be the result of running Create Job with register, |storageKey|,
   |scopeURL|, |scriptURL|, null, and null.
4. Set |job|'s worker type to |workerType|.
5. Set |job|'s update via cache mode to |updateViaCache|.
6. Invoke Schedule Job with |job|.

<dfn export>Unregister Service Workers</dfn>
: Input
:: |storageKey|, a storage key

1. Let |scopes| be an empty list.
2. For each (|entryStorageKey|, |entryScope|) of registration map's keys:
   1. If |entryStorageKey| equals |storageKey|, append |entryScope| to |scopes|.
3. For each |scope| of |scopes|:
   1. Let |job| be the result of running Create Job with unregister,
      |storageKey|, |scope|, null, null, and null.
   2. Invoke Schedule Job with |job|.
```

spec/index.bs
Brandr0id: Updated this, thanks!

spec/index.bs
Brandr0id: I took a stab at updating this to a parallel queue.

spec/index.bs
Brandr0id: Fixed.

spec/index.bs
Brandr0id: Updated!

spec/index.bs
Brandr0id: removed.

spec/index.bs
Brandr0id: updated

spec/index.bs
Brandr0id: Updated the wording a bit here.

spec/index.bs
Brandr0id: Updated

spec/index.bs
Brandr0id: fixed

spec/index.bs
Brandr0id: left this in but added some notes.

spec/index.bs
Brandr0id: updated

spec/index.bs
Brandr0id: I believe it is implicitly handled as it'll stay registered until it is unregistered explicitly or the storage for it is cleared.

spec/index.bs
```diff
@@ -1312,11 +1315,13 @@ or failure.
...
         1. If |allowed_config_url| is not [=url/equal=] to |configUrl|, return failure.
+    1. Let |idpOrigin| be |configUrl|'s [=url/origin=].
     1. If |wellKnown|'s {{IdentityProviderWellKnown/identity_handler}} is present, then
-        [=in parallel=], [=register the identity handler=] given |configUrl| and |wellKnown|'s
+        [=enqueue steps=] to |idpOrigin|'s [=identity handler queue=] to
+        [=register the identity handler=] given |configUrl| and |wellKnown|'s
         {{IdentityProviderWellKnown/identity_handler}}.
-    1. Otherwise, [=in parallel=], [=unregister the identity handler=] given |configUrl|'s
-        [=url/origin=].
+    1. Otherwise, [=enqueue steps=] to |idpOrigin|'s [=identity handler queue=] to
```
npm1, line 1323: Is this a common pattern, eg having a queue per origin? Is it needed here, or is having a single queue for this type of task enough?

npm1 (commented):
Before we continue, we should probably align with service worker spec editors regarding whether we need to implement this using the fire functional event, or whether they are ok exporting some algorithms so that they can be used from FedCM

